PR #1, opened 2021-11-16
Bug 564503 Allow alphabetical sorting in Outline view

sebthom:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=564503


mickaelistria:
While this solution works, it's a bit limited since it for instance doesn't affect the Quick Outline (Ctrl+Shift+O), while one could expect it would show same order as Outline view. I think (without being certain) trying to use a commonSorter from the navigation framework would be a more reusable solution that could also affect Quick Outline and other trees based on the same content without extra change.

sebthom:
I tried to specify a commonSorter but it is ignored when opening the quick outline. Btw, the Java Quick Outline also doesn't sort if sorting is enabled in the outline view.

```xml
   <extension
         point="org.eclipse.ui.navigator.navigatorContent">
      <navigatorContent
            activeByDefault="true"
            contentProvider="org.eclipse.lsp4e.outline.LSSymbolsContentProvider"
            labelProvider="org.eclipse.lsp4e.outline.SymbolsLabelProvider"
            id="org.eclipse.lsp4e.outline.content"
            name="LS Symbols">
         <triggerPoints>
            <or></or>
         </triggerPoints>
         <enablement>
            <instanceof
                  value="org.eclipse.lsp4e.LanguageServiceAccessor$LSPDocumentInfo">
            </instanceof>
         </enablement>
         <commonSorter
               class="org.eclipse.lsp4e.outline.OutlineSorter">
         </commonSorter>
      </navigatorContent>
   </extension>
```

I also added some breakpoints in the LSSymbolsContentProvider but the are not invoked when I open the quick outline. So I currently don't understand how the tree of the quick outline is actually populated.

sebthom:
I managed to change the quick outline too. It was using implemented using a separate content provider. I changed it to use LSSymbolsContentProvider which has the nice side effect that the current symbol is now automatically shown and selected in the quick outline. Before the quick outline was always collapsed.

mickaelistria:
Thanks. Let's merge it because it's doing a good job; maybe later we'll come back to it to have common sorter (eg if we want the file symbols to be shown in Project Explorer as children of the file), but it's not a requirement at the moment.

PR #7, opened 2021-11-19
Fix adapter casting in CNFOutlinePage

sebthom:
Use `getAdapter(ITextViewer.class)` instead of `getAdapter(ITextOperationTarget.class)` to retrieve an `ITextViewer` instance.

### DIFF
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/outline/CNFOutlinePage.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/outline/CNFOutlinePage.java
@@ -24,7 +24,6 @@ import org.eclipse.core.runtime.preferences.InstanceScope;
 import org.eclipse.jdt.annotation.Nullable;
 import org.eclipse.jface.text.BadLocationException;
 import org.eclipse.jface.text.IDocument;
-import org.eclipse.jface.text.ITextOperationTarget;
 import org.eclipse.jface.text.ITextSelection;
 import org.eclipse.jface.text.ITextViewer;
 import org.eclipse.jface.text.ITextViewerExtension5;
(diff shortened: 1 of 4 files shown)


org.eclipse.lsp4e/src/org/eclipse/lsp4e/outline/CNFOutlinePage.java
```diff
@@ -73,11 +72,7 @@ public CNFOutlinePage(LanguageServer languageServer, @Nullable ITextEditor textE
 		preferences = InstanceScope.INSTANCE.getNode(LanguageServerPlugin.PLUGIN_ID);
 		preferences.addPreferenceChangeListener(this);
 		this.textEditor = textEditor;
-		if (textEditor != null) {
-			this.textEditorViewer = ((ITextViewer) textEditor.getAdapter(ITextOperationTarget.class));
-		} else {
-			this.textEditorViewer = null;
-		}
+		this.textEditorViewer = textEditor == null ? null : textEditor.getAdapter(ITextViewer.class);
```
mickaelistria, line 75: I think this will now return null in many cases. The reason to use ITextOperationTarget is a hack: the getAdapter() method won't return a viewer if queried for the viewer but null; querying for the ITextOperationTarget usually returns the viewer.
One possibility would be first to extract some method `getTextViewer(ITextEditor editor)`. In this method, we can first try adapting to ITextViewer (ideally with `Adapters.getAdapter(editor, ITextViewer.cclass)` instead of the direct call to `editor.getAdapter()`), then if this return null try adapting to ITextOperationTarget, and verify result is instanceof ITextViewer before returning it, or null.

mickaelistria (changes requested):

sebthom:
@mickaelistria done

org.eclipse.lsp4e/src/org/eclipse/lsp4e/outline/CNFOutlinePage.java
sebthom: done

mickaelistria:
Thanks!

PR #9, opened 2021-11-20
Add support for <commonSorter> in outline

sebthom:
- The outline and the quick outline now respect the `<commonSorter>` of the navigatorContent extension configuration.
- Additionally the quick outline doesn't use hard coded content/label providers anymore but the ones specified in the navigatorContent extension configuration.
- You will now actually see the `Computing...` message in the outline when it is initially loaded and not a blank view for 1 to 2 seconds.

### DIFF
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/symbols/LSPSymbolInFileDialog.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/symbols/LSPSymbolInFileDialog.java
@@ -19,14 +19,10 @@ import org.eclipse.jface.text.BadLocationException;
 import org.eclipse.jface.text.IDocument;
 import org.eclipse.jface.viewers.IStructuredSelection;
 import org.eclipse.jface.viewers.TreeViewer;
-import org.eclipse.jface.viewers.Viewer;
 import org.eclipse.lsp4e.LSPEclipseUtils;
 import org.eclipse.lsp4e.LanguageServerPlugin;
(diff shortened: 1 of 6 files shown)


org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/symbols/LSPSymbolInFileDialog.java
```diff
@@ -41,40 +37,42 @@
...
-		}
-		viewer.setLabelProvider(new SymbolsLabelProvider());
+		final var contentService = new NavigatorContentService(CNFOutlinePage.ID, viewer);
+		filteredTree.addDisposeListener(ev -> contentService.dispose());
+		final var contentProvider = contentService.createCommonContentProvider();
+		viewer.setContentProvider(contentProvider);
+
+		var sorter = new CommonViewerSorter();
+		sorter.setContentService(contentService);
+		viewer.setComparator(sorter);
```
mickaelistria, line 73: What does happen if alphabetical order is disabled? Is the order of elements (as provided by the model/LS) respected? It's important that we respect LS ordering by default. It would be great is some test get added about that.

sebthom: Yes. The OutlineSorter I implemented respects the current sort setting. If sorting is disabled (default) the compare funtion returns `0` meaning the items are shown in the order they are received from the LS.

PR #17, opened 2021-11-24
minor code cleanup

sebthom:
I ran the Eclipse Code Cleanup action with a few uncritical rules:
- Initialize collection at creation
- Make inner classes static where possible
- Primitive parsing
- Use boolean literals instead of Boolean.TRUE/FALSE when used as a primitive
- Operate on Maps directly
- Remove unused imports
- Simplify lambda expression and method reference syntax
- Use lambda where possible
- Use diamond operator
- Use Autoboxing / Unboxing
- valueOf() rather than instantiation


mickaelistria:
Thanks!

jonahgraham (commented):
LGTM on the debug parts.

PR #18, opened 2021-11-24
add stale bot and dependabot

sebthom:


.github/stale.yml
```diff
@@ -0,0 +1,24 @@
+# Configuration for probot-stale - https://github.com/probot/stale
+
+# Number of days of inactivity before an issue becomes stale
+daysUntilStale: 90
+
+# Number of days of inactivity before a stale issue is closed
+daysUntilClose: 7
```
akurtakov, line 7: That's a bit too short for my taste. I would prefer if this is e.g. 30 days so people have time to react if coming back from vacation and etc. @mickaelistria wdyt?

mickaelistria:
I don't think we should automatically close issue as long as the amount of open issues is not a problem. There can be issues that are several years old and worth being fixed but receiving no particular attention for a long time, but get addressed later (I happened to work on several of such issues over the last weeks). Autoclosing would make those issues ignored forever, it's a risk of reducing quality.
So as long as the backlog remains easy to navigate, it's probably not worth the risk.

sebthom:
To prevent auto-closing of issues they can be labeled with `pinned` or some other custom label. 

I guess a lot of old issues in the old bugzilla tracker didn't apply to the current version anymore and I doubt they would ever be worked-on/reevaluated/closed again.

mickaelistria:
What's the benefit in adding a label automatically?

sebthom:
My suggestion here is to see the auto-closing as a mechanism that requires users revalidate an issue after a given time to ensure that it is still valid. If an LSP4E maintainer thinks an issue is worth fixing/investigating but not atm. he could pin the issue so it does not get closed. Otherwise you end up with multi-year old issues that may not even be relevant anymore and only pollute the issue tracker. Like this for example https://bugs.eclipse.org/bugs/show_bug.cgi?id=537198 or this https://bugs.eclipse.org/bugs/show_bug.cgi?id=539126 How knows that state of these issues?

Also as an external contributor I find it more encouraging to contribute if I can pick an issue from the issue list knowing that it probably still is relevant to be solved.

mickaelistria:
So that would mean that by default, all issues get closed without particular attention? Does this map the real lifecycle of an issue? IMO, an issue is more ususally still relevant unless (humanly) determined otherwise than the other way round. Those bots tend to assume that most issues usually become magically irrelevant. I don't share this impression about issues and would rather see maintainers doing the triaging manually than relying on a bot that pretends that issue that were reported at some point aren't relevant any more.
At least, that's my opinion until the maintainers feel unable to consider the backlog efficiently, which is not the case yet.

sebthom:
ok, I removed the stalebot config from the PR

mickaelistria:
Thanks.
I think the stalebot will be interested in a not too far future, when the backlog grows and becomes too annoying to deal with.

PR #19, opened 2021-11-24
Don't sort outline by default

sebthom:
While I did set it to false by default in code at https://github.com/eclipse/lsp4e/blob/5132a7f80eabe4474ae8369dcf8d9e3448f47d95/org.eclipse.lsp4e/src/org/eclipse/lsp4e/outline/OutlineSorter.java#L67

I overlooked the declaration in the plugin.xml

### DIFF
--- a/org.eclipse.lsp4e/META-INF/MANIFEST.MF
+++ b/org.eclipse.lsp4e/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: Language Server Protocol client for Eclipse IDE (Incubation)
 Bundle-SymbolicName: org.eclipse.lsp4e;singleton:=true
-Bundle-Version: 0.13.8.qualifier
+Bundle-Version: 0.13.9.qualifier
 Bundle-RequiredExecutionEnvironment: JavaSE-11
 Require-Bundle: org.eclipse.core.runtime;bundle-version="3.12.0",
  org.eclipse.equinox.common;bundle-version="3.8.0",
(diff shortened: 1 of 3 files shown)


mickaelistria:
Because of recent 0.20.1 release yesterday; you'd need to bump version for the modified bundle.

sebthom:
@mickaelistria fixed. Btw. why is the lsp verrsion (0.20.1) not reflected by the plugins. I.e. why are not all modules bumped to the same new next minor version (0.20.2)?

mickaelistria:
Because we use [semantic versioning](https://wiki.eclipse.org/Version_Numbering) for the bundles; and sometimes, only 1 of the bundles is modified/bumped for a release. Sometimes, the change is a feature change (+0.1.0) and sometimes it's just a fix (+0.0.1). In practice, the "project" version is semantically meaningless, all that matter are bundle versions.

org.eclipse.lsp4e/META-INF/MANIFEST.MF
```diff
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: Language Server Protocol client for Eclipse IDE (Incubation)
 Bundle-SymbolicName: org.eclipse.lsp4e;singleton:=true
-Bundle-Version: 0.13.8.qualifier
+Bundle-Version: 0.13.9.qualifier
```
mickaelistria, line 5: Same change is needed in pom file.

sebthom: Done

mickaelistria:
Perfect, thanks!

PR #20, opened 2021-11-25
Don't move cursor/change text selection when opening quick outline

sebthom:
This PR fixes the issue that when opening the quick outline the nearest outline item is automatically selected in the text editor, i.e. the selection in the text editor is changed merely by opening the quick outline.

*Example:*
The editor cursor is in line 54. I press CTRL+O to open the quick outline and the `bindEvents` method name is selected automatically.

screen

With this PR the nearest item is still selected in the outline tree but the text editor selection is not changed only by opening the outline.


mickaelistria:
Thanks!

PR #24, opened 2021-12-07
Do not raise exception if the file does not exists

rubenporras:
Instead of that, use only the file name to extract the content types.


mickaelistria:
Thanks!

Issue #30, opened 2021-12-14
UI freeze on LSPLinkedEditingAutoEditStrategy.customizeDocumentCommand


apupier:
recently upgraded to Fedora 35 and Eclipse 2021-12

reproduced 2 times:
* a commit description hover appeared and do not want to vanish.I was in multipage pom editor, in source view, displayed git blame and hovered over it
* Scrolling through the source tab of the multipage editor of a pom.xml.

both times were with this https://github.com/camel-tooling/camel-language-server/ (but no ttried yet with anothe rone)

```
Name: main
State: WAITING on java.util.concurrent.CompletableFuture$Signaller@7d37456e
Total blocked: 242  Total waited: 156

Stack trace: 
java.base@17/jdk.internal.misc.Unsafe.park(Native Method)
java.base@17/java.util.concurrent.locks.LockSupport.park(LockSupport.java:211)
java.base@17/java.util.concurrent.CompletableFuture$Signaller.block(CompletableFuture.java:1864)
java.base@17/java.util.concurrent.ForkJoinPool.unmanagedBlock(ForkJoinPool.java:3463)
java.base@17/java.util.concurrent.ForkJoinPool.managedBlock(ForkJoinPool.java:3434)
java.base@17/java.util.concurrent.CompletableFuture.waitingGet(CompletableFuture.java:1898)
java.base@17/java.util.concurrent.CompletableFuture.get(CompletableFuture.java:2072)
org.eclipse.lsp4e.operations.linkedediting.LSPLinkedEditingAutoEditStrategy.customizeDocumentCommand(LSPLinkedEditingAutoEditStrategy.java:41)
org.eclipse.jface.text.TextViewer.customizeDocumentCommand(TextViewer.java:3705)
org.eclipse.jface.text.TextViewer.handleVerifyEvent(TextViewer.java:3740)
org.eclipse.jface.text.source.projection.ProjectionViewer.handleVerifyEvent(ProjectionViewer.java:1248)
org.eclipse.jface.text.TextViewer$TextVerifyListener.verifyText(TextViewer.java:426)
org.eclipse.swt.widgets.TypedListener.handleEvent(TypedListener.java:277)
org.eclipse.swt.widgets.EventTable.sendEvent(EventTable.java:89)
org.eclipse.swt.widgets.Display.sendEvent(Display.java:5895)
org.eclipse.swt.widgets.Widget.sendEvent(Widget.java:1515)
[...]

PR #34, opened 2021-12-17
UI freeze on LSPLinkedEditingAutoEditStrategy.customizeDocumentComman…

vrubezhny:
…d #30

Issue: https://github.com/eclipse/lsp4e/issues/30

Change-Id: I4b3ffc0c23d6c3fb5411a2e097eedded3dfb33d6


mickaelistria:
Although I'm not sure it's the best solution (there might be some better way to avoid locks), having a timeout here seems like a good workaround for the issue and a good thing in general. So let's merge.

martinlippert:
Just out of curiosity: why was the sorting removed in this change? Was that related to the deadlock/timeout?

PR #41, opened 2022-01-08
Ensure workspaceFolders uses standard file:// prefix

### LINKED ISSUE #29: Solargraph not working after eclipse-update from 2021-09 to 2021-12
I've created this issue because I cannot reopen #27 
@PyvesB checked my logfiles and it seems that the completion did not work because auf a change in lsp4e the deprecates rootUri in favor of workspacesFolders where workspaceFolders are send with a different uri-format.

in my case the URIs starting with with file:/Z:... instead of file:///Z:... which cannot be handled by solargraph.
It seems that rootUri was send with file:///Z:... and therefore it worked in eclipse 2021-09

@PyvesB added a really good esplanation to the issue I've created in his project: https://github.com/PyvesB/eclipse-solargraph/issues/17#issuecomment-991607202

PyvesB:
Fixes #29.

Signed-off-by: Pierre-Yves Bigourdan <10694593+PyvesB@users.noreply.github.com>

### DIFF
--- a/org.eclipse.lsp4e.test/META-INF/MANIFEST.MF
+++ b/org.eclipse.lsp4e.test/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: Tests for language server bundle (Incubation)
 Bundle-SymbolicName: org.eclipse.lsp4e.test;singleton:=true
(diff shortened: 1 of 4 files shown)


mickaelistria:
Version of the test bundle needs to be bumped

mickaelistria:
Version of the test bundle needs to be bumped

mickaelistria:
Failing tests are not related, let's merge.
Thanks!

PR #47, opened 2022-01-18
Support finding resources for remote URIs

rubenporras:


rubenporras:
It is the same as before but rebased on top of master.

mickaelistria:
Thanks!

Issue #52, opened 2022-01-21
Linked Edit ranges erroneous when ranges are on the same line


mickaelistria:
Using for example Wild Web Developer with XML support, for file
```xml
<tag>lol</tag>
```
If one does edit on of the tag quickly or switching the caret, the result becomes faulty. Error can take many forms: characters included at wrong location, broken XML, or even IDE error dialog caused by some exception about invalid ranges.

mickaelistria:
Causes https://github.com/eclipse-m2e/m2e-core/issues/513

PR #61, opened 2022-01-28
Issue #52 - Use JFace LinkedModeUI for linkedEditRanges

mickaelistria:
This API is working well and fixes some usual bugs.


org.eclipse.lsp4e.test/src/org/eclipse/lsp4e/test/linkedediting/LinkedEditingTest.java
```diff
@@ -102,44 +96,8 @@ protected boolean condition() {
...
-			lastChange = didChangeExpectation.get(1000, TimeUnit.MILLISECONDS);
-			assertNotNull(lastChange.getContentChanges());
-			assertEquals(1, lastChange.getContentChanges().size());
-			change0 = lastChange.getContentChanges().get(0);
-			assertEquals(expectedChangeText, change0.getText());
-			assertEquals(expectedChangeText, viewer.getDocument().get());
-		} catch (Exception e) {
-			e.printStackTrace();
-			fail(e.getMessage());
-		}
```
mickaelistria, line 143: I removed that because we basically only need to verify that link edit mode is on, we then fully rely on it.
The LinkedModeUI is actually triggering an extra change instead of using a documentCommand as done previously. We may want to improve it (in JFace) at some point).

Issue #35, opened 2021-12-22
Every Outline view now has an extra "Sort Outline" toggle


nitind:
org.eclipse.lsp4e 0.13.8.202111241523 bundle in the Enterprise Java and Web 2021-12 Release package.

Every editor I open is getting an additional "Sort Outline" toggle in its Outline view, even ones that already have their own sort option, and editors that don't even have outlines. From what I can tell, it's been added to the View itself, not just to the ContentOutlinePages' site's actionbars where it makes sense.

<img width="310" alt="Screen Shot 2022-01-15 at 12 43 36 AM" src="https://user-images.githubusercontent.com/2684153/149610622-8802dd5e-1485-490d-8d12-6ed5cf1ec4aa.png">

Additionally, the label does not need to include the word, "Outline", and it does not appear to be properly externalized.

mickaelistria:
@vrubezhny Please investigate; the contribution needs a visibleWhen element that would only apply if the outline view comes from LSP4E.

PR #64, opened 2022-02-01
Every Outline view now has an extra "Sort Outline" toggle #35

vrubezhny:
Change-Id: If2aa18072e2154aad5dfa10c0cd363cf081d47d6

### DIFF
--- a/org.eclipse.lsp4e/plugin.properties
+++ b/org.eclipse.lsp4e/plugin.properties
@@ -25,5 +25,7 @@ notification.category.label = LSP
 notification.event.label = LSP Notification
 command.toggle.highlight.label = Toggle Mark Occurrences
 command.toggle.highlight.name = Toggle Highlight
+command.toggle.outline.sort.name = Sort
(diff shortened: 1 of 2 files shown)


org.eclipse.lsp4e/plugin.properties
```diff
@@ -25,5 +25,7 @@ notification.category.label = LSP
 notification.event.label = LSP Notification
 command.toggle.highlight.label = Toggle Mark Occurrences
 command.toggle.highlight.name = Toggle Highlight
+command.toggle.outline.sort.name = Sort
```
cdietrich, line 28: change intended?

vrubezhny: Usually such command's tool tip says just "Sort", if you mean the change from "Sort Outline" to "Sort" label text then yes, it's intended. Externalization - yes, for sure

mickaelistria:
Works as expected, thanks!

PR #76, opened 2022-02-08
[Debug] Adapters only participate to variable resolution on related docs

mickaelistria:
Ignore documents that are not related to a current DSPStackFrame.

### DIFF
--- a/org.eclipse.lsp4e.debug/src/org/eclipse/lsp4e/debug/breakpoints/DocumentUtils.java
+++ b/org.eclipse.lsp4e.debug/src/org/eclipse/lsp4e/debug/breakpoints/DocumentUtils.java
@@ -1,0 +1,106 @@
+/*******************************************************************************
+ * Copyright (c) 2022 Red Hat, Inc. and others.
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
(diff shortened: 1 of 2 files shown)


jonahgraham (approved):
This seems fine to me.

But I don't know why the limitation is needed. Prevent extra JSON-RPC traffic or pop-ups when looking at another file?

FWIW The Eclipse CDT debugger does not limit like this. Not sure what else does/doesn't though - perhaps CDT is the outlier. IIUC JDT is more intelligent and uses parse tree, so won't even show variables on other text in the same file unless it is the same variable.

mickaelistria:
It's for https://bugs.eclipse.org/bugs/show_bug.cgi?id=540448 which reuses the adapters. So without this filter, this particular adapter may still return values for non related files that contain a token which maps a variable that can be resolved for the current DAP frame.
For example, if I'm debugging some JS `for (int i = 0; i < 100; i++) { }` and get the value of `i` printed with WIP for https://bugs.eclipse.org/bugs/show_bug.cgi?id=540448  , then all editors that are capable of code minings will show the value of `i` variable from this JS execution for every `i` token they contain.

jonahgraham:
Thanks @mickaelistria for the explanation.

PR #81, opened 2022-02-19
[Debug] Add frameId to EvaluateArguments request

PyvesB:
Resolves `// TODO args.setFrameId(0);` for `EvaluateArguments` requests, in turn fixing https://github.com/castwide/readapt/issues/11.

### DIFF
--- a/org.eclipse.lsp4e.debug/src/org/eclipse/lsp4e/debug/console/DSPStreamsProxy.java
+++ b/org.eclipse.lsp4e.debug/src/org/eclipse/lsp4e/debug/console/DSPStreamsProxy.java
@@ -1,17 +1,23 @@
 /*******************************************************************************
- * Copyright (c) 2017 Kichwa Coders Ltd. and others.
+ * Copyright (c) 2017-2022 Kichwa Coders Ltd. and others.
  * This program and the accompanying materials are made
  * available under the terms of the Eclipse Public License 2.0
  * which is available at https://www.eclipse.org/legal/epl-2.0/
  *
  * SPDX-License-Identifier: EPL-2.0
+ * Contributors:
+ *  Pierre-Yves Bigourdan <pyvesdev@gmail.com> - Add frameId to EvaluateArguments request
  *******************************************************************************/
 package org.eclipse.lsp4e.debug.console;
 
 import java.io.IOException;
 import java.util.concurrent.CompletableFuture;
 
+import org.eclipse.core.runtime.Adapters;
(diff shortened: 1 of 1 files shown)


org.eclipse.lsp4e.debug/src/org/eclipse/lsp4e/debug/console/DSPStreamsProxy.java
```diff
@@ -45,12 +50,28 @@ public DSPStreamMonitor getOutputStreamMonitor() {
...
-			future.thenAcceptAsync(response -> {
+			debugProtocolServer.threads().thenComposeAsync(threadsResponse -> {
+				Thread[] threads = threadsResponse.getThreads();
+				if (threads.length == 0) {
+					StackTraceResponse noStackFrames = new StackTraceResponse();
+					noStackFrames.setStackFrames(new StackFrame[0]);
+					return CompletableFuture.completedStage(noStackFrames);
+				}
+				StackTraceArguments arguments = new StackTraceArguments();
+				arguments.setThreadId(threads[0].getId());
```
mickaelistria, line 61: What does guarantee that we always want `threads[0]` ?

org.eclipse.lsp4e.debug/src/org/eclipse/lsp4e/debug/console/DSPStreamsProxy.java
```diff
@@ -45,12 +50,28 @@ public DSPStreamMonitor getOutputStreamMonitor() {
...
+				arguments.setStartFrame(0);
+				arguments.setLevels(1);
+				return debugProtocolServer.stackTrace(arguments);
+			}).thenComposeAsync(stackTraceResponse -> {
+				EvaluateArguments args = new EvaluateArguments();
+				args.setContext(EvaluateArgumentsContext.REPL);
+				args.setExpression(trimmed);
+				StackFrame[] stackFrames = stackTraceResponse.getStackFrames();
+				if (stackFrames.length > 0) {
+					args.setFrameId(stackFrames[0].getId());
```
mickaelistria, line 71: What does guarantee that we always want `stackFrames[0]` ?

mickaelistria (commented):
I'm not sure defaulting to the 1st returned thread/stacktrace is a good solution overall (unless there is only 1). We should deduce the desired thread or stacktrace from the `DebugUITools.getDebugContext()`.

PyvesB:
Thanks for the feedback, I've suggested an alternative patch accordingly.

mickaelistria:
Thanks!

PR #84, opened 2022-02-24
Improve performance by getting the document one for each resource

rubenporras:
Improve performance by getting the document one for each resource
instead of several times: one for each new diagnostic (when calling
computeMarkerAttributes when creating the marker), once for each updated
diagnostic (when calling updateMarker for markers to update).


org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/diagnostics/LSPDiagnosticsToMarkers.java
```diff
@@ -138,11 +138,12 @@ private void updateMarkers(PublishDiagnosticsParams diagnostics, IResource resou
 		}
 		IWorkspaceRunnable runnable = monitor -> {
 			if (resource.exists()) {
+				IDocument document = LSPEclipseUtils.getDocument(resource);
```
rubenporras, line 141: The call LSPEclipseUtils.getDocument(resource) has an extra code path which was not present in this file. I do not think that is a problem, but if it is a problem I can rework the PR avoid the extra code path.

mickaelistria:
While this change seems good in theory, it currently seems to introduce a regression (see CI output) with some markers that may not be correctly replaced/cleared.

rubenporras:
Thanks, I have noted that my change last week did not have the regression, but the farther optimized version I sent this morning has. I will take a look.

PR #87, opened 2022-03-08
Allow extended marker attributes

rubenporras:


mickaelistria:
Can you please rebase on top of master? Commits were squashed.

rubenporras:
> Can you please rebase on top of master? Commits were squashed.

Yes, I will do that if we want to merge this one

rubenporras:
This version I like less than PR-88, as it requires to calculate the markers derived from the resource for each diagnostic at it is only slightly simpler.
Just let me know which one you would like to merge and then I rebase & squash it.

rubenporras:
Let us close this PR and follow on PR-88

PR #88, opened 2022-03-08
Allow extended marker attributes2

rubenporras:


org.eclipse.lsp4e/src/org/eclipse/lsp4e/MarkerAttributeComputer.java
```diff
@@ -57,21 +58,41 @@ public int attributeCount() {
...
 
+	/**
+	 * Computes the attributes of a marker derived from its resource.
+	 *
+	 * @param resource,
+	 *            the {@link IResource} that contains the document.
+	 * @return a map with the marker attributes
+	 */
+	public @Nullable Map<String, Object> computeMarkerAttributes(@NonNull IResource resource) {
+		return null;
```
rubenporras, line 69: I could return here as well new ArrayList<>() or Collections.emptyList() if you prefer

rubenporras:
I would prefer this version compared to PR 87. If you like this one I would rebase and squash it.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/MarkerAttributeComputer.java
mickaelistria: Still, I really don't think this responsibility belongs to LSP4E. LSP4E is about binding LSP info to Eclipse model. What you need here (specific attributes for your specific marker type) doesn't really seem to be related to LSP at all, but more an applicative constraint on your end that depends more on your marker type than on the Language Server.

mickaelistria:
I actually believe that the version in #88 is better from LSP4E POV.
Another solution would be that the method takes that Map as in/out parameter, so it would be `fillMarkerAttributes(diagnostic, document, resource, Map<String, Object> attributes)` and one would jsut fill the attributes map.
It would allow you to check for existence or absence of the specific attributes to decide whether to include them or not. IMO, it seems sane to assume that one may want to look at existing attributes before setting some.

mickaelistria:
> Is this what you had in mind?

Not really, what I have in mind is more
```java
/*******************************************************************************
 * Copyright (c) 2016, 2017 Red Hat Inc. and others.
 * This program and the accompanying materials are made
 * available under the terms of the Eclipse Public License 2.0
 * which is available at https://www.eclipse.org/legal/epl-2.0/
 *
 * SPDX-License-Identifier: EPL-2.0
 *
 * Contributors:
 *  Rubén Porras Campo (Avaloq) - extracted to separate file
 *******************************************************************************/
package org.eclipse.lsp4e;

import java.util.HashMap;
import java.util.Map;

import org.eclipse.core.resources.IMarker;
import org.eclipse.core.resources.IResource;
import org.eclipse.jdt.annotation.NonNull;
import org.eclipse.jdt.annotation.Nullable;
import org.eclipse.lsp4j.Diagnostic;

/**
 * Computes and customize the attributes of a {@link org.eclipse.core.resources.IMarker} for given Diagnostic and context.
 */
public interface MarkerAttributeComputer {

	// Specific marker attributes defined by LSP4E
	public static final String LSP_DIAGNOSTIC = "lspDiagnostic"; //$NON-NLS-1$

	/**
	 * Computes the attributes of a marker for the given document and diagnostic.
	 *
	 * @param document,
	 *            the {@link Document} attached to the given resource
	 * @param diagnostic,
	 *            the {@link Diagnostic} to me mapped to a marker
	 * @param resource,
	 *            the {@link IResource} that contains the document
	 * @param markerAttributes,
	 *            the attributes for the marker. This method can read and modify that map.
	 */
	public void customizeMarkerForDiagnostic(@Nullable IDocument document, @NonNull IResource resource,
			@NonNull Diagnostic diagnostic, @NonNull Map<String, Object> markerAttributes);
}
```

The part
```java
		attributes.put(MarkerAttributeComputer.LSP_DIAGNOSTIC, diagnostic);
		attributes.put(IMarker.MESSAGE, diagnostic.getMessage());
		attributes.put(IMarker.SEVERITY, LSPEclipseUtils.toEclipseMarkerSeverity(diagnostic.getSeverity()));

		if (document != null) {
			Range range = diagnostic.getRange();
			int documentLength = document.getLength();
			try {
				int start = Math.min(LSPEclipseUtils.toOffset(range.getStart(), document), documentLength);
				int end = Math.min(LSPEclipseUtils.toOffset(range.getEnd(), document), documentLength);
				int lineOfStartOffset = document.getLineOfOffset(start);
				if (start == end && documentLength > end) {
					end++;
					if (document.getLineOfOffset(end) != lineOfStartOffset) {
						start--;
						end--;
					}
				}
				attributes.put(IMarker.CHAR_START, start);
				attributes.put(IMarker.CHAR_END, end);
				attributes.put(IMarker.LINE_NUMBER, lineOfStartOffset + 1);
			} catch (BadLocationException ex) {
				LanguageServerPlugin.logError(ex);
			}
		}
```
Should actually be set by default, in LSPDiagnosticToMarker. One benefit with that is that we may in some future allow multiple contributions to apply to a given marker, and also better isolate the "common" from the "custom" parts and don't need to remind user to call the parent implementation.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/MarkerAttributeComputer.java
```diff
@@ -11,89 +11,47 @@
...
-	 * to optimize size of the map at creation time.
+	 * Defines the expected number of attributes that will be added during
+	 * customization, it is used to optimize size of the map at creation time.
 	 *
 	 * @return the expected number of attributes that will be computed
 	 */
 	public int attributeCount() {
-		return 8;
+		return 0;
 	}
```
mickaelistria, line 37: How is this expected to be used in practice?

org.eclipse.lsp4e/src/org/eclipse/lsp4e/MarkerAttributeComputer.java
```diff
@@ -11,89 +11,47 @@
...
-				attributes.put(IMarker.CHAR_END, end);
-				attributes.put(IMarker.LINE_NUMBER, lineOfStartOffset + 1);
-			} catch (BadLocationException ex) {
-				LanguageServerPlugin.logError(ex);
-			}
-		}
-		return attributes;
+	public void customizeMarkerForDiagnostic(@Nullable IDocument document, @NonNull Diagnostic diagnostic,
+			@NonNull IResource resource, @NonNull Map<String, Object> attributes) {
+		// empty
```
mickaelistria, line 54: What about making it an interface then, and attribute count be a default method?

mickaelistria (commented):
Just 2 final (?) comments to discuss and I think we're done.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/MarkerAttributeComputer.java
rubenporras: I have removed it, let us keep it simple. Even if we add a bunch of extra attributes, the map should still behave correctly. If I understand correctly, the HashMap implementation will only resize the map when one reaches 64 attributes, and I doubt that will happen.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/MarkerAttributeComputer.java
rubenporras: Yes, let us do that. We are back to the very first implementation with an interface, I am happy with that :)

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/diagnostics/LSPDiagnosticsToMarkers.java
```diff
@@ -194,4 +194,42 @@ private IMarker getExistingMarkerFor(IDocument document, Diagnostic diagnostic,
...
+				attributes.put(IMarker.CHAR_END, end);
+				attributes.put(IMarker.LINE_NUMBER, lineOfStartOffset + 1);
+			} catch (BadLocationException ex) {
+				LanguageServerPlugin.logError(ex);
+			}
+		}
+
+
+		if (markerAttributeComputer != null) {
+			markerAttributeComputer.addMarkerAttributesForDiagnostic(diagnostic, document, resource, attributes);
```
rubenporras: my JDT null analysis tells me there is a potential NPE in here, it looks like a bug in the null analysis of eclipse because we check for null just before. The marker is annoying and it will be annoying for anyone fetching the codebase, so I have changed it to use Optional.

mickaelistria:
OK to merge now?

rubenporras:
Yes, OK to merge now. Thanks

PR #91, opened 2022-03-10
Lazy check for the existence of a resource

rubenporras:
Lazy check for the existence of a resource, as we already checked that
the resource exist before calling updateMarkers.

Instead, if the resource has ceased to exist between the point we call
updateMarkers and the point where we create/update the markers, catch
the CoreException and ignore it.


mickaelistria:
Thanks!

PR #92, opened 2022-03-16
Add a method to get an initializd LS for a given Language Server

rubenporras:
This is needed to initialize server before the user open a file in the editor, as we need to start/connect to the LanguageServer before the first editor uses it.

### DIFF
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServiceAccessor.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServiceAccessor.java
@@ -396,6 +396,21 @@ public class LanguageServiceAccessor {
 	}
 
 	/**
+	 * Return existing {@link LanguageServerWrapper} for the given definition. If
+	 * not found, create a new one with the given definition.
+	 *
+	 * @param project
+	 * @param serverDefinition
+	 * @return
+	 * @throws IOException
+	 */
+	@Deprecated
+	public static LanguageServerWrapper getLSWrapperForDefinition(@NonNull IProject project,
+			@NonNull LanguageServerDefinition serverDefinition) throws IOException {
+		return 	getLSWrapperForConnection(project, serverDefinition, null);
+	}
+
+	/**
 	 * Return existing {@link LanguageServerWrapper} for the given connection. If
 	 * not found, create a new one with the given connection and register it for
 	 * this project/content-type.


mickaelistria:
I don't get why it's necessary. The ConnectDocumentToLanguageServerSetupParticipant should take care of starting the LS if necessary. Why do you need to explicitly start the LS for a given project?

rubenporras:
You are right, the LS is started when a document is opened and an LS is registered for the content type of the document. Our problem is that we would like to start the LS before the user actually opens a file in the editor. 

The reason is that our workbench supports conceptually something like the target in eclipse. One of our developers selects a "target" and then we want to start language server and tell them which "target" is set, base on that the language server will setup some state and do an initial build (The initial build in our case just loads the precomputed state and will publish the markers of this state to the user, that is different to the eclipse target). In addition the LS needs some time to start. 

We could let most of this happen when the user opens the first file, but since our users already have the setting of the target as part of its workflow, we like to do that in advance to improve user experience. Otherwise when he opens the first file, he will need to wait.

So mainly starting the LS upfront is about usability.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServiceAccessor.java
```diff
@@ -395,6 +395,20 @@ private static boolean capabilitiesComply(LanguageServerWrapper wrapper,
...
+	/**
+	 * Return existing {@link LanguageServerWrapper} for the given definition. If
+	 * not found, create a new one with the given definition.
+	 *
+	 * @param project
+	 * @param serverDefinition
+	 * @return
+	 * @throws IOException
+	 */
+	public static LanguageServerWrapper getLSWrapperForDefinition(@NonNull IProject project,
```
mickaelistria, line 407: I'd suggest just `getLSWrapper` if possible.

rubenporras:
Thanks for the name suggestion, I like it better too.

I have changed it.

mickaelistria:
Do you see any place in the existing code that could leverage that method? As this is not officially API and there is no consumer for it, it can be too easily be removed or broken without any notice. If nothing better can be found, please at least add a test.

rubenporras:
Hi @mickaelistria ,

yes, I will take a look and add a test or make the code reuse the new API.

Regards

rubenporras:
Hi @mickaelistria ,

I did the changes to the existing tests and created a new one just for the new API. The deprecated API now uses internally the new getLSWrapper.

Regards

PR #93, opened 2022-03-18
Support linked files in IFileBufferListener

rubenporras:
Support linked files in IFileBufferListener by using LSPEclipseUtils#toUri, which will call an adapter to map the IResource to an URI.


mickaelistria:
IFileBuffer doesn't rely on IFile, and we cannot expect a buffer location to be always resolvable to a IFile as LSP4E also supports opening files from filesystem without having them imported into the workspace. We need to fallback to buffer.getLocation() in case no IFile can be found.

mickaelistria:
https://ci.eclipse.org/lsp4e/job/lsp4e-github/job/PR-93/7/testReport/org.eclipse.lsp4e.test.edit/DocumentDidCloseTest/testCloseExternalFile/ has been constantly failing for a few builds. Can you please check whether it's related?

rubenporras:
Hi @mickaelistria ,

yes, I will check and report in the PR.

Thanks

rubenporras:
Hi @mickaelistria ,

it was indeed a problem in the test in case the file does not have an URI which can be determined. I have fixed that.

Regards

PR #98, opened 2022-03-19
Run updateMarker job with lock on specific resource only

sebthom:
`ResourcesPlugin.getWorkspace().run(runnable, new NullProgressMonitor());` executes the job with a resource lock on the whole workspace which can result in interruption of auto build jobs of non-related projects at https://git.eclipse.org/c/platform/eclipse.platform.resources.git/tree/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/Workspace.java#n2262

### DIFF
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/diagnostics/LSPDiagnosticsToMarkers.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/diagnostics/LSPDiagnosticsToMarkers.java
@@ -27,8 +27,8 @@ import java.util.function.Consumer;
 import org.eclipse.core.resources.IMarker;
 import org.eclipse.core.resources.IProject;
 import org.eclipse.core.resources.IResource;
+import org.eclipse.core.resources.IWorkspace;
 import org.eclipse.core.resources.IWorkspaceRunnable;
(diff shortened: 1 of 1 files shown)


org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/diagnostics/LSPDiagnosticsToMarkers.java
```diff
@@ -167,7 +168,7 @@ private void updateMarkers(PublishDiagnosticsParams diagnostics, IResource resou
 				}
 			}
 		};
-		ResourcesPlugin.getWorkspace().run(runnable, new NullProgressMonitor());
+		ResourcesPlugin.getWorkspace().run(runnable, resource, IWorkspace.AVOID_UPDATE, new NullProgressMonitor());
```
mickaelistria, line 171: The best scheduling rule would be `resource.getWorkspace().getRuleFactory().markerRule(resource)`

sebthom: Ok, I changed it accordingly.

mickaelistria:
Thanks!

Issue #94, opened 2022-03-14
Linked editing doesn't exit on whitespaces and other non-word chars


vrubezhny:
When editing a "linked" element, like `<h3|></h3>`, for example, typing in any characters at position pointed by "|" in opening tag gets reflected (copied) to the linked closing tag which is a wrong behavior.

mickaelistria:
https://github.com/eclipse/lsp4e/commit/f427e37b222fbc56e481127fc7e950ba383a6462 should improve this a bit. However, it doesn't (yet) add support for `wordPattern`.

sebthom:
Fixed by https://github.com/eclipse/lsp4e/pull/100 and https://github.com/eclipse/lsp4e/pull/95

PR #100, opened 2022-03-30
Linked editing doesn't exit on whitespaces and other non-word chars #94

vrubezhny:
LinkedRditingRange's wordPattern value is used to control the linked editing. The value of the wordPattern is to be provided
by the active LS, otherwize a common Unicode Indentifier  pattern is used to validate the input.

Change-Id: I6291f7136fbdb8f4ee3f067d8b17c74400067676

### DIFF
--- a/org.eclipse.lsp4e.test/src/org/eclipse/lsp4e/test/linkedediting/LinkedEditingTest.java
+++ b/org.eclipse.lsp4e.test/src/org/eclipse/lsp4e/test/linkedediting/LinkedEditingTest.java
@@ -11,6 +11,9 @@
  *******************************************************************************/
 package org.eclipse.lsp4e.test.linkedediting;
 
+import static org.junit.Assert.assertEquals;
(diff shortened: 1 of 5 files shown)


mickaelistria:
This looks good to me. Shall we merge now?

vrubezhny:
It looks like there's a difference in how a wordPattern is processed in VSCode and in this fix. See https://github.com/eclipse/lemminx/pull/1188#discussion_r838759117

Probably we need to make it  to be inline with VCSode way.

mickaelistria:
Thanks!

mickaelistria:
There is one test failing with this PR now.

vrubezhny:
> There is one test failing with this PR now.

I don't think that the test failure was anyhow related to the change. The only problem I found in Jenkins logs is a `ConcurrentModificationException` occurrence, for which I have added a fix to this PR. 

Currently the PR is build with success.

PR #102, opened 2022-04-07
Export org.eclipse.lsp4e.operations.codeactions

rubenporras:
As a continuation of https://github.com/eclipse/lsp4e/pull/83, we need to export org.eclipse.lsp4e.operations.codeactions so that it is possible to register org.eclipse.lsp4e.operations.codeactions.LSPCodeActionMarkerResolution as marker resolution generator for the marker type with the extended attributes.

Otherwise, the new marker will not generate any marker resolution.

### DIFF
--- a/org.eclipse.lsp4e/META-INF/MANIFEST.MF
+++ b/org.eclipse.lsp4e/META-INF/MANIFEST.MF
@@ -45,6 +45,7 @@ Bundle-Localization: plugin
 Bundle-ActivationPolicy: lazy
 Bundle-Activator: org.eclipse.lsp4e.LanguageServerPlugin
 Export-Package: org.eclipse.lsp4e;x-internal:=true,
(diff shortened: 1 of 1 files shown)


rubenporras:
@mickaelistria , if we want to keep the exported API to a minimum (not sure it is an issue as it is flagged as x-internal) I could move the class LSPCodeActionMarkerResolution to a different package which is already exported, like org.eclipse.lsp4e, or move the rest of the classes to a new package like org.eclipse.lsp4e.operations.codeactions.interal, but I do not like this second option that much.

Going over all other classes and making then final or reducing the number of private methods I think is not worthwhile doing, so I would prefer not to do that.

What do you think?

rubenporras:
@mickaelistria, ping :)

What do you think of this PR (see my previous comment)?

mickaelistria:
Let's start with x-internal and we may refactor it later.

PR #105, opened 2022-04-11
Fix hover information widget size

### LINKED ISSUE #104: Hover widget is incorrectly sized
Seems to be Windows related.

Got it on the Qute language server which is part of Quarkus Tools but can be reproduced with github.com/jeffmaury/simplels

On Windows:

image

On Linux:

LSP4E

jeffmaury:
Fixes #104

Signed-off-by: Jeff MAURY <jmaury@redhat.com>
Change-Id: I67e5fadfad6f5d96d79f295617dacd6acfc1ddb7


mickaelistria:
Thank you!

jeffmaury:
What about the second use case (several lines): should we remove the code that limits the size to the computed constraint one ?

PR #107, opened 2022-04-21
Choose 'best' resource when file url is ambiguous

kdvolder:
In Eclipse workspace it is possible that a single file corresponds
to multiple workspace resources. This typically arises in 'multi-project'
setups where both the parent project and its nested children are
imported into the workspace.

This creates all sorts of confusions and subtle bugs.

The logical solution seems to be to deliberatly choose the
'most nested' resource instead of just arbitarily picking the
first one found.

This small PR does just that in LSPEclipseUtils.

This indirectly fixes issues in STS such as:

https://github.com/spring-projects/sts4/issues/769
https://github.com/spring-projects/sts4/issues/747

cc @BoykoAlex


martinlippert:
@mickaelistria What do you think about this one? I think the overall issue to avoid opening the file that is associated with a parent project (in case of multi module maven projects for example) and instead the file "in the real project" makes super sense, but I wonder it you have any suggestions for the implementation?

mickaelistria:
Thanks for this good change!

PR #111, opened 2022-04-29
Java editor LS Quick Assist

BoykoAlex:
1. Support `codeAction/resolve` for edit property
2. LS based Quick Assist processor for Java editor
3. Single undo operation for WorkspaceEdit (UndoManager was missing from the apply operaion)
4. Bugs in `LSPCodeActionQuickAssistProcessor`
  - Can't cache range. Each subsequent invocation of QA in the same doc different location brings up the same cached proposals
  - Don't think caching proposals is a good idea. Computation method can be reentrant but not recursive. Problem is old proposals are never cleared. I might be misunderstanding this bit though...

### DIFF
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/codeactions/LSPCodeActionQuickAssistProcessor.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/codeactions/LSPCodeActionQuickAssistProcessor.java
@@ -51,6 +51,11 @@ public class LSPCodeActionQuickAssistProcessor implements IQuickAssistProcessor 
 
 	private List<LSPDocumentInfo> infos;
 
+	// Data necessary for caching proposals
+	private Object lock = new Object();
+	private IQuickAssistInvocationContext cachedContext;
+	private List<ICompletionProposal> proposals = Collections.synchronizedList(new ArrayList<>());
+
 	private static final ICompletionProposal COMPUTING = new ICompletionProposal() {
 
 		@Override
@@ -99,8 +104,9 @@ public class LSPCodeActionQuickAssistProcessor implements IQuickAssistProcessor 
 
 	@Override
 	public boolean canAssist(IQuickAssistInvocationContext invocationContext) {
-		if (this.infos == null || this.infos.isEmpty()) {
-			infos = getLSPDocumentInfos(invocationContext.getSourceViewer().getDocument());
+		IDocument document = invocationContext.getSourceViewer().getDocument();
+		if (this.infos == null || this.infos.isEmpty() || this.infos.get(0).getDocument() != document) {
+			infos = getLSPDocumentInfos(document);
 			if (infos.isEmpty()) {
 				return false;
 			}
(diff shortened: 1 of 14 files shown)


org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/rename/LSPRenameProcessor.java
```diff
@@ -197,7 +197,7 @@ public Change createChange(IProgressMonitor pm) throws CoreException, OperationC
 			throw new CoreException(
 					new Status(IStatus.ERROR, LanguageServerPlugin.PLUGIN_ID, Messages.rename_processor_required));
 		}
-		return LSPEclipseUtils.toCompositeChange(rename);
+		return LSPEclipseUtils.toCompositeChange(rename, "Rename"); //$NON-NLS-1$
```
mickaelistria, line 200: Shouldn't it be internationalized?

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/codeactions/LSPCodeActionQuickAssistProcessor.java
```diff
@@ -175,25 +167,12 @@ public ICompletionProposal[] computeQuickAssistProposals(IQuickAssistInvocationC
...
-			infos = LanguageServiceAccessor.getLSPDocumentInfosFor(document,
-					LSPCodeActionMarkerResolution::providesCodeActions);
-			ITextSelection selection = (ITextSelection) textEditor.getSelectionProvider().getSelection();
-			try {
-				this.range = new Range(LSPEclipseUtils.toPosition(selection.getOffset(), document),
-						LSPEclipseUtils.toPosition(selection.getOffset() + selection.getLength(), document));
-			} catch (BadLocationException e) {
-				LanguageServerPlugin.logError(e);
-			}
+	private List<LSPDocumentInfo> getLSPDocumentInfos(IDocument document) {
```
mickaelistria, line 170: Couldn't it be static?

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/codeactions/LSPCodeActionQuickAssistProcessor.java
```diff
@@ -138,19 +132,18 @@ public boolean canAssist(IQuickAssistInvocationContext invocationContext) {
...
 		}
 
 		// Get the codeActions
-		CodeActionParams params = prepareCodeActionParams(this.infos);
+		CodeActionParams params = prepareCodeActionParams(infos, invocationContext.getOffset(), invocationContext.getLength());
+		List<ICompletionProposal> proposals = Collections.synchronizedList(new ArrayList<>());
 		List<CompletableFuture<Void>> futures = Collections.emptyList();
 		try {
 			// Prevent infinite recursion by only computing proposals if there aren't any
-			if (proposals.contains(COMPUTING) || proposals.isEmpty()) {
```
mickaelistria, line 152: Why does it have to be removed? IIRC the case described in comment above still applies.

mickaelistria (commented):
This looks good. Some minor comments inline.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/codeactions/LSPCodeActionQuickAssistProcessor.java
BoykoAlex: I didn't understand the comment to be honest. I don't think this method can make recursive calls... I think it can be re-entrant (maybe the comment was referring to re-entrance) and hence there was proposals caching etc. but i didn't see the benefit (might be missing/misunderstanding) of this caching...

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/codeactions/LSPCodeActionQuickAssistProcessor.java
BoykoAlex: Yes, good point

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/rename/LSPRenameProcessor.java
BoykoAlex: Definitely should be internationalized. Done.

BoykoAlex:
@mickaelistria The first 2 comments - adjusted the PR. The third about recursion needs a bit of discussion because i might be missing something...

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/codeactions/LSPCodeActionQuickAssistProcessor.java
mickaelistria: I'm not sure this comment is clear enough, but it could be that the same LSPCodeActionQuickAssistProcessor can be invoked in several threads and that we don't want it to always recompute if there is no need for it. If this guard doesn't cause any problem for your usage, I suggest you keep it and we consider removing it in a separate bug.

BoykoAlex: I have finally understood why proposals were cached. If proposals computation takes over 200ms then once they are computed the same method is invoked later (re-entrance) with the same invocation context object. Proposals need to be wiped out if context object is the same as before. I have added comments in the code.

BoykoAlex:
@mickaelistria should be good to go now hopefully

PR #113, opened 2022-05-02
Upgrade org.eclipse.mylyn.wikitext.* to 3.0.42

rubenporras:
to keep it in sync with
/target-platform-latest/target-platform-latest.target


rubenporras:
To be honest, I do not understand how having a manifest requiring 3.0.40 and a target file providing 3.0.42 can work, so if anyone could explain it, I would really appreciate it ;)
In any case, I think it makes sense to keep it sync, so even though I do not understand the details, I have decided to open this PR.

mickaelistria:
This change is not necessary because 3.0.42 doesn't have substantial change compared to 3.0.40 (which itself had important change compared to previoys 3.0.x).
the `bundle-version="3.0.40"` should be read, as per[ OSGi specification](https://docs.osgi.org/specification/osgi.core/8.0.0/framework.module.html#framework.module.requirebundle) as >= 3.0.40. Since 3.0.42 is a match and there is no particular reason to reject 3.0.40; it's better to stick to 3.0.40 as minimal compatible version.

rubenporras:
Thanks, I understand it now.

PR #115, opened 2022-05-06
Only use Quickfix kind of CodeActions for Quickfixes

BoykoAlex:
There are many various refactorings returned from the LS. However, CodeActions are fetched in the context of showing Quick Fixes (i.e. problem marker hover) then only QuickFix kind of CodeActions are applicable and the rest ignored.


mickaelistria:
I merged too fast! Some tests are now failing, most likely because of that change. I think I'm going to revert it as this is likely incorrect: there is a CodeActionContext that does tell the LS that the request CodeActions are the ones for the given Diagnostics. From here, it's up to the LS to filter the result, not to the client.

mickaelistria:
Reverted as 02dc61e5b784f9d62be8c1702565330d83f62682

BoykoAlex:
@mickaelistria my bad, indeed `CodeActionContext` is thing to use. Please revert my change and sorry about this.

mickaelistria:
No problem ;)

PR #116, opened 2022-05-06
WorkspaceFolders: Fix interpretation of event flags

ahmedneilhussain:
LSP4e implements the LSP Workspace Folders protocol [if supported by the remote back end] using the Eclipse ResourceListener mechanism. Looking at the current implementation and reading the Eclipse source code, it isn't using the eclipse API quite correctly. 

The flags integer for an IResourceDelta seems to be a synthesis of an enum "Kind" (bottom 8 bits) and a bit mask (top 24 bits). The existing LSP4e code tries to detect a project open/close with `delta.getKind() == IResourceDelta.OPEN && project.isAccessible()`.
That won't work, because `OPEN` is a qualifying bitmask flag,  not a kind - `OPEN = 0x4000` but `getKind()` returns `flags & 0xFF`.
The `kind` for an open/close event is a `CHANGE`, but with the `OPEN` flag set.

We also add `PRE_DELETE` to the filter of `ResourceListener` events we are observing: this gives a cleaner behaviour on hard deletes if the underlying language server is also observing the file system, as sometimes delete events are only sent when the file tree is already gone.

### DIFF
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServerWrapper.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServerWrapper.java
@@ -169,13 +169,7 @@ public class LanguageServerWrapper {
 	 */
 	private final @NonNull Map<@NonNull String, @NonNull Runnable> dynamicRegistrations = new HashMap<>();
 	private boolean initiallySupportsWorkspaceFolders = false;
-	private final @NonNull IResourceChangeListener workspaceFolderUpdater = event -> {
-		WorkspaceFoldersChangeEvent workspaceFolderEvent = toWorkspaceFolderEvent(event);
-		if (workspaceFolderEvent == null || (workspaceFolderEvent.getAdded().isEmpty() && workspaceFolderEvent.getRemoved().isEmpty())) {
-			return;
-		}
-		this.languageServer.getWorkspaceService().didChangeWorkspaceFolders(new DidChangeWorkspaceFoldersParams(workspaceFolderEvent));
-	};
+	private final @NonNull IResourceChangeListener workspaceFolderUpdater = new WorkspaceFolderListener();
 
 	/* Backwards compatible constructor */
 	public LanguageServerWrapper(@NonNull IProject project, @NonNull LanguageServerDefinition serverDefinition) {
@@ -502,54 +496,6 @@ public class LanguageServerWrapper {
 		}.schedule();
 	}
 
-	private static final @Nullable WorkspaceFoldersChangeEvent toWorkspaceFolderEvent(
-			IResourceChangeEvent e) {
-
-		// If a project delete then the delta is null, but we get the project in the top-level resource
-		WorkspaceFoldersChangeEvent wsFolderEvent = new WorkspaceFoldersChangeEvent();
-		if (e.getType() == IResourceChangeEvent.PRE_DELETE) {
-			final IResource resource = e.getResource();
-			if (resource instanceof IProject) {
-				wsFolderEvent.getRemoved()
-						.add(LSPEclipseUtils.toWorkspaceFolder((IProject)resource));
-				return wsFolderEvent;
-			} else {
-				return null;
-			}
-		} else if (e.getType() != IResourceChangeEvent.POST_CHANGE) {
-			return null;
-		}
-
-		// Use the visitor implementation to extract the low-level detail from delta
-		try {
-			e.getDelta().accept(delta -> {
-				if (delta.getResource().getType() == IResource.PROJECT) {
-					IProject project = (IProject) delta.getResource();
-					final WorkspaceFolder wsFolder = LSPEclipseUtils.toWorkspaceFolder(project);
-					if ((delta.getKind() == IResourceDelta.ADDED || (delta.getKind() == IResourceDelta.CHANGED
-							&& (delta.getFlags() & IResourceDelta.OPEN) == IResourceDelta.OPEN))
-							&& project.isAccessible()
-							&& wsFolder != null && !wsFolder.getUri().isEmpty()) {
-						wsFolderEvent.getAdded().add(wsFolder);
-					} else if ((delta.getKind() == IResourceDelta.REMOVED
-							|| (delta.getKind() == IResourceDelta.CHANGED
-									&& (delta.getFlags() & IResourceDelta.OPEN) == IResourceDelta.OPEN))
-									&& !project.isAccessible()
-									&& wsFolder != null && !wsFolder.getUri().isEmpty()) {
-						wsFolderEvent.getRemoved().add(wsFolder);
-					}
-					// TODO: handle renamed/moved (on filesystem)
-				}
-				return delta.getResource().getType() == IResource.ROOT;
-			});
-		} catch (CoreException ex) {
-			LanguageServerPlugin.logError(ex);
-		}
-		if (wsFolderEvent.getAdded().isEmpty() && wsFolderEvent.getRemoved().isEmpty()) {
-			return null;
-		}
-		return wsFolderEvent;
-	}
 
 	/**
 	 * Check whether this LS is suitable for provided project. Starts the LS if not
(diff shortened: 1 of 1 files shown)


org.eclipse.lsp4e.test/src/org/eclipse/lsp4e/test/WorkspaceFoldersTest.java
```diff
@@ -0,0 +1,209 @@
+package org.eclipse.lsp4e.test;
```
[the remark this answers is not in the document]
ahmedneilhussain: Hope this is ok

mickaelistria:
There is now 1 failing test (regression) that is related to that change, see https://ci.eclipse.org/lsp4e/job/lsp4e-github/job/PR-116/2/display/redirect

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServerWrapper.java
```diff
@@ -495,25 +496,47 @@ public IStatus runInWorkspace(IProgressMonitor monitor) throws CoreException {
...
+		if (e.getType() == IResourceChangeEvent.PRE_DELETE) {
+			final IResource resource = e.getResource();
+			if (resource instanceof IProject) {
+				wsFolderEvent.getRemoved()
+						.add(LSPEclipseUtils.toWorkspaceFolder((IProject)resource));
+				return wsFolderEvent;
+			} else {
+				return null;
+			}
+		} else if (e.getType() != IResourceChangeEvent.POST_CHANGE) {
```
rubenporras, line 519: could we do this at the very beginning of the method, so that if this condition is met, we do not need to create an WorkspaceFoldersChangeEvent that we will never use?

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServerWrapper.java
```diff
@@ -495,25 +496,47 @@ public IStatus runInWorkspace(IProgressMonitor monitor) throws CoreException {
...
-					if ((delta.getKind() == IResourceDelta.ADDED || delta.getKind() == IResourceDelta.OPEN) && project.isAccessible()) {
-						wsFolderEvent.getAdded().add(LSPEclipseUtils.toWorkspaceFolder((IProject)delta.getResource()));
-					} else if (delta.getKind() == IResourceDelta.REMOVED || (delta.getKind() == IResourceDelta.OPEN && !project.isAccessible())) {
-						wsFolderEvent.getRemoved().add(LSPEclipseUtils.toWorkspaceFolder((IProject)delta.getResource()));
+					IProject project = (IProject) delta.getResource();
+					final WorkspaceFolder wsFolder = LSPEclipseUtils.toWorkspaceFolder(project);
+					if ((delta.getKind() == IResourceDelta.ADDED || (delta.getKind() == IResourceDelta.CHANGED
+							&& (delta.getFlags() & IResourceDelta.OPEN) == IResourceDelta.OPEN))
+							&& project.isAccessible()
+							&& wsFolder != null && !wsFolder.getUri().isEmpty()) {
```
rubenporras, line 532: getUri() can return null, so we need another condition: !wsFolder.getUri() != null

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServerWrapper.java
```diff
@@ -495,25 +496,47 @@ public IStatus runInWorkspace(IProgressMonitor monitor) throws CoreException {
...
+						wsFolderEvent.getAdded().add(wsFolder);
+					} else if ((delta.getKind() == IResourceDelta.REMOVED
+							|| (delta.getKind() == IResourceDelta.CHANGED
+									&& (delta.getFlags() & IResourceDelta.OPEN) == IResourceDelta.OPEN))
+									&& !project.isAccessible()
+									&& wsFolder != null && !wsFolder.getUri().isEmpty()) {
```
rubenporras, line 538: getUri() can return null, so we need another condition: !wsFolder.getUri() != null

rubenporras, line 538: this condition and the previous one are almost identical, could we refactor to a private method with a boolean flag `isAccessbile`?

mickaelistria:
Sorry for my previous message, build failure was not related.

ahmedneilhussain:
> Sorry for my previous message, build failure was not related.

No problem - it was in a suspiciously related area of functionality!

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServerWrapper.java
ahmedneilhussain: It would be nice to break up the code a bit, but this class is already rather large, and so is e.g. `LSPEclipseUtils`. What would you say to replacing the lambda `workspaceFolderUpdater` at line 172 with a proper class implementing `IResourceChangeListener`, then all of this logic could be broken up and put there?

mickaelistria:
@ahmedneilhussain all those are good ideas. However, please create separate pull request for each of them as much as possible. Smaller PRs allow faster progress.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServerWrapper.java
rubenporras: I am also a relative newcomer to programming on the Eclipse platform. Normally I would code as by API and remove the guard, since it should never be hit. According to git history @mickaelistria  introduced the guard, so maybe he has some insight why he added it in the first place.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServerWrapper.java
rubenporras: I have no strong opinion about refactoring to a nested class. Regarding the null check, I think you need it in any case because when a project is deleted the URI will be null. I think this can happen regardless of the de-registration in `stop()`.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServerWrapper.java
mickaelistria: > if the subscription at line 499 is honoured by Eclipse it should not to be needed, is that correct?

Correct. However, it's always safer to re-validate important assumptions in the code. Adding a check has no significant impact on the performance but has a very positive impact on our capacity to change the code safely, so overall it's beneficial.

mickaelistria:
@rubenporras @ahmedneilhussain Do you think this pull request is fine to merge in this current state, or are there some necessary changes left to implement ? (and refactoring and other stuff can be considered in further PRs)

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServerWrapper.java
mickaelistria: Not according to LSP specification. So a LS that does return null here is not compliant with the spec.

ahmedneilhussain:
> @ahmedneilhussain all those are good ideas. However, please create separate pull request for each of them as much as possible. Smaller PRs allow faster progress.

Sorry for slow response guys, I had today off officially but decided it was only a little bit of work just to tidy the code into a separate class.
What do you guys think @mickaelistria @rubenporras ? I think this is much cleaner and will be much easier to read and follow in future, particularly if we need to add support for move/rename (which I hope to be able to offer at some point). The logic for decoding and translating the eclipse resource change flags is particularly hairy and is a lot easier to read with meaningfully named helper methods.
Ideally I'd have the whole thing as a freestanding class but I wasn't in a rush to expose the `LanguageServerWrapper.languageServer` field just for this.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServerWrapper.java
ahmedneilhussain: Thanks. Noted for future!

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServerWrapper.java
ahmedneilhussain: Trivial to do, so added the guard anyway as part of the tidy up.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServerWrapper.java
ahmedneilhussain: Added as part of tidy up (see below).

mickaelistria:
Thanks!

mickaelistria:
I think it was already fine for a few interations, so I merged it so it can be easily tested and further improvements are still welcome in further PRs.

PR #117, opened 2022-05-09
Support willSaveWaitUntil

rubenporras:
@mickaelistria ,

we would like to implement Format-on-Save functionality.  After doing some research, we thought that we can use the willSaveWaitUntil (https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_willSaveWaitUntil) to implement Format-on-Save as an action we would execute on the server right before saving the document. So we would like to add support for it in LSP4E.

Do you think this is a good way of implementing Format-on-Save on LSP4E? 

Assuming it is, what do you think of this implementation? If not, what would be the alternative?

My current thoughts on the actual implementation are:
1. I am not sure if `documentSaved` is the right place to call `willSaveWaitUntil`. It does the trick (The call is done when the document is saved, provided the capability is registered), but it looks like too late, as one cannot know why the document is saved, so I just use TextDocumentSaveReason.Manual, even though it could be delay/focus-out.
2. I would like to wait a little bit before calling `didSave`, to assure ordering, but I am not sure how to do that. Maybe I can just wait for the upcoming patches in this area, which are being prepared by others, maybe I can call `get` on the future with a couple of seconds timeout?
3. I am not sure if I need to implement some handling for the List<TextEdit> which are coming from server, currently I see that the server returns some edits, but I do not see them applied in the editor. Do you know that?

The points I am more interested about are if you think willSaveWaitUntil is a valid approach and 1). 2) is not as important for the time being. If you know about 3), that would be great, but if not, I think will sooner or later figure it out myself.

Regards


mickaelistria:
> After doing some research, we thought that we can use the willSaveWaitUntil

That can indeed be an interesting way to achieve it. And anyway, support for missing LSP operations is always welcome!

> I am not sure if documentSaved is the right place to call willSaveWaitUntil. It does the trick (The call is done when the document is saved, provided the capability is registered), but it looks like too late, as one cannot know why the document is 
saved, so I just use TextDocumentSaveReason.Manual, even though it could be delay/focus-out.

Yes, the issue becomes that the document is saved on Eclipse side; but then it's modified as part of the save, and I don't think it's saved again. So it may indeed be too late. We should find some pre-save hook where to plug that, but I don't know if there is a good API for that in Platform.

> I would like to wait a little bit before calling didSave, to assure ordering, but I am not sure how to do that.

You can most likely keep the future that you generate after `willSaveWaitUntil` and invoke `.thenCombineAsync(() -> languageServer.didSave(saveParams))` to ensure the order. 

> Maybe I can just wait for the upcoming patches in this area, which are being prepared by others, maybe I can call get on the future with a couple of seconds timeout?

I think it's better if you try to remediate it now and not wait for others.

> I am not sure if I need to implement some handling for the List which are coming from server, currently I see that the server returns some edits, but I do not see them applied in the editor. Do you know that?

You'll need to process the output of the request with something like `LSPEclipseUtils.applyTextEdit(...)` for the change to be applied to the client.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/DocumentContentSynchronizer.java
```diff
@@ -163,6 +169,42 @@ private boolean createChangeEvent(DocumentEvent event) {
...
+				return ls.getTextDocumentService().willSaveWaitUntil(params).get(500, TimeUnit.MILLISECONDS);
+			} catch (InterruptedException | ExecutionException e) {
+				LanguageServerPlugin.logError(e);
+				return Collections.emptyList();
+			} catch (TimeoutException e) {
+				return Collections.emptyList();
+			}
+		});
+		try {
+			willSaveUntilFuture.thenAccept(edits -> LSPEclipseUtils.applyEdits(document, edits)).get(500, TimeUnit.MILLISECONDS);
```
mickaelistria, line 202: I think it reads and overall performs better if you run
```
languageServerWrapper.getInitializedServer().thenComposeAsync(ls -> ls.getTextDocumentService().willSaveWaitUntil(params).thanApplyAsync(edits -> LSPEclipseUtils.applyEdits(document, params))
```
But this is indeed prone to race conditions (eg `save` may be invoked before `willSave...` is queried. So the willSave future should probably be stored on the listener and the actual `save` should take care of first waiting for the willSave future if necessary.

rubenporras: By looking at `org.eclipse.core.internal.filebuffers.FileStoreFileBuffer.commit(IProgressMonitor, boolean)`, the method will fire always first `fireStateChanging`, which iterates through the listeners and calls the handlers. As far as I understand the iterator waits for each handler to finish. Then `fireDirtyStateChanged` is called. So in this code path, `save` will always be called after `willSave...` is finish. Do you have other code paths in mind?

mickaelistria: ok, got it. It's indeed fine to make the listener blocking then.

mickaelistria: Please try the change mentioned in earlier comment anyway if possible, it makes code simpler and allows to only join on a single future and thus has a better control on the timeout being a "global" timeout.

rubenporras: In addition to my last comment, `save...` should not wait for the willSave future, the work doing in that future must be finished by the time the buffer is saved (otherwise the buffer would be saved, the work would modify the buffer again in the background and the buffer would get dirty). Since we call `DocumentContentSynchronizer.documentSaved` on the DirtyStateChanged event, it is too late to wait on documentSaved.

I will take a look at the comment on how to rewrite it for better readability and performance, I have briefly looked at it, and the code
`languageServerWrapper.getInitializedServer()
		.thenComposeAsync(ls -> ls.getTextDocumentService().willSaveWaitUntil(params))
		.thanApplyAsync(edits -> LSPEclipseUtils.applyEdits(document, edits));`
has the compiler error

The method thanApplyAsync((<no type> edits) -> {}) is undefined for the type CompletableFuture<List<TextEdit>>.
I need to understand that one to be able to submit a newer version, and currently I do not get it yet. I find this API quite difficult to work with :(.

mickaelistria: > n addition to my last comment, save... should not wait for the willSave future, the work doing in that future must be finished by the time the buffer is saved

Yes, the "applyEdits" would need to happen before.

> The method thanApplyAsync(( edits) -> {}) is undefined for the type CompletableFuture<List>.

Sorry I meant `.thenAcceptAsync()`.

mickaelistria:
Do you think you could add a test case for this new feature?

org.eclipse.lsp4e/src/org/eclipse/lsp4e/DocumentContentSynchronizer.java
rubenporras: thanks, that worked

mickaelistria:
I think a test requires that you tweak the MockLanguageServer so it enabled the willSave capability and return a mock edit on willSave (eg replacing all file content with `foo`); and then the test will just have to open a file and save it and verify its content is replaced before save.

rubenporras:
From my point of view, the change would be ready.

mickaelistria:
Thanks for adding a test!

PR #118, opened 2022-05-15
Adapt to LSP4J 0.13

mickaelistria:


org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/symbols/LSPSymbolInWorkspaceDialog.java
```diff
@@ -185,4 +189,26 @@ protected Control createExtendedContentArea(Composite parent) {
 		return null;
 	}
 
+	private static List<WorkspaceSymbol> toWorkspaceSymbols(List<? extends SymbolInformation> source) {
+		return source == null ?
+				List.of() :
+				source.stream().map(LSPSymbolInWorkspaceDialog::toWorkspaceSymbol).collect(Collectors.toList());
```
rubenporras, line 195: what if `toWorkspaceSymbol` returns null? Should we not filter null objects before collecting to a list?

org.eclipse.lsp4e.debug/META-INF/MANIFEST.MF
```diff
@@ -9,9 +9,9 @@ Require-Bundle: org.eclipse.ui,
  org.eclipse.debug.core,
  org.eclipse.debug.ui,
  org.eclipse.jface,
- org.eclipse.lsp4j.jsonrpc;bundle-version="[0.12.0,0.13.0)",
- org.eclipse.lsp4j.jsonrpc.debug;bundle-version="[0.12.0,0.13.0)",
- org.eclipse.lsp4j.debug;bundle-version="[0.12.0,0.13.0)",
+ org.eclipse.lsp4j.jsonrpc;bundle-version="[0.13.0,0.14.0)",
```
rubenporras, line 12: I am not sure if Xtext depends on LSP4J 0.12, so this might break our current product until they have compatible release. Do you plan to merge this before doing a release? Did you check with the Xtext project?

mickaelistria: XText and LSP4E can use different versions of LSP4J. The question is more whether XText has the right dependency ranges.
I don't plan to check with XText as I don't use it and never have it installed nor touched the code. So I think you'd do a better job than me at coordinating this LSP4E change with XText ;)

rubenporras: Xtext has the ranges [0.12.0,0.13.0), so I think the bundles will resolve. I am more concerned about some of the breaking changes in LSP4J between 0.12 and 0.13.
I can help Xtext dealing with that, if they do not have resources, no problem with that, but it would still be better to merge this after doing an LSP4E release, otherwise there will be some stress on our side.

cdietrich: see https://github.com/eclipse/xtext-core/compare/cd_lsp4j0130
there is no release out yet
so i cannot pr on xtext side yet

jonahgraham:
LSP4J release is out - https://github.com/eclipse/lsp4j/releases/tag/v0.13.0 

p2 url: https://download.eclipse.org/lsp4j/updates/releases/0.13.0/

PR #121, opened 2022-05-18
Concurrency fixes

ahmedneilhussain:
As discussed on the lsp4e-dev mailing list, these are some fixes we made internally to try and address some weird behaviour we used to get, particularly under load. The existing implementation submits document update events to a thread pool where they may get scheduled in any order. This can cause the client's and server's view of the state of a document to diverge, resulting in phantom error markers and so on.

### DIFF
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LSPEclipseUtils.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LSPEclipseUtils.java
@@ -413,22 +413,27 @@ public class LSPEclipseUtils {
 			return;
 		}
 
-		IDocumentUndoManager manager = DocumentUndoManagerRegistry.getDocumentUndoManager(document);
-		if (manager != null) {
-			manager.beginCompoundChange();
-		}
-
 		MultiTextEdit edit = new MultiTextEdit();
 		for (TextEdit textEdit : edits) {
 			if (textEdit != null) {
 				try {
 					int offset = LSPEclipseUtils.toOffset(textEdit.getRange().getStart(), document);
 					int length = LSPEclipseUtils.toOffset(textEdit.getRange().getEnd(), document) - offset;
+					if (length < 0) {
+ 						// Must be a bad location: we bail out to avoid corrupting the document.
+ 						LanguageServerPlugin.logInfo("Invalid location information found during formatting"); //$NON-NLS-1$
+ 						return;
+ 					}
 					edit.addChild(new ReplaceEdit(offset, length, textEdit.getNewText()));
 				} catch (BadLocationException e) {
 					LanguageServerPlugin.logError(e);
 				}
 			}
+		}
+
+		IDocumentUndoManager manager = DocumentUndoManagerRegistry.getDocumentUndoManager(document);
+		if (manager != null) {
+			manager.beginCompoundChange();
 		}
 		try {
 			RewriteSessionEditProcessor editProcessor = new RewriteSessionEditProcessor(document, edit,
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/format/LSPFormatter.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/format/LSPFormatter.java
@@ -21,6 +21,7 @@ import java.util.concurrent.CompletableFuture;
 import org.eclipse.jdt.annotation.NonNull;
 import org.eclipse.jface.preference.IPreferenceStore;
 import org.eclipse.jface.text.BadLocationException;
+import org.eclipse.jface.text.Document;
 import org.eclipse.jface.text.IDocument;
 import org.eclipse.jface.text.ITextSelection;
 import org.eclipse.lsp4e.LSPEclipseUtils;
@@ -53,11 +54,25 @@ public class LSPFormatter {
 		}
 		// TODO consider a better strategy for that, maybe iterate on all LS until one gives a result
 		LSPDocumentInfo info = infos.iterator().next();
-		try {
-			return requestFormatting(info, textSelection);
-		} catch (BadLocationException e) {
-			LanguageServerPlugin.logError(e);
-		}
+		final long currentModificationStamp = document instanceof Document
+ 				? ((Document) document).getModificationStamp()
+ 				: -1;
+ 		// Only attempt formatting if the server is up-to-date with the latest changes
+ 		if (info.getDocumentModificationStamp() == currentModificationStamp) {
+ 			try {
+ 				return requestFormatting(info, textSelection).thenApply(edits -> {
+ 					if (document instanceof Document) {
+ 						if (currentModificationStamp != ((Document) document).getModificationStamp()) {
+ 							// The document has been updated since the formatting request
+ 							return null;
+ 						}
+ 					}
+ 					return edits;
+ 				});
+ 			} catch (BadLocationException e) {
+ 				LanguageServerPlugin.logError(e);
+ 			}
+ 		}
 		return CompletableFuture.completedFuture(Collections.emptyList());
 	}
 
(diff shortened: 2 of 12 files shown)


mickaelistria:
This change is too big to be reviewed property, also it's conflicting with master. Can you please rebase it and create smaller individual changes describing the particular problem they fix/the particular improvement they add?

mickaelistria:
Actually, the grain of commits isn't too bad. However, it would be nice if you could re-organize them to squash the test change together with the fix (eg edit ordering test goes in 1st patch and so on).

mickaelistria:
Also, I'm curious: how likely is it that the document listener receives events in the wrong order? I know that starting many futures is likely to cause them in the wrong order, but where they "created" in the right order? If yes, we may not need to deal too much with timestamps, and can just trust the `.then...` methods on Future to run things in the same order as events are received.
Did you find a case where events got received in the wrong order? as document edition is usually sequential, I'd be a bit surprised.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/format/LSPFormatter.java
```diff
@@ -45,34 +45,35 @@ public void applyEdits(IDocument document, List<? extends TextEdit> edits) {
...
-		// use range formatting if standard formatting is not supported or text is selected
-		if (capabilities != null
-				&& isDocumentRangeFormattingSupported(capabilities)
-				&& (!isDocumentFormattingSupported(capabilities)
-						|| textSelection.getLength() != 0)) {
+		// use range formatting if standard formatting is not supported or text is
+		// selected
+		if (capabilities != null && isDocumentRangeFormattingSupported(capabilities)
+				&& (!isDocumentFormattingSupported(capabilities) || textSelection.getLength() != 0)) {
 			DocumentRangeFormattingParams params = new DocumentRangeFormattingParams();
```
rubenporras, line 77: this is only a change in format, right? If so I would revert it to keep the diff clean

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LSPEclipseUtils.java
```diff
@@ -397,23 +397,28 @@ public static void applyEdits(IDocument document, List<? extends TextEdit> edits
...
 		MultiTextEdit edit = new MultiTextEdit();
 		for (TextEdit textEdit : edits) {
 			if (textEdit != null) {
 				try {
 					int offset = LSPEclipseUtils.toOffset(textEdit.getRange().getStart(), document);
 					int length = LSPEclipseUtils.toOffset(textEdit.getRange().getEnd(), document) - offset;
+					if (length < 0) {
+ 						// Must be a bad location: we bail out to avoid corrupting the document.
+ 						LanguageServerPlugin.logInfo("Invalid location information found during formatting"); //$NON-NLS-1$
+ 						return;
```
rubenporras, line 409: maybe just 
throw new BadLocationException("Negative length during formatting"); //$NON-NLS-1$
?

rubenporras (commented):
Thanks for the PR.
I do not currently have time to do a proper review, but I can still leave some minor comments.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/format/LSPFormatter.java
ahmedneilhussain: Yes, fair enough. I think I might have run auto-format on the file but with subtly different settings.

Is there a recommended `formatter.xml` for the project?

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LSPEclipseUtils.java
```diff
@@ -851,8 +856,10 @@ public static URI toUri(File file) {
 		}
 	}
 
-	@NonNull
 	public static WorkspaceFolder toWorkspaceFolder(@NonNull IProject project) {
+		if (project.getLocationURI() == null) {
+			return null;
+		}
 		WorkspaceFolder folder = new WorkspaceFolder();
```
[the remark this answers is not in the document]
ahmedneilhussain: Apologies, this is a fragment of work for https://github.com/eclipse/lsp4e/pull/116 escaped into the wrong changes. It's not needed now anyway.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LSPEclipseUtils.java
ahmedneilhussain: I don't think it makes any difference at the moment, since this method works atomically and none of the edits will get applied if one of them causes an error. None of the current clients of this method would behave differently, but maybe throwing an exception would be more future-proof, as it might allow future consumers of it to bail out of other actions if it flagged up an inconsistency between client and server?

ahmedneilhussain:
> Also, I'm curious: how likely is it that the document listener receives events in the wrong order? I know that starting many futures is likely to cause them in the wrong order, but where they "created" in the right order? If yes, we may not need to deal too much with timestamps, and can just trust the `.then...` methods on Future to run things in the same order as events are received.
> Did you find a case where events got received in the wrong order? as document edition is usually sequential, I'd be a bit surprised.

We definitely had problems with events getting received out of order, but of course such problems are hard to reproduce consistently. Certainly the `.then...` methods avoid the problem we had before when the futures were just compounded in parallel on the original `didOpenFuture` and could then get dispatched in any order from the thread pool.

Yes, in theory this change should suffice for changes received via `IDocumentListener` calls, as those should all be on the event thread and thus serialise. I'm not sure the save/close events, which come from a different interface, have the same guarantee. We had particular problems with formatting, and in particular our attempt to achieve format-on-save.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LSPEclipseUtils.java
rubenporras: Yes, I agree it is not very important. I think it would be nicer code, but probably just a matter of personal preference.

rubenporras:
I do not know about other cases, but I think format-on-save.you would not need it. If you use the willSaveWaitUntil request, for which I added support a few commits ago, the events should como in order, shouldn't they? The chain of events would be:

1) notifcation before save will send willSaveWaitUntil 
2) a list of edits is received as response
3) LSP4E applies them
4) editor is saved

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LSPEclipseUtils.java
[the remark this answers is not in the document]
rubenporras: Thanks

ahmedneilhussain:
> I do not know about other cases, but I think format-on-save.you would not need it. If you use the willSaveWaitUntil request, for which I added support a few commits ago, the events should como in order, shouldn't they? 

Yes, for our purposes we should be able to use that (we'll have to change our language server to do it, though, and not all LS support it).

For the more general point about whether we are now excessively cautious, I was looking through our commit history [our changes were made more organically over time: the PR collapses some incrementally-made fixes together] and one of them was titled _Ensure didSave does not overtake didChange_ - we found this to be a problem in practice on Windows particularly.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LSPEclipseUtils.java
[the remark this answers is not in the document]
ahmedneilhussain: Ugh sorry @rubenporras my rebase yesterday did not include the commit backing out this change. Now done https://github.com/eclipse/lsp4e/commit/5ff09816b579f9b3e5bb9bd7f9bf7625f8638f81

mickaelistria:
I see the 1st patch does a lot: chain futures, deal with documentationModificationStamp and more. It makes it hard to identify what's necessary and which problem it fixes. Would it be possible to separate it into smaller-grain and more focused commits (there can be multiple PRs as well).
I think we can probably achieve a similar result without those modification stamps, and I believe the Format action shouldn't be responsible for checking whether the document is valid. At least, the formatting issue seems to be worth being tracked in a separate issue/commit/PR.

ahmedneilhussain:
As agreed above have split these changes apart and submitted the first logical chunk as https://github.com/eclipse/lsp4e/pull/138

Closing.

PR #127, opened 2022-05-24
Refactoring: add support for the platform's preview UI

ahmedneilhussain:
After wading through the eclipse code we found that this change allowed the diffs to display properly in the refactoring preview dialog:

image

### DIFF
--- a/org.eclipse.lsp4e.test/src/org/eclipse/lsp4e/test/rename/LSPTextChangeTest.java
+++ b/org.eclipse.lsp4e.test/src/org/eclipse/lsp4e/test/rename/LSPTextChangeTest.java
@@ -19,6 +19,7 @@ import java.nio.file.Files;
 import org.eclipse.core.resources.IFile;
 import org.eclipse.core.resources.IProject;
 import org.eclipse.core.runtime.NullProgressMonitor;
+import org.eclipse.jface.text.IDocument;
 import org.eclipse.lsp4e.LSPEclipseUtils;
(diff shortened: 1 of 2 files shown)


ahmedneilhussain:
The summary in the list is obviously not very helpful. I did some digging and it seems clients can override this themselves. From https://help.eclipse.org/2020-09/nftopic/org.eclipse.platform.doc.isv/reference/api/org/eclipse/ltk/ui/refactoring/TextEditChangeNode.html

A special preview node to represent TextEditBasedChanges in the preview tree.
This class should be subclassed by clients wishing to provide language aware substructures for special TextEditBasedChanges. The preview infrastructure accesses to preview node for a TextEditBasedChange by asking the change for an adapter of type TextEditChangeNode. If no adapter is returned, this default implementation will be used to present TextEditBasedChange in the preview tree.

ahmedneilhussain:
So I think it is "just"  a case of creating appropriate subclasses and wiring them in via adapters. Some of this can certainly be done outside of lsp4e by the consuming plugin, but there might need to be a bit of support added so that it can be done without multiple languages interfering with one another. Happy to research this further at some point, but my boss just wanted the preview diffs initially so it's languishing somewhere in my backlog of things to experiment with in the medium term.

mickaelistria:
Thanks, in order to make build happy just after a recent release, can you please increase the version of org.eclipse.lsp4e bundle by +0.0.1 ?

PR #132, opened 2022-05-25
Fix for IllegalStateException: Duplicate RPC method runInTerminal.

vrubezhny:
There is a recent commit is LSP4J - Move runInTerminal #620 (eclipse/lsp4j@bdabc77) - that moves `runInTerminal` method
declaration from IDebugProtocolServer to IDebugProtocolClient in LSP4J.
So an according change is expected to be made in LSP4E's DSPDebugTarget as it implements IDebugProtocolClient, because until it's not,
we'll have runInTerminal to appear twice when finding for the RPC methods, which generates the mentioned `IllegalStateException:
Duplicate RPC method runInTerminal.` exception preventing a debug launch from being started.

Change-Id: I487f9f492c622601f3eb047c6016d63141877c55

### DIFF
--- a/org.eclipse.lsp4e.debug/src/org/eclipse/lsp4e/debug/debugmodel/DSPDebugTarget.java
+++ b/org.eclipse.lsp4e.debug/src/org/eclipse/lsp4e/debug/debugmodel/DSPDebugTarget.java
@@ -79,7 +79,6 @@ import org.eclipse.lsp4j.debug.services.IDebugProtocolClient;
 import org.eclipse.lsp4j.debug.services.IDebugProtocolServer;
 import org.eclipse.lsp4j.jsonrpc.Launcher;
 import org.eclipse.lsp4j.jsonrpc.MessageConsumer;
-import org.eclipse.lsp4j.jsonrpc.services.JsonRequest;
 import org.eclipse.lsp4j.jsonrpc.validation.ReflectiveMessageValidator;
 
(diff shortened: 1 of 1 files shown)


mickaelistria:
Thanks! Let's merge and release now!
By the way, Change-Id and Signed-Off-By are not required, you can forget about them ;)

vrubezhny:
> Thanks! Let's merge and release now! By the way, Change-Id and Signed-Off-By are not required, you can forget about them ;)

Change-Id is added automatically (by a hook) - so it'll continue to be added in my future commits. 
And I see nothing bad in adding a Singed-Off-By - it's just a single option in `git commit` command -honestly, I don't see any reason in getting rid of it.

Good that the issue is caught and fixed.

akurtakov:
> > Thanks! Let's merge and release now! By the way, Change-Id and Signed-Off-By are not required, you can forget about them ;)
> 
> Change-Id is added automatically (by a hook) - so it'll continue to be added in my future commits. And I see nothing bad in adding a Singed-Off-By - it's just a single option in `git commit` command -honestly, I don't see any reason in getting rid of it.

Change-Id is added by a local hook on your machine so you can remove that one.
All I had to do to stop EGit from creating it was to remove 
```
[gerrit]
	createchangeid = true
```
section from gitreponame/.git/config file .
> 
> Good that the issue is caught and fixed.

Issue #123, opened 2022-05-19
Marker resolutions "Computing..." not being updated automatically


BoykoAlex:
Due to a bug report that we got for the Spring tooling (https://github.com/spring-projects/sts4/issues/593) I analyzed the marker resolution mechanism a bit to see what is going on here. To me it seems like the mechanism that is used in  LSPCodeActionMarkerResolution.checkMarkerResoultion  at the end first sets the COMPUTING attribute to the marker, then triggers the computation of the code actions for the marker via the language server - and returns after this computation completed (or times out).  Then, the marker attribute is changed from the COMPUTING to the actions in an async action. That results is that on first invocation, the marker resolutions are most likely "COMPUTING..." (because the async "set actions to marker" hasn't beed executed yet) and that "COMPUTING..." doesn't get changed on the UI. You have to invoke the quick fix action again. Then it shows up the right actions, since they got added to the marker in the meantime.  I thought about fixing this my including the "set actions to marker" future in the timeout instead of the codeAction future, but I am worried about this not being executed if a timeout occurs on one of the codeAction futures.  Any thoughts on this?  Or should we instead update the quick fix popup as soon as actions are set to a marker - in case that is possible?

Due to a bug report that we got for the Spring tooling (https://github.com/spring-projects/sts4/issues/593) I analyzed the marker resolution mechanism a bit to see what is going on here. To me it seems like the mechanism that is used in

LSPCodeActionMarkerResolution.checkMarkerResoultion

at the end first sets the COMPUTING attribute to the marker, then triggers the computation of the code actions for the marker via the language server - and returns after this computation completed (or times out).

Then, the marker attribute is changed from the COMPUTING to the actions in an async action. That results is that on first invocation, the marker resolutions are most likely "COMPUTING..." (because the async "set actions to marker" hasn't beed executed yet) and that "COMPUTING..." doesn't get changed on the UI. You have to invoke the quick fix action again. Then it shows up the right actions, since they got added to the marker in the meantime.

I thought about fixing this my including the "set actions to marker" future in the timeout instead of the codeAction future, but I am worried about this not being executed if a timeout occurs on one of the codeAction futures.

Any thoughts on this?

Or should we instead update the quick fix popup as soon as actions are set to a marker - in case that is possible?

BoykoAlex:
[...]

PR #133, opened 2022-05-26
Update quick fixes in the UI once computed

BoykoAlex:
Adresses #123 

Draft of a possible solution. Reflection pieces can be substituted by the new API added to eclipse platform ui text. However, the first step is to check whether this is something we may use as starting point to carve the final solution.

cc: @martinlippert @mickaelistria 

**Note:** Changes to `LSPCodeActionQuickAssistProcessor` are unrelated to the issue but i feel they are common sense fix to the existing implementation.

### DIFF
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/codeactions/LSPCodeActionMarkerResolution.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/codeactions/LSPCodeActionMarkerResolution.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2016-2017 Red Hat Inc. and others.
+ * Copyright (c) 2016-2022 Red Hat Inc. and others.
  * This program and the accompanying materials are made
  * available under the terms of the Eclipse Public License 2.0
  * which is available at https://www.eclipse.org/legal/epl-2.0/
@@ -12,6 +12,8 @@
 package org.eclipse.lsp4e.operations.codeactions;
 
 import java.io.IOException;
+import java.lang.reflect.Field;
+import java.lang.reflect.Method;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
@@ -27,6 +29,15 @@ import org.eclipse.core.resources.IResource;
 import org.eclipse.core.runtime.CoreException;
 import org.eclipse.jdt.annotation.NonNull;
 import org.eclipse.jface.resource.JFaceResources;
+import org.eclipse.jface.text.AbstractInformationControlManager;
+import org.eclipse.jface.text.ITextHover;
+import org.eclipse.jface.text.ITextViewer;
+import org.eclipse.jface.text.ITextViewerExtension2;
+import org.eclipse.jface.text.TextViewer;
+import org.eclipse.jface.text.contentassist.ContentAssistant;
+import org.eclipse.jface.text.quickassist.IQuickAssistAssistant;
+import org.eclipse.jface.text.quickassist.QuickAssistAssistant;
+import org.eclipse.jface.text.source.ISourceViewerExtension3;
 import org.eclipse.lsp4e.LSPEclipseUtils;
 import org.eclipse.lsp4e.LanguageServerPlugin;
 import org.eclipse.lsp4e.LanguageServersRegistry;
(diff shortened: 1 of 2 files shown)


org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/codeactions/LSPCodeActionMarkerResolution.java
```diff
@@ -147,31 +161,74 @@ private void checkMarkerResoultion(IMarker marker) throws IOException, CoreExcep
...
+		try {
+			// Quick assist proposals popup case
+			if (textViewer instanceof ISourceViewerExtension3) {
+				IQuickAssistAssistant quickAssistant = ((ISourceViewerExtension3)textViewer).getQuickAssistAssistant();
+					Field f = QuickAssistAssistant.class.getDeclaredField("fQuickAssistAssistantImpl"); //$NON-NLS-1$
+					f.setAccessible(true);
+					ContentAssistant ca = (ContentAssistant) f.get(quickAssistant);
+					Method m = ContentAssistant.class.getDeclaredMethod("isProposalPopupActive"); //$NON-NLS-1$
+					m.setAccessible(true);
+					boolean isProposalPopupActive = (Boolean) m.invoke(ca);
```
mickaelistria, line 211: I think we can leverage the (internal) implementation detail that `QuickAssistAssistantImpl extends ContentAssistant` and ContentAssistant allows some listeners that might be useful to deduce whether the dialog is active or not.

mickaelistria: Another possibility is to also check the current Shell and find a Table child and verify whether it does contain the usual "Computing..." item so we know we can replace it.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/codeactions/LSPCodeActionMarkerResolution.java
```diff
@@ -147,31 +161,74 @@ private void checkMarkerResoultion(IMarker marker) throws IOException, CoreExcep
...
+			}
+			// Hover case
+			if (textViewer instanceof ITextViewerExtension2) {
+				ITextHover hover = ((ITextViewerExtension2) textViewer).getCurrentTextHover();
+				boolean hoverShowing = hover != null;
+				if (hoverShowing) {
+					Field f = TextViewer.class.getDeclaredField("fTextHoverManager"); //$NON-NLS-1$
+					f.setAccessible(true);
+					AbstractInformationControlManager manager = (AbstractInformationControlManager) f.get(textViewer);
+					manager.showInformation();
```
mickaelistria, line 224: Would re-calling the Hover action against the editor result in a refresh? Something like `editor.getAction(ITextEditorActionConstants.SHOW_INFORMATION)`

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/codeactions/LSPCodeActionMarkerResolution.java
BoykoAlex: Tried adding a `ICompletionListener` to quick assistant. Didn't work. The `LSPCodeActionMarkerResolution` is created when the completion proposal session is started already... This results in indeterminate state for quick assistant proposals being active or not because it is unclear what was the starting state. If no session start/end event received it is unclear if quick assist proposals are active or not.
The Shell-Table Computing entry in the table solution would work i think... However, I feel it'd be less hacky to add `boolean isProposalPopupActive()` to `IQuickAssistant` interface. If you think it is a good option, I'd create a PR for Eclipse Platform Text.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/codeactions/LSPCodeActionMarkerResolution.java
BoykoAlex: I've tried `editor.getAction(ITextEditorActionConstants.SHOW_INFORMATION)` previously and it does bring the popup  correctly but there are 2 caches:
- Needs the editor cursor to be located where the popup to appear
- The popup does not go away until a click outside the shell is received. On the contrary, the regular hover goes away when the mouse is moved off the hover shell.

BoykoAlex:
@mickaelistria what do you think about the comments i left?

BoykoAlex:
Abandoned in favour of #290

PR #134, opened 2022-05-26
JUnit test is added for for IllegalStateException: Duplicate RPC meth…

vrubezhny:
…od runInTerminal issue

### DIFF
--- a/org.eclipse.lsp4e.test/src/org/eclipse/lsp4e/test/debug/DebugTest.java
+++ b/org.eclipse.lsp4e.test/src/org/eclipse/lsp4e/test/debug/DebugTest.java
@@ -1,0 +1,60 @@
+/*******************************************************************************
+ * Copyright (c) 2022 Red Hat Inc. and others.
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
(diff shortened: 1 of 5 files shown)


org.eclipse.lsp4e.test/src/org/eclipse/lsp4e/test/debug/DebugTest.java
```diff
@@ -0,0 +1,49 @@
+package org.eclipse.lsp4e.test.debug;
```
mickaelistria, line 1: Please add copyright header.

mickaelistria (changes requested):
I think this test is too internal and doesn't really test LSP4E debug is functioning, but is instead and internal test which still has some value. I wold simply rename `testInitializedDSPDebugTarget` to something that clarifies it's more internal stuff, eg `testSupportedJSONRPCMethods`

vrubezhny:
> I think this test is too internal and doesn't really test LSP4E debug is functioning, but is instead and internal test which still has some value. I wold simply rename `testInitializedDSPDebugTarget` to something that clarifies it's more internal stuff, eg `testSupportedJSONRPCMethods`

Yes, the intention was to test `IDebugProtocolClient` implementation (`DSPDebugTarget`), not the launch itself. 
Agreed on renaming the  test. Copyright header is added.

PR #137, opened 2022-05-26
Fix buffer leak in diagnostics handler

ahmedneilhussain:
When a language server mounts a project, it can send back diagnostics for the whole project, not merely those files that have been opened in the IDE so far. In order to be able to create the annotations, LSPDiagnosticsToMarkers will mount the file buffers temporarily. At present, it is not releasing them afterwards. This has caused us some issues with large projects with a great many files and markers.

See the note on reference counting at https://help.eclipse.org/2020-09/index.jsp?topic=%2Forg.eclipse.platform.doc.isv%2Freference%2Fapi%2Forg%2Feclipse%2Fcore%2Ffilebuffers%2FIFileBufferManager.html


mickaelistria:
Thank you for this good change!

PR #138, opened 2022-05-27
concurrency: Ensure document changes are sent sequentially to the server

ahmedneilhussain:
As discussed on the lsp4e-dev mailing list, these are some fixes we made internally to try and address some weird behaviour we used to get, particularly under load. The existing implementation submits document update events to a thread pool where they may get scheduled in any order. This can cause the client's and server's view of the state of a document to diverge, resulting in phantom error markers and so on.

This work was originally submitted in https://github.com/eclipse/lsp4e/pull/121 but with more extensive changes that have proved contentious. As agreed, I have cut this down to a minimal fix for serialising document updates and providing a mechanism for other code that wants/needs to be inserted into that stream of events.

### DIFF
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LSPEclipseUtils.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LSPEclipseUtils.java
@@ -407,28 +407,29 @@ public class LSPEclipseUtils {
 	 *            document to modify
 	 * @param edits
 	 *            list of LSP TextEdits
+	 * @throws BadLocationException
 	 */
-	public static void applyEdits(IDocument document, List<? extends TextEdit> edits) {
+	public static void applyEdits(IDocument document, List<? extends TextEdit> edits) throws BadLocationException {
 		if (document == null || edits == null || edits.isEmpty()) {
 			return;
-		}
-
-		IDocumentUndoManager manager = DocumentUndoManagerRegistry.getDocumentUndoManager(document);
-		if (manager != null) {
-			manager.beginCompoundChange();
 		}
 
 		MultiTextEdit edit = new MultiTextEdit();
 		for (TextEdit textEdit : edits) {
 			if (textEdit != null) {
-				try {
-					int offset = LSPEclipseUtils.toOffset(textEdit.getRange().getStart(), document);
-					int length = LSPEclipseUtils.toOffset(textEdit.getRange().getEnd(), document) - offset;
-					edit.addChild(new ReplaceEdit(offset, length, textEdit.getNewText()));
-				} catch (BadLocationException e) {
-					LanguageServerPlugin.logError(e);
+				int offset = LSPEclipseUtils.toOffset(textEdit.getRange().getStart(), document);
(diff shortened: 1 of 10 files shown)


org.eclipse.lsp4e/src/org/eclipse/lsp4e/LSPEclipseUtils.java
```diff
@@ -407,29 +407,30 @@ public static void applyEdit(TextEdit textEdit, IDocument document) throws BadLo
...
+					throw new BadLocationException("Invalid location information found applying edits"); //$NON-NLS-1$
 				}
+				edit.addChild(new ReplaceEdit(offset, length, textEdit.getNewText()));
 			}
 		}
+
+		IDocumentUndoManager manager = DocumentUndoManagerRegistry.getDocumentUndoManager(document);
+		if (manager != null) {
+			manager.beginCompoundChange();
+		}
```
mickaelistria, line 433: I believe this particular patch of the change may already fix some bugs. Would it make sense to have it in a dedicate commit?

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServerWrapper.java
```diff
@@ -866,6 +868,27 @@ int getVersion(URI uri) {
...
+	/**
+ 	 * Submit an asynchronous call (i.e. to the language server) that be inserted into the current position w.r.t.
+ 	 * document change events
+ 	 *
+ 	 * @param <U> Computation return type
+ 	 * @param file Document to serialise on
+ 	 * @param fn Asynchronous computation on the language server
+ 	 * @return Asynchronous result object.
+ 	 */
+ 	public <U> @Nullable CompletableFuture< U> executeOnCurrentVersionAsync(IFile file,
```
mickaelistria, line 880: I like the idea of adding such a method, however the language server don't only work on IFile but instead work on URIs or even IDocument. IMO the signature should replace IFile by one of those types.

mickaelistria: Also, could this be made package visible only? The less APIs we expose, the better.

ahmedneilhussain: Done, and changed to URI instead of File

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LSPEclipseUtils.java
ahmedneilhussain: Split out as requested

ahmedneilhussain:
Rebased onto master. It ran into some changes to the unit tests that someone seems to have made ;-)

mickaelistria:
I've reorganized a bit the commits, replaced usage of locks/synchronized by AtomicReference and removed some methods that are currently not used; pushed as https://github.com/eclipse/lsp4e/pull/144/commits/50d1453df388c7823141bde9e65a094f84c5dcec . Thanks!

PR #139, opened 2022-05-30
Maintenance/mitigate flaky tests

ahmedneilhussain:
This would probably be better entitled 'mitigate noisy tests' - I found there was an awful lot of noise from the teardown code having race conditions with some of the lazily server startup/document connection. I think this was mostly benign but made it very hard to try and see what the problem might be when investigating test failures. Although it turns out benign exceptions that just get logged are not so benign if `NoErrorLoggedRule` is being used!  The logs are a lot quieter now, hopefully, and I've found and tried to fix a few definite cases of flakiness.

The changes can be summarised:

- Add some hooks to allow the test teardown logic to wait for some of the startup/connect logic to complete. It does make the tests run a bit slower but as a full run is only about 100s on my machine that's pretty respectable for UI tests that require the whole IDE to start up. An alternative might have been to allow the setup logic to detect when the server was already being shut down and not bother continuing, but I think it's safer and better to make test code accommodate production code than vice versa.
- Added some hooks so that files created on-the-fly are tracked and removed by some common code in `TestUtils` and `AllClearRule` so that where possible files are not deleted while document connection might still be pending in a Job queue or worker pool. Changed a lot of existing tests to use this common framework rather than doing it with duplicated logic (minus the synchronisation)
- Changed the test stream connection provider to only close the outbound streams in both directions, and relying on the `EOF` to close the corresponding paired stream. This got rid of a lot of the `AsynchronousCloseException`s
- Changed the tests that look for annotations of a specific colour to use a Euclidian distance and a fudge factor rather than an exact comparison, as an exact match doesn't play well with platform anti-aliasing.

I think b8340c3bcb24bf5dcda3aaee450cf07a74a76531 is an outright bug fix rather than a testing improvement, because the stop() and disconnect() could both call one another resulting in multiple 'shutdown' jobs being scheduled. I nearly introduced a bug because I forgot that language servers can be restarted once stopped but I hope I've got this right. Would be very grateful if someone else can take a look.


mickaelistria:
Thanks. I'll try to cherry-pick and apply the most trivial patches first and will then come back to discuss the less simple ones.

org.eclipse.lsp4e.test/src/org/eclipse/lsp4e/test/linkedediting/LinkedEditingTest.java
```diff
@@ -141,13 +141,12 @@ protected boolean condition() {
 		}.waitForCondition(Display.getCurrent(), 3000);
 
 		IAnnotationModel model = sourceViewer.getAnnotationModel();
-		Annotation annotation = findAnnotation(sourceViewer, 14);
-		Assert.assertNotNull(annotation);
-//		System.out.println("testLinkedEditingExitPolicy: annotation found at offset 14: " + annotation.getType());
-		assertTrue(annotation.getType().startsWith("org.eclipse.ui.internal.workbench.texteditor.link"));
+		List<Annotation> annotations = findAnnotations(sourceViewer, 14);
+		Assert.assertFalse(annotations.isEmpty());
```
mickaelistria, line 145: We don't need to check that.

org.eclipse.lsp4e.test/src/org/eclipse/lsp4e/test/linkedediting/LinkedEditingTest.java
```diff
@@ -176,26 +174,26 @@ protected boolean condition() {
 		}.waitForCondition(Display.getCurrent(), 3000);
 		
 		model = sourceViewer.getAnnotationModel();
-		annotation = findAnnotation(sourceViewer, 15);
+		annotations = findAnnotations(sourceViewer, 15);
 		// No "linked" annotation is to be found at this offset
-		if (annotation != null) {
-//			System.out.println("testLinkedEditingExitPolicy: annotation found at offset 15: " + annotation.getType());
-			assertFalse(annotation.getType().startsWith("org.eclipse.ui.internal.workbench.texteditor.link"));
+		if (!annotations.isEmpty()) {
```
mickaelistria, line 179: No need to check that, just next assert is sufficient.

org.eclipse.lsp4e.test/src/org/eclipse/lsp4e/test/linkedediting/LinkedEditingTest.java
```diff
@@ -141,13 +141,12 @@ protected boolean condition() {
 		}.waitForCondition(Display.getCurrent(), 3000);
 
 		IAnnotationModel model = sourceViewer.getAnnotationModel();
-		Annotation annotation = findAnnotation(sourceViewer, 14);
-		Assert.assertNotNull(annotation);
-//		System.out.println("testLinkedEditingExitPolicy: annotation found at offset 14: " + annotation.getType());
-		assertTrue(annotation.getType().startsWith("org.eclipse.ui.internal.workbench.texteditor.link"));
+		List<Annotation> annotations = findAnnotations(sourceViewer, 14);
+		Assert.assertFalse(annotations.isEmpty());
+		assertTrue(annotations.stream().anyMatch(a -> a.getType().startsWith("org.eclipse.ui.internal.workbench.texteditor.link")));
```
mickaelistria, line 146: better use `.filter(...).findAny()` here and `assertTrue(annotation.isPresent())`

org.eclipse.lsp4e.test/src/org/eclipse/lsp4e/test/linkedediting/LinkedEditingTest.java
ahmedneilhussain: Hi @mickaelistria you've already cherry-picked the commit and fixed this, is that right?

mickaelistria:
Can you please tro to rebase it?

ahmedneilhussain:
> Can you please tro to rebase it?

HI @mickaelistria I've got it consistent with master now (I think) and all of the changes you cherry-picked excluded.

sebthom:
@ahmedneilhussain can you please bump the bundle version of `org.eclipse.lsp4e.tests.mock`. this solves the build error:
```
[ERROR] Failed to execute goal org.eclipse.tycho.extras:tycho-p2-extras-plugin:2.7.3:compare-version-with-baselines (compare-version-with-baseline) on project org.eclipse.lsp4e.tests.mock: 
Only qualifier changed for (org.eclipse.lsp4e.tests.mock/0.14.4.202207111108). Expected to have bigger x.y.z than what is available in baseline (0.14.4.202205311633) -> [Help 1]
```

org.eclipse.lsp4e.test/src/org/eclipse/lsp4e/test/rename/RenameTest.java
```diff
@@ -103,10 +103,12 @@ public void testAsyncRenameHandlerEnablement() throws Exception {
...
-
-		Thread.sleep(2 * delay);
-		assertTrue(command.isEnabled());
-		assertTrue(command.isHandled());
+		
+		Thread.sleep(delay * 3);
+		
+		// Put back so shutdown doesn't time out
+		MockLanguageServer.INSTANCE.setTimeToProceedQueries(0);
+		assertTrue(command.isEnabled() && command.isHandled());
```
sebthom, line 111: could you please keep the asserts separate:
```
assertTrue(command.isEnabled());
assertTrue(command.isHandled());
```

org.eclipse.lsp4e.test/src/org/eclipse/lsp4e/test/TestUtils.java
```diff
@@ -59,6 +59,8 @@ public class TestUtils {
 	private TestUtils() {
 		// this class shouldn't be instantiated
 	}
+	
+	private static List<File> tempFiles = new ArrayList<>();
```
sebthom, line 64: This adds a second static field with the same name `tempFiles` to the class.

image

org.eclipse.lsp4e.test/src/org/eclipse/lsp4e/test/AllCleanRule.java
```diff
@@ -53,14 +59,17 @@ protected void finished(Description description) {
 	}
 
 	private void clear() {
-		PlatformUI.getWorkbench().getActiveWorkbenchWindow().getActivePage().closeAllEditors(false);
+		// Give the platform three attempts to shut down windows
+		for (int i = 3; i > 0 && !PlatformUI.getWorkbench().getActiveWorkbenchWindow().getActivePage().closeAllEditors(false); i--) {}
```
sebthom, line 63: What's the point of doing this in a loop without a sleep or something in between?

mickaelistria:
Thanks!

org.eclipse.lsp4e.test/src/org/eclipse/lsp4e/test/TestUtils.java
ahmedneilhussain: Sorry, was a merge clash caused by this part of the original PR being cherry picked out and integrated independently a few weeks ago. Fixed anyway!

org.eclipse.lsp4e.test/src/org/eclipse/lsp4e/test/rename/RenameTest.java
ahmedneilhussain: Sorry - the email notification of your comment only arrived after I had independently fixed the build errors and @mickaelistria had merged... Please feel free to revert to two separate asserts.

org.eclipse.lsp4e.test/src/org/eclipse/lsp4e/test/AllCleanRule.java
ahmedneilhussain: It's a fairly long-running call so I didn't feel the need to stick a sleep in between, but I take your point!

PR #145, opened 2022-06-01
Use the custom URI map provided by LSPEclipseUtils.toUri on connect

rubenporras:
Use the custom URI map provided by LSPEclipseUtils.toUri on connect in
Ida41f97f25929523381ee7b2077cdf358bd1f9f7. This is indeed a leftover of
that commit that went missing when crating the PR out of our fork.


mickaelistria:
Thanks!

PR #148, opened 2022-06-09
handlers: Scope lsp4e handlers to language server-enabled editors

ahmedneilhussain:
I think the command handlers for lsp4e are scoped too widely - whenever text is generally selected in an editor. Restricting this to editors that actually have the language server enabled for them, and also using `activeWhen` rather than `enabledWhen` more consistently: I think the former is intended for controlling whether plugins have any interest at all in the current view/editor, and the latter for deciding whether a command makes sense or not for the detailed selection.

### DIFF
--- a/org.eclipse.lsp4e/plugin.xml
+++ b/org.eclipse.lsp4e/plugin.xml
@@ -91,7 +91,7 @@
       <handler
             class="org.eclipse.lsp4e.operations.references.LSFindReferences"
             commandId="org.eclipse.ui.genericeditor.findReferences">
-         <enabledWhen>
+         <activeWhen>
             <and>
                <with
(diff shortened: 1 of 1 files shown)


ahmedneilhussain:
See Bug 564839 https://github.com/eclipse/lsp4e/commit/d9084418d533430af4d366c43734ebde165f7a75 for a similar change that was made before

mickaelistria:
Please bump version of the bundle by +0.0.1 by the way as this the first change since last release.

mickaelistria:
Rebased and merged as https://github.com/eclipse/lsp4e/commit/cad85f9f9c615a2adb3854e80cfcc4294d70a745

PR #150, opened 2022-06-14
Name executor service that launches language servers

rubenporras:


org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServerWrapper.java
```diff
@@ -241,7 +242,7 @@ public synchronized void start() throws IOException {
 			}).thenApply((server) -> {
 				try {
 					LanguageClientImpl client = serverDefinition.createLanguageClient();
-					ExecutorService executorService = Executors.newCachedThreadPool();
+					ExecutorService executorService = Executors.newCachedThreadPool(new ThreadFactoryBuilder().setNameFormat("Language-server-launcher-%d").build()); //$NON-NLS-1$
```
mickaelistria, line 245: That's a good idea. Should we show the language server id in the name as well to easily debug cases where multiple LS as running?

rubenporras: Yes, let us do that. I have updated the PR.

mickaelistria:
Thank you!

PR #152, opened 2022-06-16
formatting: ensure formatting requests are sequenced wrt updates

ahmedneilhussain:
This was originally part of the concurrency fixes we submitted: I lost track of what eventually made it into the reduced PR and forgot to resubmit this change independently.

We needed to make a fix along these lines as without it it's possible for formatting requests to become interleaved with in-flight document change notifications and the document to become garbled.

### DIFF
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/DocumentContentSynchronizer.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/DocumentContentSynchronizer.java
@@ -119,7 +119,7 @@ final class DocumentContentSynchronizer implements IDocumentListener {
  	 * @param fn Asynchronous computation on the language server
  	 * @return Asynchronous result object.
  	 */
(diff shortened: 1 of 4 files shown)


mickaelistria:
Do you think you could create a test that reproduces the current issue (and validates the fix)

ahmedneilhussain:
> Do you think you could create a test that reproduces the current issue (and validates the fix)

I've added a test, but making it fail reliably without the fix would be hard without adding some hooks and delays into the production client - adding delays into the server won't help because DocumentDidChange is a notification, not a method that will return a future you can block on...

ahmedneilhussain:
@mickaelistria 
Had a chat with my boss about this and we think it's not quite enough as it stands. The following sequence is still possible:

- The user requests formatting
- The document is updated by the user
- Formatting completes and edits are applied on the wrong document

The solution (which we sort of had in our original batch of concurrency fixes before separating them all out) is to check that the document version when receiving the edits is unchanged from the point when formatting was requested, and bail out if they differ.

I'll rework along these lines and resubmit.

PR #154, opened 2022-06-22
Improve marker resolution performance

rubenporras:
by not connecting the file where we retrieve marker information. That is
an unnecessary step which can be expensive (for example the Xtext
language server rebuilds the source) and causes timeouts on
checkMarkerResoultion

### DIFF
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServiceAccessor.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServiceAccessor.java
@@ -145,15 +145,7 @@ public class LanguageServiceAccessor {
 	public static @NonNull List<CompletableFuture<LanguageServer>> getInitializedLanguageServers(@NonNull IFile file,
 			@Nullable Predicate<ServerCapabilities> request) throws IOException {
 		synchronized (startedServers) {
-			Collection<LanguageServerWrapper> wrappers = getLSWrappers(file, request);
-			return wrappers.stream().map(wrapper -> wrapper.getInitializedServer().thenApplyAsync(server -> {
-				try {
-					wrapper.connect(file, null);
-				} catch (IOException e) {
-					LanguageServerPlugin.logError(e);
-				}
-				return server;
-			})).collect(Collectors.toList());
+			return getLSWrappers(file, request).stream().map(LanguageServerWrapper::getInitializedServer).collect(Collectors.toList());
 		}
 	}
 
@@ -231,7 +223,6 @@ public class LanguageServiceAccessor {
 			throws IOException {
 		LanguageServerWrapper wrapper = getLSWrapper(file.getProject(), lsDefinition, file.getFullPath());
 		if (capabilitiesComply(wrapper, capabilitiesPredicate)) {
-			wrapper.connect(file, null);
 			return wrapper.getInitializedServer();
 		}
 		return null;
@@ -256,7 +247,6 @@ public class LanguageServiceAccessor {
 		IPath initialPath = LSPEclipseUtils.toPath(document);
 		LanguageServerWrapper wrapper = getLSWrapperForConnection(document, lsDefinition, initialPath);
 		if (capabilitiesComply(wrapper, capabilitiesPredicate)) {
-			wrapper.connect(document);
 			return wrapper.getInitializedServer();
 		}
 		return null;


org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServiceAccessor.java
```diff
@@ -144,13 +144,20 @@ public boolean isActive() {
 
 	public static @NonNull List<CompletableFuture<LanguageServer>> getInitializedLanguageServers(@NonNull IFile file,
 			@Nullable Predicate<ServerCapabilities> request) throws IOException {
+			return getInitializedLanguageServers(file, request, true);
+	}
```
rubenporras, line 148: this method is now redundant, so I could deprecate it

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServiceAccessor.java
```diff
@@ -229,9 +236,28 @@ public static CompletableFuture<LanguageServer> getInitializedLanguageServer(@No
 			@NonNull LanguageServerDefinition lsDefinition,
 			Predicate<ServerCapabilities> capabilitiesPredicate)
 			throws IOException {
+		return getInitializedLanguageServer(file, lsDefinition, capabilitiesPredicate, true);
+	}
```
rubenporras, line 240: this method is now redundant, so I could deprecate it

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServiceAccessor.java
mickaelistria: Yes please `@Deprecate(forRemoval=true)`

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServiceAccessor.java
```diff
@@ -144,13 +144,20 @@ public boolean isActive() {
 
 	public static @NonNull List<CompletableFuture<LanguageServer>> getInitializedLanguageServers(@NonNull IFile file,
 			@Nullable Predicate<ServerCapabilities> request) throws IOException {
+			return getInitializedLanguageServers(file, request, true);
+	}
+
+	public static @NonNull List<CompletableFuture<LanguageServer>> getInitializedLanguageServers(@NonNull IFile file,
+			@Nullable Predicate<ServerCapabilities> request, boolean connectFile) throws IOException {
```
mickaelistria, line 151: I suggest `forceConnectFile` instead, as the file may already be connected. Also, do you have any suggestion what could be a better term than `connect` ?

mickaelistria (commented):
i like this idea a lot. We indeed often don't need to `connect` a file to get some features. I'm actually even curious when connecting a file is really necessary; apart from open/edit/close? Maybe we should never connect by default and only do it when necessary?

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServiceAccessor.java
rubenporras: Thanks for the review. I have done the changes.

You mean, a better name for `LanguageServerWrapper#connect`? I guess `open` would be a better name, in the end the effect of it is to created the document synchronizer which sends a didOpen notification, isn't it?. I do not know if it is worth renaming the methods now that they already exist.

rubenporras:
I would also think one should only `connect` if it is needed, but I did not checked the codebase. This part of the codebase is a problem for us because in the markers view one can have many markers, and displaying them would send the didOpen for a lot of documents.

rubenporras:
From the last version of the PR it is clear that there is nothing in LSP4E that needs the side-effect of calling `LanguageServerWrapper#connect` when asking for initialized servers, so the easiest change would just be to remove the side-effect and leave the methods as they are. 
It could be a breaking change if some other project relies on the side-effect. Is there a policy in LSP4E for that kind of changes? Is it something which can be done if it is documented, for example in the changelog?

mickaelistria:
> so the easiest change would just be to remove the side-effect and leave the methods as they are.

Great, let's try it!

> Could be a breaking change if some other project relies on the side-effect. Is there a policy in LSP4E for that kind of changes?

The policy has been to stay on 0.x stream to somehow emphasize their is no API policy ;)

PR #155, opened 2022-06-23
Remove NPE in LanguageServiceAccessor.getLSWrapper(IProject,

rubenporras:
LanguageServerDefinition, IPath) when the project is null and there are
started language server wrappers.


org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServiceAccessor.java
```diff
@@ -428,12 +428,12 @@ public static LanguageServerWrapper getLSWrapperForConnection(@NonNull IProject
 		return getLSWrapper(project, serverDefinition);
 	}
 
-	private static LanguageServerWrapper getLSWrapper(@NonNull IProject project,
```
rubenporras, line 431: this declaration was incorrect, as there is a check for null project on line 443

Issue #157, opened 2022-06-26
Allow to launch project specific Language Server (i.e. different binary version) for a given ContentType


sebthom:
I have the use case where I want to configure the SDK version of a programming language on project level. Each version of the programming language SDK bundles it's own language server. 

Currently the executable path to the language server is configured in the constructor of a language server launcher (aka StreamConnectionProvider). See for example [WWD/YAMLLanguageServer.java#L79](https://github.com/eclipse/wildwebdeveloper/blob/master/org.eclipse.wildwebdeveloper/src/org/eclipse/wildwebdeveloper/yaml/YAMLLanguageServer.java#L79) or  [WWD/AngularLanguageServer.java#L45](https://github.com/eclipse/wildwebdeveloper/blob/master/org.eclipse.wildwebdeveloper/src/org/eclipse/wildwebdeveloper/angular/AngularLanguageServer.java#L45).
Once the StreamConnectionProvider object is constructed the  `LanguageServerWrapper` immediately calls the [StreamConnectionProvider.start()](https://github.com/eclipse/lsp4e/blob/2ece5f843933b35cd89bb931486b92b52dbdd61e/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServerWrapper.java#L235) method expecting the language server to be launched.

However, at this point no information about the related project is provided and thus no information about the corresponding SDK's language server version to be used can be obtained. This information is only available later when `LanguageServerWrapper` calls [StreamConnectionProvider.getInitializationOptions(rootURI)](https://github.com/eclipse/lsp4e/blob/2ece5f843933b35cd89bb931486b92b52dbdd61e/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServerWrapper.java#L368).

Currently I see these options to provide the `rootURI` information before the  `StreamConnectionProvider.start()` method is called:
1. add a new lifecycle method that provides the `rootURI` before the `start()` method is called, e.g. `StreamConnectionProvider.setRootURI(URI rootUri)`.
1. Provide the `rootURI` to the start method and overload it like this to be backward compatible:
	  ```java
	  public interface StreamConnectionProvider {
		  /**
		   * @deprecated Use/override {@link #start(URI)}
		   */
		  @Deprecated
		  public default void start() throws IOException {	}

		  public default void start(URI rootUri) throws IOException {
			  start();
		  }
	  ```
[...]

PR #158, opened 2022-06-26
Call lspStreamProvider.getInitializationOptions(rootURI) before .start()

sebthom:
This implements the third option *"Move the call to StreamConnectionProvider.getInitializationOptions(rootURI) method before the call to StreamConnectionProvider.start()"* described in #157

### DIFF
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServerWrapper.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServerWrapper.java
@@ -224,6 +224,7 @@ public class LanguageServerWrapper {
 			}
 		}
 		if (this.initializeFuture == null) {
+			final URI rootURI = getRootURI();
 			this.initializeFuture = CompletableFuture.supplyAsync(() -> {
 				try {
 					if (LoggingStreamConnectionProviderProxy.shouldLog(serverDefinition.id)) {
@@ -232,6 +233,7 @@ public class LanguageServerWrapper {
 					} else {
 						this.lspStreamProvider = serverDefinition.createConnectionProvider();
 					}
+					initParams.setInitializationOptions(this.lspStreamProvider.getInitializationOptions(rootURI));
 					lspStreamProvider.start();
 				} catch (Exception e) {
 					LanguageServerPlugin.logError(e);
@@ -239,7 +241,7 @@ public class LanguageServerWrapper {
 					stop();
 				}
 				return null;
-			}).thenApply((server) -> {
+			}).thenApply(unused -> {
 				try {
 					LanguageClientImpl client = serverDefinition.createLanguageClient();
 					String theardNameFormat = "LS-" + serverDefinition.id + "-launcher-%d"; //$NON-NLS-1$ //$NON-NLS-2$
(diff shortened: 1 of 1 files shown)


org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServerWrapper.java
```diff
@@ -224,6 +224,26 @@ public synchronized void start() throws IOException {
...
+				if (initialPath != null) {
+					File projectDirectory = initialPath.toFile();
+					if (projectDirectory.isFile()) {
+						projectDirectory = projectDirectory.getParentFile();
+					}
+					rootURI = LSPEclipseUtils.toUri(projectDirectory);
+				} else {
+					rootURI = LSPEclipseUtils.toUri(new File("/")); //$NON-NLS-1$
+				}
+			}
```
rubenporras, line 245: I understand that you are just moving the code, but if you would move this block to a method such as
```
private URI getRootUri(IPath initialPath) {
  ...
}
```
you could remove the three lines comment and the code would be cleaner.

mickaelistria: +1 this would be welcome.

sebthom: The comment is only about the first line of the block. keeping or removing it has nothing to do with putting that block in a method. For someone who worked eclipse null analysis the comment is superficial. So for me this is as meaningful as a comment above a variable assignment that describes that in the next line a value is assigned to a the variable on the left side. So I am all in for removing the comment but I don't really see value in putting the block into a separate method.

rubenporras: In the end it is only about taste, to me the method call looks like a more natural way to remove the null analysis complaint. Probably nobody would wonder why there is a method call, while now apparently someone wondered (or thought others would wonder and added the comment) why there is an assign such as `final IPath initialPath = this.initialPath;`.

In any case, it is not very important, so I would suggest we forget it ;)

Still, while looking at the code more carefully I have realize the same type of assignment is done on line 250 `IProject project = this.initialProject;`, what about moving the comment before that line so that it appears before its first occurrence and not the second?

Also, line 249 `URI rootURI = null` can be moved inside the if block, and then, the block should use project, not this.initialProjoct. All together we would have

```
					// This is required due to overzealous static analysis. Dereferencing
					// this.initialProject / this.initialPath directly will trigger a "potential null"
					// warning/error. Checking for this.<member> == null is not enough.					
					IProject project = this.initialProject;
					if (project != null && project.exists()) {
						URI rootURI = LSPEclipseUtils.toUri(project);
						initParams.setRootUri(rootURI.toString());
						initParams.setRootPath(rootURI.getPath());
					} else {

```

Would you mind doing that as part of the PR?

BTW: The change makes sense to me.

sebthom: How about like this? https://github.com/eclipse/lsp4e/pull/158/commits/c44abe29c93bdb2cab02343ef5728b87d6d4a6bd

mickaelistria:
Thanks folks!

PR #163, opened 2022-06-28
Add missing synchronization guards when accessing startedServers Set

sebthom:


mickaelistria:
Which problem is it fixing? Do you have steps to reproduce issues that are fixed with this patch?

sebthom:
I saw that the synchronization blocks are missing while investigating the deadlocks.

The current code as is technically broken. Either all access paths need to be synchronized or none.

mickaelistria:
> Either all access paths need to be synchronized or none.

Actually, if the structure is sufficiently thread-safe on access, read-only operations may not need to be synchronized.

sebthom:
The methods that are currently unguarded iterate over a set that might by modified by other methods at the same time. This basically begs for potential ConcurrentModificationExceptions.

I would agree with you that is change might not be required if LanguageServiceAccessor was an internal class where you have control over which of its methods are called when by other code. But this is a public class where all methods can be concurrently invoked by external code as pleased.

Issue #164, opened 2022-06-29
conflicting keys


jukzi:
When i start eclipse i see conflicting Keys logged.
for example:
```
A conflict occurred for ALT+SHIFT+R:
Binding(ALT+SHIFT+R,
	ParameterizedCommand(Command(org.eclipse.jdt.ui.edit.text.java.rename.element,Rename - Refactoring ,
		Rename the selected element,
		Category(org.eclipse.jdt.ui.category.refactoring,Refactor - Java,Java Refactoring Actions,true),
		WorkbenchHandlerServiceHandler("org.eclipse.jdt.ui.edit.text.java.rename.element"),
		,,true),null),
	org.eclipse.ui.defaultAcceleratorConfiguration,
	org.eclipse.ui.contexts.window,,,system)
Binding(ALT+SHIFT+R,
	ParameterizedCommand(Command(org.eclipse.ui.edit.rename,Rename,
		Rename the selected item,
		Category(org.eclipse.ui.category.file,File,null,true),
		WorkbenchHandlerServiceHandler("org.eclipse.ui.edit.rename"),
		,,true),null),
	org.eclipse.ui.defaultAcceleratorConfiguration,
	org.eclipse.ui.contexts.window,,,system)

```
while "org.eclipse.ui.edit.rename" should have key "F2" (see https://github.com/eclipse-platform/eclipse.platform.ui/blob/master/bundles/org.eclipse.ui/plugin.xml#L135)

this is overridden by lsp4e here:
https://github.com/eclipse/lsp4e/blob/master/org.eclipse.lsp4e/plugin.xml#L337

lsp4e should not override commands of other plugins.

Also there is a conflict for CTRL+SHIFT+T
org.eclipse.jdt.ui.navigate.open.type
vs
org.eclipse.lsp4e.symbolinworkspace
see https://github.com/eclipse/lsp4e/blob/master/org.eclipse.lsp4e/plugin.xml#L326
could lsp4e please use keys that do not conflict with jdt?

```
A conflict occurred for CTRL+SHIFT+T:
Binding(CTRL+SHIFT+T,
	ParameterizedCommand(Command(org.eclipse.jdt.ui.navigate.open.type,Open Type,
		Open a type in a Java editor,
		Category(org.eclipse.ui.category.navigate,Navigate,null,true),
		org.eclipse.ui.internal.WorkbenchHandlerServiceHandler@514abad9,
		,,true),null),
[...]

PR #165, opened 2022-06-29
do not override org.eclipse.ui.edit.rename #164

jukzi:
for https://github.com/eclipse/lsp4e/issues/164

### DIFF
--- a/org.eclipse.lsp4e/plugin.xml
+++ b/org.eclipse.lsp4e/plugin.xml
@@ -331,25 +331,6 @@
             schemeId="org.eclipse.ui.defaultAcceleratorConfiguration"
             sequence="M1+M2+F">
       </key>
-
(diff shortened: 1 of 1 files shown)


mickaelistria:
Users actually expect this shortcut to work; and handler enablement rules make it actually not conflicting in practice. So it's most likely better to keep it.

jukzi:
windows users are used to "F2". see https://www.groovypost.com/howto/rename-grouped-files-windows-hotkey/
It's not good that a plugin overrides eclipse preferences.. If you would like to change it for eclipse you should ask for a change in org.eclipse.ui instead of overriding it in another plugin.

jukzi:
see also https://bugs.eclipse.org/bugs/show_bug.cgi?id=538244 for what users expect: F2

PR #166, opened 2022-06-29
LanguageServerWrapper: Don't set rootURI if initialProject and initialPath are null or invalid

sebthom:
...in case LanguageServerWrapper initialProject and initialPath are null or
point non-existing location


mickaelistria:
I'm not sure it's a good choice. In many cases, the projects that are visible in the workspace are not children of the workspace location on filesystem. I'm thinking maybe it would be possible to pass no root in such case?

sebthom:
> I'm thinking maybe it would be possible to pass no root in such case?

@mickaelistria This makes even more sense to me. I pushed an update.

rubenporras (commented):
The code makes sense to me. Did you checked if the two test failures are related to the commit?

sebthom:
> Did you checked if the two test failures are related to the commit?

Good catch! Fixed it.

rubenporras:
@sebthom , looks good to me know, but I would leave @mickaelistria the final review.

mickaelistria:
I didn't try it yet but the idea and code seems good. Let's merge and release to snapshots to give time to get feedback.

Issue #161, opened 2022-06-28
Deadlock in LSPFoldingReconciler


sebthom:
Just encountered a deadlock again while pasting an XML comment into a Target Definition file with Eclipse 2022-06 and lsp4e 0.13.12.202206011407. I think this is similar to #22 

```java
Found one Java-level deadlock:
=============================
"main":
  waiting to lock monitor 0x00000194e9615ff0 (object 0x000000070ab59c48, a java.util.HashSet),
  which is held by "org.eclipse.lsp4e.operations.folding.LSPFoldingReconciler"

"org.eclipse.lsp4e.operations.folding.LSPFoldingReconciler":
  waiting to lock monitor 0x00000194ef124bb0 (object 0x000000070222b270, a java.lang.Object),
  which is held by "main"

Java stack information for the threads listed above:
===================================================
"main":
        at org.eclipse.lsp4e.LanguageServiceAccessor.getLSWrappers(LanguageServiceAccessor.java:350)
        - waiting to lock <0x000000070ab59c48> (a java.util.HashSet)
        at org.eclipse.lsp4e.LanguageServiceAccessor.getLanguageServers(LanguageServiceAccessor.java:601)
        at org.eclipse.lsp4e.operations.linkedediting.LSPLinkedEditingBase.collectLinkedEditingRanges(LSPLinkedEditingBase.java:71)
        at org.eclipse.lsp4e.operations.linkedediting.LSPLinkedEditingReconcilingStrategy.updateLinkedEditing(LSPLinkedEditingReconcilingStrategy.java:179)
        at org.eclipse.lsp4e.operations.linkedediting.LSPLinkedEditingReconcilingStrategy.documentChanged(LSPLinkedEditingReconcilingStrategy.java:168)
        at org.eclipse.jface.text.AbstractDocument.doFireDocumentChanged2(AbstractDocument.java:748)
        at org.eclipse.jface.text.AbstractDocument.doFireDocumentChanged(AbstractDocument.java:717)
        at org.eclipse.jface.text.AbstractDocument.doFireDocumentChanged(AbstractDocument.java:701)
        at org.eclipse.jface.text.AbstractDocument.fireDocumentChanged(AbstractDocument.java:775)
        at org.eclipse.jface.text.AbstractDocument.replace(AbstractDocument.java:1100)
        at org.eclipse.core.internal.filebuffers.SynchronizableDocument.replace(SynchronizableDocument.java:176)
[...]

Issue #22, opened 2021-11-30
Deadlock with codeminings


mickaelistria:
IDE froze while editing a .target file (extends XML content-type), `jstack` says:
```
Found one Java-level deadlock:
=============================
"main":
  waiting to lock monitor 0x00007fc9ca4f1c00 (object 0x0000000085a63440, a java.util.HashSet),
  which is held by "org.eclipse.jface.text.codemining.CodeMiningReconciler"

"org.eclipse.jface.text.codemining.CodeMiningReconciler":
  waiting to lock monitor 0x00007fc8f4054730 (object 0x0000000097e90c00, a java.lang.Object),
  which is held by "main"

Java stack information for the threads listed above:
===================================================
"main":
	at org.eclipse.lsp4e.LanguageServiceAccessor.getLSWrappers(LanguageServiceAccessor.java:350)
	- waiting to lock <0x0000000085a63440> (a java.util.HashSet)
	at org.eclipse.lsp4e.LanguageServiceAccessor.getLanguageServers(LanguageServiceAccessor.java:588)
	at org.eclipse.lsp4e.operations.linkedediting.LSPLinkedEditingBase.collectLinkedEditingRanges(LSPLinkedEditingBase.java:71)
	at org.eclipse.lsp4e.operations.linkedediting.LSPLinkedEditingReconcilingStrategy.updateLinkedEditingHighlights(LSPLinkedEditingReconcilingStrategy.java:185)
	at org.eclipse.lsp4e.operations.linkedediting.LSPLinkedEditingReconcilingStrategy.documentChanged(LSPLinkedEditingReconcilingStrategy.java:174)
	at org.eclipse.jface.text.AbstractDocument.doFireDocumentChanged2(AbstractDocument.java:748)
	at org.eclipse.jface.text.AbstractDocument.doFireDocumentChanged(AbstractDocument.java:717)
	at org.eclipse.jface.text.AbstractDocument.doFireDocumentChanged(AbstractDocument.java:701)
	at org.eclipse.jface.text.AbstractDocument.fireDocumentChanged(AbstractDocument.java:775)
	at org.eclipse.jface.text.AbstractDocument.replace(AbstractDocument.java:1100)
	at org.eclipse.core.internal.filebuffers.SynchronizableDocument.replace(SynchronizableDocument.java:176)
	- locked <0x0000000097e90c00> (a java.lang.Object)
	at org.eclipse.jface.text.AbstractDocument.replace(AbstractDocument.java:1118)
[...]

PR #167, opened 2022-06-29
LanguageServiceAccessor: Rework thread-safe access to startedServers to prevent dead locks

sebthom:
This is my proposal to address the dead lock issues reported via #161 and #22.

The deadlock occurs when `LSPFoldingReconciler` or `CodeMiningReconciler` are triggered in background threads at the same time when text is modified in the UI thread.

Both executions end up in acquiring a lock on the `SynchronizableDocument` as well as on the `LanguageServiceAccessor.startedServers` field. Unfortunately in case of the Reconcilers first the lock on the `startedServers` field is acquired and then some underlying internal Eclipse platform code tries to get a lock on the document, whereas in case of text modifications in the UI thread first the document gets locked and then it is attempted to acquire the `startedServers` field. Because of race conditions this can lead to deadlocks.

Here is an example:
```python
"main":
        at org.eclipse.lsp4e.LanguageServiceAccessor.getLSWrappers(LanguageServiceAccessor.java:350)
        - waiting to lock <0x000000070ab59c48> (a java.util.HashSet) # <-- SECOND: locking the serversStarted field is attempted
        at org.eclipse.lsp4e.LanguageServiceAccessor.getLanguageServers(LanguageServiceAccessor.java:601)
        at org.eclipse.lsp4e.operations.linkedediting.LSPLinkedEditingBase.collectLinkedEditingRanges(LSPLinkedEditingBase.java:71)
        at org.eclipse.lsp4e.operations.linkedediting.LSPLinkedEditingReconcilingStrategy.updateLinkedEditing(LSPLinkedEditingReconcilingStrategy.java:179)
        at org.eclipse.lsp4e.operations.linkedediting.LSPLinkedEditingReconcilingStrategy.documentChanged(LSPLinkedEditingReconcilingStrategy.java:168)
        at org.eclipse.jface.text.AbstractDocument.doFireDocumentChanged2(AbstractDocument.java:748)
        at org.eclipse.jface.text.AbstractDocument.fireDocumentChanged(AbstractDocument.java:775)
        at org.eclipse.core.internal.filebuffers.SynchronizableDocument.replace(SynchronizableDocument.java:164)
        - locked <0x000000070222b270> (a java.lang.Object) # <-- FIRST: document is locked
        at org.eclipse.jface.text.projection.ProjectionTextStore.replace(ProjectionTextStore.java:106)
        at org.eclipse.jface.text.AbstractDocument.replace(AbstractDocument.java:1093)
        at org.eclipse.jface.text.DefaultDocumentAdapter.replaceTextRange(DefaultDocumentAdapter.java:223)
        at org.eclipse.swt.custom.StyledText.modifyContent(StyledText.java:7639)
        at org.eclipse.swt.custom.StyledText.sendKeyEvent(StyledText.java:8539)
        at org.eclipse.swt.custom.StyledText.paste(StyledText.java:7713)

"org.eclipse.lsp4e.operations.folding.LSPFoldingReconciler":
        at org.eclipse.core.internal.filebuffers.SynchronizableDocument.getChar(SynchronizableDocument.java:140)
        - waiting to lock <0x000000070222b270> (a java.lang.Object)# <-- SECOND: locking the serversStarted field is attempted
        at org.eclipse.core.internal.filebuffers.DocumentReader$DocumentCharSequence.charAt(DocumentReader.java:66)
        at org.eclipse.core.internal.filebuffers.DocumentReader.read(DocumentReader.java:171)
        at org.eclipse.wst.xml.core.internal.contenttype.XMLResourceEncodingDetector.parseInput(XMLResourceEncodingDetector.java:79)
        at org.eclipse.wst.xml.core.internal.contenttype.ContentDescriberForXML.determineValidity(ContentDescriberForXML.java:119)
        at org.eclipse.wst.xml.core.internal.contenttype.ContentDescriberForXML.describe(ContentDescriberForXML.java:77)
        at org.eclipse.core.internal.content.ContentTypeCatalog.collectMatchingByContents(ContentTypeCatalog.java:226)
        at org.eclipse.core.internal.content.ContentTypeMatcher.getDescriptionFor(ContentTypeMatcher.java:87)
        at org.eclipse.core.internal.filebuffers.ResourceTextFileBuffer.getContentType(ResourceTextFileBuffer.java:196)
        at org.eclipse.lsp4e.LSPEclipseUtils.getDocumentContentTypes(LSPEclipseUtils.java:934)
        at org.eclipse.lsp4e.LanguageServersRegistry.matches(LanguageServersRegistry.java:433)
        at org.eclipse.lsp4e.LanguageServiceAccessor.lambda$5(LanguageServiceAccessor.java:355)
        at org.eclipse.lsp4e.LanguageServiceAccessor.getLSWrappers(LanguageServiceAccessor.java:362)
        - locked <0x000000070ab59c48> (a java.util.HashSet) # <-- FIRST: document is locked
        at org.eclipse.lsp4e.LanguageServiceAccessor.getLanguageServers(LanguageServiceAccessor.java:601)
        at org.eclipse.lsp4e.operations.folding.LSPFoldingReconcilingStrategy.reconcile(LSPFoldingReconcilingStrategy.java:126)
```

My attempt to solve it or to make a deadlock situation much less likely is by reducing the need for acquiring a lock on the `startedServers` field only to situations where a matching Language Server is not yet running. Currently for any lookup the field lock is acquired.

To achieve this I first use a thread-safe collection type for `startedServers` to prevent any potential ConcurrentModificationExceptions during read access. And second I only guard code blocks that may modify the `startedServers` field in a double-check-locking style.

For code review I suggest to only check this commit with the ignore white space changes option turned on: [1f3b5565fcdb1d51464139d99e884fa498b34a37?w=1](https://github.com/eclipse/lsp4e/pull/167/commits/1f3b5565fcdb1d51464139d99e884fa498b34a37?w=1)

### DIFF
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServiceAccessor.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServiceAccessor.java
@@ -16,17 +16,16 @@ package org.eclipse.lsp4e;
 
 import java.io.IOException;
 import java.net.URI;
+import java.util.ArrayDeque;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.LinkedHashSet;
-import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
-import java.util.Queue;
 import java.util.Set;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ExecutionException;
@@ -35,7 +34,6 @@ import java.util.stream.Collectors;
 
 import org.eclipse.core.resources.IFile;
 import org.eclipse.core.resources.IProject;
-import org.eclipse.core.runtime.CoreException;
 import org.eclipse.core.runtime.IPath;
 import org.eclipse.core.runtime.Path;
 import org.eclipse.core.runtime.content.IContentDescription;
@@ -63,8 +61,8 @@ public class LanguageServiceAccessor {
 		// this class shouldn't be instantiated
 	}
 
-	private static Set<LanguageServerWrapper> startedServers = new HashSet<>();
-	private static Map<StreamConnectionProvider, LanguageServerDefinition> providersToLSDefinitions = new HashMap<>();
+	private static final Set<LanguageServerWrapper> startedServers = new HashSet<>();
+	private static final Map<StreamConnectionProvider, LanguageServerDefinition> providersToLSDefinitions = new HashMap<>();
 
 	/**
 	 * This is meant for test code to clear state that might have leaked from other
@@ -100,7 +98,7 @@ public class LanguageServiceAccessor {
 		/**
 		 * TODO consider directly returning a {@link TextDocumentIdentifier}
 		 *
-		 * @return
+		 * @return the file URI
 		 */
 		public @NonNull URI getFileUri() {
 			return this.fileUri;
@@ -161,33 +159,33 @@ public class LanguageServiceAccessor {
 				}
 			}
 		}
-
 	}
 
 	public static void enableLanguageServerContentType(
-			@NonNull ContentTypeToLanguageServerDefinition contentTypeToLSDefinition,
-			@NonNull IEditorReference[] editors) {
-		for (IEditorReference editor : editors) {
+			@NonNull final ContentTypeToLanguageServerDefinition contentTypeToLSDefinition,
+			@NonNull final IEditorReference[] editors) {
+		final IContentType contentType = contentTypeToLSDefinition.getKey();
+		if(contentType == null)
+			return;
+		final LanguageServerDefinition lsDefinition = contentTypeToLSDefinition.getValue();
+		if(lsDefinition == null)
+			return;
+
+		for (final IEditorReference editor : editors) {
 			try {
 				if (editor.getEditorInput() instanceof FileEditorInput) {
-					IFile editorFile = ((FileEditorInput) editor.getEditorInput()).getFile();
-					IContentType contentType = contentTypeToLSDefinition.getKey();
-					LanguageServerDefinition lsDefinition = contentTypeToLSDefinition.getValue();
-					IContentDescription contentDesc = editorFile.getContentDescription();
-					if (contentTypeToLSDefinition.isEnabled() && contentType != null && contentDesc != null
-							&& contentType.equals(contentDesc.getContentType()) && lsDefinition != null) {
-						try {
-							getInitializedLanguageServer(editorFile, lsDefinition, capabilities -> true);
-						} catch (IOException e) {
-							LanguageServerPlugin.logError(e);
-						}
+					final IFile editorFile = ((FileEditorInput) editor.getEditorInput()).getFile();
+					final IContentDescription contentDesc = editorFile.getContentDescription();
+					if(contentDesc == null)
+						continue;
+					if (contentType.equals(contentDesc.getContentType()) && contentTypeToLSDefinition.isEnabled()) {
+						getInitializedLanguageServer(editorFile, lsDefinition, capabilities -> true);
 					}
 				}
-			} catch (CoreException e) {
+			} catch (Exception e) {
 				LanguageServerPlugin.logError(e);
 			}
 		}
-
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServiceAccessor.java
```diff
@@ -279,110 +274,130 @@ private static boolean capabilitiesComply(LanguageServerWrapper wrapper,
...
+					continue;
 				}
-				if (contentType.getBaseType() != null) {
-					contentTypes.add(contentType.getBaseType());
+
+				synchronized (startedServers) {
+					// check again while holding the write lock
+					final var matchingServers = startedServers.stream() //
+							.filter(serverSelector) //
+							.filter(serverSelector2)
```
rubenporras, line 380: can we name serverSelector and  serverSelector2 in a more meaningful way that is not 1 and 2?

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServiceAccessor.java
```diff
@@ -279,110 +274,130 @@ private static boolean capabilitiesComply(LanguageServerWrapper wrapper,
...
-		res.addAll(getMatchingStartedWrappers(file, request));
+		final var res = new LinkedHashSet<LanguageServerWrapper>(getMatchingStartedWrappers(file, request));
 
 		// look for running language servers via content-type
-		Queue<IContentType> contentTypes = new LinkedList<>();
-		Set<IContentType> addedContentTypes = new HashSet<>();
-		contentTypes.addAll(LSPEclipseUtils.getFileContentTypes(file));
-		addedContentTypes.addAll(contentTypes);
+		final var contentTypes = new ArrayDeque<IContentType>(LSPEclipseUtils.getFileContentTypes(file));
+		final var processedContentTypes = new HashSet<IContentType>();
```
rubenporras, line 294: could we save LSPEclipseUtils.getFileContentTypes(file) into a local variable so that it can be used to initialize the capacity of processedContentTypes?

rubenporras:
I like the idea of the idea of the change and it looks good. Shall we merge it and gain some testing?

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServiceAccessor.java
```diff
@@ -174,8 +171,7 @@ public static void enableLanguageServerContentType(
 					LanguageServerDefinition lsDefinition = contentTypeToLSDefinition.getValue();
 					IContentDescription contentDesc = editorFile.getContentDescription();
 					if (contentTypeToLSDefinition.isEnabled() && contentType != null && contentDesc != null
-							&& contentType.equals(contentDesc.getContentType())
-							&& lsDefinition != null) {
+							&& contentType.equals(contentDesc.getContentType()) && lsDefinition != null) {
```
rubenporras, line 174: the code was already like this, but `lsDefinition != null` can be already checked before calling `editorFile.getContentDescription();`

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServiceAccessor.java
```diff
@@ -279,110 +274,128 @@ private static boolean capabilitiesComply(LanguageServerWrapper wrapper,
...
 
-		res.addAll(getMatchingStartedWrappers(file, request));
+		final var res = new LinkedHashSet<LanguageServerWrapper>(getMatchingStartedWrappers(file, request));
 
 		// look for running language servers via content-type
-		Queue<IContentType> contentTypes = new LinkedList<>();
-		Set<IContentType> addedContentTypes = new HashSet<>();
-		contentTypes.addAll(LSPEclipseUtils.getFileContentTypes(file));
-		addedContentTypes.addAll(contentTypes);
+		final var contentTypes = new ArrayDeque<IContentType>(LSPEclipseUtils.getFileContentTypes(file));
```
rubenporras, line 293: Nitty gritty: not sure how much of a difference it is, but `LSPEclipseUtils.getFileContentTypes(file)` returns an ArrayList which has the size precomputed, while the ArrayDeque must calculate the value.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServiceAccessor.java
```diff
@@ -279,110 +274,128 @@ private static boolean capabilitiesComply(LanguageServerWrapper wrapper,
...
-					}
+
+			for (final ContentTypeToLanguageServerDefinition mapping : lsRegistry.findProviderFor(contentType)) {
+				if (mapping == null || !mapping.isEnabled()) {
+					continue;
+				}
+				final LanguageServerDefinition serverDefinition = mapping.getValue();
+				if (serverDefinition == null) {
+					continue;
+				}
```
rubenporras, line 310: this code is repeated below, maybe extract it to a method 
```
isValidMapping(ContentTypeToLanguageServerDefinition mapping) {
                                if (mapping == null || !mapping.isEnabled()) {
					return false;
				}
				final LanguageServerDefinition serverDefinition = mapping.getValue();
				if (serverDefinition == null) {
					return false;
				}
}
```
or similar?

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServiceAccessor.java
```diff
@@ -420,73 +433,79 @@ public static LanguageServerWrapper getLSWrapperForConnection(@NonNull IProject
...
+			return wrapper;
 		}
-		return wrapper;
 	}
 
 	private static LanguageServerWrapper getLSWrapperForConnection(@NonNull IDocument document,
 			@NonNull LanguageServerDefinition serverDefinition, @Nullable IPath initialPath) throws IOException {
-		LanguageServerWrapper wrapper = null;
+
+		final Predicate<LanguageServerWrapper> serverSelector = w -> w.canOperate(document);
```
rubenporras, line 465: are we not missing the condition `&& w.serverDefinition.equals(serverDefinition)` ? According to my diff, it is no longer there. Or is it by intention?

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServiceAccessor.java
sebthom: I gave it a try it realized that the serverDefinition variable is used by the subsequent code and thus there isn't much saved.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServiceAccessor.java
sebthom: > are we not missing the condition `&& w.serverDefinition.equals(serverDefinition)` ? According to my diff, it is no longer there. Or is it by intention?

Man, you are thorough. Good catch, appreciated!!

mickaelistria:
I don't think I'm really good at evaluating thread-safety to give a useful review. But I do see more final variables, some earlier returns, usage of `removeIf` instead of iterations... So overall I have a positive feeling about it, and can happily trust you if you think it's good to merge.

sebthom:
> I don't think I'm really good at evaluating thread-safety to give a useful review. But I do see more final variables, some earlier returns, usage of `removeIf` instead of iterations... So overall I have a positive feeling about it, and can happily trust you if you think it's good to merge.

@mickaelistria the key change is in the [third commit](https://github.com/eclipse/lsp4e/pull/167/commits/9658510dc024f263897a62b7fd120cc5901c033c?w=1). where the synchronized blocks around `startedServers` are removed, this reduces the risk of the deadlock we encounter.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServiceAccessor.java
rubenporras: ACK

rubenporras (approved):
I also cannot really tell if there are some problems with the change or not, but my gut feeling is that it goes in the right direction, and I checked it as good as I could, so from my side, you can merge it as well if you like it :)

PR #169, opened 2022-06-29
add github actions builds

sebthom:


mickaelistria:
At the moment, we're satisfied with Jenkins builds. What value would this extra build offer that makes it worth the extra maintenance?

rubenporras:
for what is worth, I also think like Sebastian that it would be very nice to CI feedback without creating a PR to the source repo. Usually I create a PR just to get the CI feedback, even when changes are not ready.

org.eclipse.lsp4e.tests.mock/META-INF/MANIFEST.MF
```diff
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: Mock Language Server to test LSP4E
 Bundle-SymbolicName: org.eclipse.lsp4e.tests.mock
-Bundle-Version: 0.14.4.qualifier
+Bundle-Version: 0.14.5.qualifier
```
mickaelistria, line 5: This shouldn't be necessary. Please revert and consider the version bump separately if really needed.

LICENSE
```diff
@@ -0,0 +1,277 @@
+Eclipse Public License - v 2.0
```
mickaelistria, line 1: Would be nice to add it in a dedicated PR.

mickaelistria (changes requested):

org.eclipse.lsp4e.jdt/META-INF/MANIFEST.MF
```diff
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: JDT Integration for LSP4E
 Bundle-SymbolicName: org.eclipse.lsp4e.jdt;singleton:=true
-Bundle-Version: 0.10.2.qualifier
+Bundle-Version: 0.10.3.qualifier
```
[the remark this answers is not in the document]
sebthom: That is the same issue we talked about in the other tm4e issue. it looks like the moment a file is modified in the repo root the build fails complaining about the versions not being bumped. I have no idea how to solve this.

LICENSE
sebthom: I can make a separate PR, but thought a dedicated commit is sufficient because the change is clearly visible in the git history.

org.eclipse.lsp4e.jdt/META-INF/MANIFEST.MF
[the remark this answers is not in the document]
mickaelistria: That's not how it's supposed to work. I'll investigate.

org.eclipse.lsp4e.tests.mock/META-INF/MANIFEST.MF
mickaelistria: I cannot reproduce a failure with a local build. Please submit the PR without changing versions to see how it does behave on CI.

org.eclipse.lsp4e.jdt/META-INF/MANIFEST.MF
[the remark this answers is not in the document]
sebthom: I reverted the version bumps, now the github action fails with
`Only qualifier changed for (org.eclipse.lsp4e.jdt/0.10.2.202207051450). Expected to have bigger x.y.z than what is available in baseline (0.10.2.202205031750)`  see https://github.com/sebthom/lsp4e/runs/7198694165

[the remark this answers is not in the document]
sebthom: @mickaelistria that's the info I needed. the gha build no longer creates a shallow clone and now succeeds without bumping the versions.

sebthom:
@rubenporras I am getting test failures in github builds. Any ideas why? https://github.com/sebthom/lsp4e/runs/7600681404?check_suite_focus=true  I am using a similar github action config for building/testing TM4E where it works fine.

org.eclipse.lsp4e.test/src/org/eclipse/lsp4e/test/LanguageServiceAccessorTest.java
```diff
@@ -434,15 +434,15 @@ public void testSingletonLS() throws Exception {
 		var testFile1 = createFile(project, "shouldUseSingletonLS.lsp-singletonLS", "");
 		IDocument document1 = getDocument(testFile1);
 		assertNotNull(document1);
-		var languageServers = getLanguageServers(document1, MATCH_ALL);
+		var languageServers = getLanguageServers(document1, MATCH_ALL).get();
```
mickaelistria, line 437: I don't think this should be necessary here. The choice of using a singleton LS doesn't require that the LanguageServer is ready; this shouldn't be something we need for the test to pass. If this is required for the test to pass, then it probably shows underlying code has some bug.

org.eclipse.lsp4e.test/src/org/eclipse/lsp4e/test/TestUtils.java
```diff
@@ -237,21 +237,28 @@ public static IEditorReference[] getEditors() {
 		return null;
 	}
 
+	public static boolean isGitHubActions() {
+		return "true".equals(System.getenv("GITHUB_ACTIONS"));
```
mickaelistria, line 241:
```suggestion
		return Boolean.getBoolean("GITHUB_ACTIONS");
```

sebthom: Will do. thanks!

org.eclipse.lsp4e.test/src/org/eclipse/lsp4e/test/LanguageServiceAccessorTest.java
sebthom: No, it is not necessary. I am currently just trying out things as I am getting a lot of very random test failures on GHA.

akurtakov:
Is this smth someone still want to finish?

sebthom:
Yes, the problem is there are still random build failures.

sebthom:
I spent a lot of time trying to get the jobs running without test failures on Windows and Ubuntu but so far there are still random test failures. Only MacOS builds are stable so I commented out Windows/Ubuntu runners for now until we find a way to stabilize the tests. Even with only mac builds I think this is a valuable addition.

mickaelistria:

PR #170, opened 2022-06-30
Cycling "LSP4E Linked Editing" job when selecting attribute value #838

vrubezhny:


mickaelistria (approved):
Change seems good to me, however I didn't try it. If you tested it and are confident it's woring well, feel free to merge.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/linkedediting/LSPLinkedEditingReconcilingStrategy.java
```diff
@@ -248,6 +252,21 @@ String getValueInRange(LinkedEditingRanges ranges, VerifyEvent event, int offset
 		return null;
 	}
 
+	private boolean rangesContainOffset(LinkedEditingRanges ranges, int offset) {
+		try {
+			for (Range range : ranges.getRanges()) {
+				int start = LSPEclipseUtils.toOffset(range.getStart(), fDocument);
+				int end = LSPEclipseUtils.toOffset(range.getEnd(), fDocument);
```
rubenporras, line 259: if we would move  LSPEclipseUtils.toOffset(range.getEnd(), fDocument) into the if we would benefit from lazy evaluation.
the same would apply to the existing logic in getValueInRange.
Maybe factor out to a method

boolean rangeContainOffset(Range range, int offset) 

and reuse it?

sebthom: I second this. Big fan of lazy evaluation myself :-)

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/linkedediting/LSPLinkedEditingReconcilingStrategy.java
```diff
@@ -248,6 +252,21 @@ String getValueInRange(LinkedEditingRanges ranges, VerifyEvent event, int offset
...
+					return true;
+				}
+			}
+		} catch (BadLocationException e) {
+			LanguageServerPlugin.logError(e);
+		}
+		return false;
+	}
+
 	private LinkedPositionGroup toJFaceGroup(LinkedEditingRanges ranges) throws BadLocationException {
```
rubenporras, line 270: also, in this method I believe we can save the offset computation for the end, and instead of
```
		for (Range range : ranges.getRanges()) {
			int startOffset = LSPEclipseUtils.toOffset(range.getStart(), fDocument);
			int length = LSPEclipseUtils.toOffset(range.getEnd(), fDocument) - startOffset;
			res.addPosition(new LinkedPosition(fDocument, startOffset, length));
		}
		return res;
```
just do
```
		for (Range range : ranges.getRanges()) {
			int startOffset = LSPEclipseUtils.toOffset(range.getStart(), fDocument);
			int length = range.getEnd() - range.getStart();
			res.addPosition(new LinkedPosition(fDocument, startOffset, length));
		}
		return res;
```

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/linkedediting/LSPLinkedEditingReconcilingStrategy.java
vrubezhny: Agree

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/linkedediting/LSPLinkedEditingReconcilingStrategy.java
```diff
@@ -248,6 +252,24 @@ String getValueInRange(LinkedEditingRanges ranges, VerifyEvent event, int offset
...
+					return true;
+				}
+			}
+		} catch (BadLocationException e) {
+			LanguageServerPlugin.logError(e);
+		}
+		return false;
+	}
+
+	private boolean rangeContainsOffset(Range range, int offset) throws BadLocationException {
```
rubenporras, line 268: I have realized that this method already exists, isn't it the same as `LSPEclipseUtils.isOffsetInRange(int, Range, IDocument)`, could we just use that one?

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/linkedediting/LSPLinkedEditingReconcilingStrategy.java
rubenporras: Yes, you are right, we cannot do the `-` on `Position` elements

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/linkedediting/LSPLinkedEditingReconcilingStrategy.java
vrubezhny: Yes, you're right. Thanks for the catch!

vrubezhny:
> Change seems good to me, however I didn't try it. If you tested it and are confident it's woring well, feel free to merge.

Yes, I have tested it manually, I believe it works well.
 The problem wasn't appearing  in HTML Editor, while for XML Editor the linked regions were returned even if the text cursor is placed into a position outside of element name. Now we're filtering out the position, starting the update job only if position is inside element name.

PR #174, opened 2022-07-05
Backport/format concurrency

ahmedneilhussain:
This is the concurrency fix for format which I resubmitted the other week having split it out of the original set of concurrency fixes. I withdrew it and have now revised it to allow the consuming code to detect when the formatting edits were provided by the server based on file content which has since changed. Merely ensuring calls to the server are sent in sequence wrt a given document is not enough, since the UI thread remains unblocked and can have further updated the document. The solution amounts to an optimistic locking model - the code that has requested formatting will be submitting the edits on the UI thread and can thus check whether the content is unchanged from the version of the document on which the format request was dispatch.

Added a couple of test cases to illustrate this.

### DIFF
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/DocumentContentSynchronizer.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/DocumentContentSynchronizer.java
@@ -73,6 +73,7 @@ final class DocumentContentSynchronizer implements IDocumentListener {
 	private DidChangeTextDocumentParams changeParams;
 	private long modificationStamp;
 	private CompletableFuture<LanguageServer> lastChangeFuture;
+	private final CompletableFuture<LanguageServer> didOpenFuture;
 	private final Object guard = new Object();
 	private long documentModificationStamp = 0L;
 	private LanguageServer languageServer;
@@ -120,7 +121,7 @@ final class DocumentContentSynchronizer implements IDocumentListener {
 
 		textDocument.setLanguageId(languageId);
 		textDocument.setVersion(++version);
-		lastChangeFuture = languageServerWrapper.getInitializedServer().thenApplyAsync(ls -> {
+		didOpenFuture = lastChangeFuture = languageServerWrapper.getInitializedServer().thenApplyAsync(ls -> {
 			this.languageServer = ls;
 			ls.getTextDocumentService().didOpen(new DidOpenTextDocumentParams(textDocument));
 			return ls;
@@ -147,8 +148,8 @@ final class DocumentContentSynchronizer implements IDocumentListener {
 		}
 	}
 
-	CompletableFuture<LanguageServer> lastChangeFuture() {
-		return lastChangeFuture;
+	CompletableFuture<LanguageServer> didOpenFuture() {
+		return didOpenFuture;
 	}
 
 	@Override
@@ -275,7 +276,7 @@ final class DocumentContentSynchronizer implements IDocumentListener {
 		// Use @link{TextDocumentSaveReason.Manual} as the platform does not give enough information to be accurate
 		WillSaveTextDocumentParams params = new WillSaveTextDocumentParams(identifier, TextDocumentSaveReason.Manual);
 
-
+		List<TextEdit> edits = null;
 		try {
 			edits = executeOnCurrentVersionAsync(ls -> ls.getTextDocumentService().willSaveWaitUntil(params))
 				.get(lsToWillSaveWaitUntilTimeout(), TimeUnit.SECONDS).get();
@@ -292,6 +293,11 @@ final class DocumentContentSynchronizer implements IDocumentListener {
 		} catch (InterruptedException e) {
 			LanguageServerPlugin.logError(e);
 			Thread.currentThread().interrupt();
+		}
+		try {
+			LSPEclipseUtils.applyEdits(document, edits);
+		} catch (BadLocationException e) {
+			LanguageServerPlugin.logError(e);
 		}
 	}
 
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/DocumentContentSynchronizer.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/DocumentContentSynchronizer.java
@@ -72,10 +72,22 @@ final class DocumentContentSynchronizer implements IDocumentListener {
 	private int version = 0;
 	private DidChangeTextDocumentParams changeParams;
 	private long modificationStamp;
-	private CompletableFuture<LanguageServer> lastChangeFuture;
 	private final CompletableFuture<LanguageServer> didOpenFuture;
+
+	/**
+	 * Ensures atomic read/write of <code>documentModificationStamp</code> and
+	 * <code>lastChangeFuture</code>
+	 */
 	private final Object guard = new Object();
+	/**
+	 * Caches (opaque) version stamp of last document update event sent to the server
+	 */
 	private long documentModificationStamp = 0L;
+	/** Caches handle to the last call made to the server for this document, allowing
+	 * calls to be chained sequentially. Protected by <code>this.guard</code>
+	 */
+	private CompletableFuture<LanguageServer> lastChangeFuture;
+
 	private LanguageServer languageServer;
 	private IPreferenceStore store;
 
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/DocumentContentSynchronizer.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/DocumentContentSynchronizer.java
@@ -24,7 +24,6 @@ import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
-import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.Function;
 
 import org.eclipse.core.filesystem.EFS;
@@ -37,6 +36,7 @@ import org.eclipse.core.runtime.content.IContentType;
 import org.eclipse.jdt.annotation.NonNull;
 import org.eclipse.jface.preference.IPreferenceStore;
 import org.eclipse.jface.text.BadLocationException;
+import org.eclipse.jface.text.Document;
 import org.eclipse.jface.text.DocumentEvent;
 import org.eclipse.jface.text.IDocument;
 import org.eclipse.jface.text.IDocumentListener;
@@ -72,7 +72,9 @@ final class DocumentContentSynchronizer implements IDocumentListener {
 	private int version = 0;
 	private DidChangeTextDocumentParams changeParams;
 	private long modificationStamp;
-	private final AtomicReference<CompletableFuture<LanguageServer>> lastChangeFuture;
+	private CompletableFuture<LanguageServer> lastChangeFuture;
+	private final Object guard = new Object();
+	private long documentModificationStamp = 0L;
 	private LanguageServer languageServer;
 	private IPreferenceStore store;
 
@@ -96,6 +98,9 @@ final class DocumentContentSynchronizer implements IDocumentListener {
 		this.document = document;
 		this.store = LanguageServerPlugin.getDefault().getPreferenceStore();
 
+		this.documentModificationStamp = document instanceof Document ? ((Document)document).getModificationStamp()
+				: Document.UNKNOWN_MODIFICATION_STAMP;
+
 		// add a document buffer
 		TextDocumentItem textDocument = new TextDocumentItem();
 		textDocument.setUri(fileUri.toString());
@@ -115,11 +120,11 @@ final class DocumentContentSynchronizer implements IDocumentListener {
 
 		textDocument.setLanguageId(languageId);
 		textDocument.setVersion(++version);
-		lastChangeFuture = new AtomicReference<>(languageServerWrapper.getInitializedServer().thenApplyAsync(ls -> {
+		lastChangeFuture = languageServerWrapper.getInitializedServer().thenApplyAsync(ls -> {
 			this.languageServer = ls;
 			ls.getTextDocumentService().didOpen(new DidOpenTextDocumentParams(textDocument));
 			return ls;
-		}));
+		});
 	}
 
 	/**
@@ -130,20 +135,20 @@ final class DocumentContentSynchronizer implements IDocumentListener {
  	 * @param fn Asynchronous computation on the language server
  	 * @return Asynchronous result object.
  	 */
-	<U> @NonNull CompletableFuture<U> executeOnCurrentVersionAsync(
+	<U> @NonNull CompletableFuture<VersionedResult<U>> executeOnCurrentVersionAsync(
 			Function<LanguageServer, ? extends CompletionStage<U>> fn) {
-		AtomicReference<CompletableFuture<U>> resValueFuture = new AtomicReference<>();
-		lastChangeFuture.updateAndGet(f -> {
-			CompletableFuture<U> valueFuture = f.thenComposeAsync(fn);
-			resValueFuture.set(valueFuture);
+		synchronized (this.guard) {
+			final long stampAtTimeExecuted = this.documentModificationStamp;
+			CompletableFuture<VersionedResult<U>> valueFuture = this.lastChangeFuture.
+					thenComposeAsync(fn).thenApply(u -> new VersionedResult<U>(u, stampAtTimeExecuted));
 			// We ignore any exceptions that happen when executing the given future
-			return valueFuture.handle((value, error) -> this.languageServer);
-		});
-		return resValueFuture.get();
+			this.lastChangeFuture = valueFuture.handle((value, error) -> this.languageServer);
+			return valueFuture;
+		}
 	}
 
 	CompletableFuture<LanguageServer> lastChangeFuture() {
-		return lastChangeFuture.get();
+		return lastChangeFuture;
 	}
 
 	@Override
@@ -157,11 +162,14 @@ final class DocumentContentSynchronizer implements IDocumentListener {
 			final DidChangeTextDocumentParams changeParamsToSend = changeParams;
 			changeParams = null;
 
-			changeParamsToSend.getTextDocument().setVersion(++version);
-			lastChangeFuture.updateAndGet(f -> f.thenApplyAsync(ls -> {
-				ls.getTextDocumentService().didChange(changeParamsToSend);
-				return ls;
-			}));
+			synchronized (this.guard) {
+				changeParamsToSend.getTextDocument().setVersion(++version);
+				this.documentModificationStamp = event.getModificationStamp();
+				this.lastChangeFuture = this.lastChangeFuture.thenApplyAsync(ls -> {
+					ls.getTextDocumentService().didChange(changeParamsToSend);
+					return ls;
+				});
+			}
 		}
 	}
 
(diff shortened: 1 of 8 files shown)


sebthom:
@ahmedneilhussain could you please rebase and solve the merge conflict? thanksl

ahmedneilhussain:
> @ahmedneilhussain could you please rebase and solve the merge conflict? thanksl

@sebthom done now. Thanks!

ahmedneilhussain:
Hi @sebthom @mickaelistria is there anything I can do to move this forward? Needs rebasing obviously but that's a moving target!

ahmedneilhussain:
Hi @mickaelistria I think that's now successfully rebased on master. Cheers, Ahmed.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServerWrapper.java
```diff
@@ -553,6 +555,24 @@ public synchronized void stop() {
...
+	/**
+	 * Submit an asynchronous call (i.e. to the language server) that be inserted into the current position w.r.t.
+	 * document change events
+	 *
+	 * @param <U> Computation return type
+	 * @param file Document to serialise on
+	 * @param fn Asynchronous computation on the language server
+	 * @return Asynchronous result object.
+	 */
+	<U> @Nullable CompletableFuture<VersionedResult<U>> executeOnCurrentVersionAsync(URI uri,
```
mickaelistria, line 567: I'm not found of a new API for that as I hope it would be guaranteed internally without requiring particular execution path. Would it be possible to change implementation of `getInitializedServer` so it returns a language server that is guaranteed to have received last document changes?

org.eclipse.lsp4e/src/org/eclipse/lsp4e/VersionedResult.java
```diff
@@ -0,0 +1,45 @@
...
+ *
+ *******************************************************************************/
+package org.eclipse.lsp4e;
+
+/**
+ * Groups a result from the language server with the modification stamp that existed on the document
+ * at the time the request to the server was made
+ * @param <T>
+ */
+public class VersionedResult<T> {
```
mickaelistria, line 20: I don't think it's worth a new API. IMO, this should be verified internally by the particular operation that requires such checks (eg made internal to `LSPFormatter`).

mickaelistria (changes requested):
Please see inline questions/comments/proposals...
I think the overall idea is the right one (ensuring `didChange` was sent before sending any other LS requests). I just wish we can make it simpler for consumer by changing existing internal implementation and not requiring new APIs.
I also noticed some changes that do make sense separately and some parts of this change that are unrelated. If you can manage to split those in different simpler commits/PRs, it would facilitate progress as we could easily close some of those and focus on the harder ones with less disturbance.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/DocumentContentSynchronizer.java
```diff
@@ -72,7 +72,10 @@ final class DocumentContentSynchronizer implements IDocumentListener {
 	private int version = 0;
 	private DidChangeTextDocumentParams changeParams;
 	private long modificationStamp;
-	private final AtomicReference<CompletableFuture<LanguageServer>> lastChangeFuture;
+	private CompletableFuture<LanguageServer> lastChangeFuture;
+	private final CompletableFuture<LanguageServer> didOpenFuture;
```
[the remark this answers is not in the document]
ahmedneilhussain: Sure, I'll add some documentation.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/DocumentContentSynchronizer.java
```diff
@@ -72,7 +72,10 @@ final class DocumentContentSynchronizer implements IDocumentListener {
...
+	private final Object guard = new Object();
+	private long documentModificationStamp = 0L;
```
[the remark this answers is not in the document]
ahmedneilhussain: AIUI `synchronised` is as strong as `volatile` in terms of its memory model guarantees, so that should happen automatically provided it is only accessed from the `synchronized` blocks. See https://jenkov.com/tutorials/java-concurrency/synchronized.html for example.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServerWrapper.java
```diff
@@ -650,7 +670,7 @@ private CompletableFuture<LanguageServer> connect(@NonNull URI uri, IDocument do
 				DocumentContentSynchronizer listener = new DocumentContentSynchronizer(this, theDocument, syncKind);
 				theDocument.addDocumentListener(listener);
 				LanguageServerWrapper.this.connectedDocuments.put(uri, listener);
-				return listener.lastChangeFuture();
+				return listener.didOpenFuture();
```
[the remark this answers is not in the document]
ahmedneilhussain: It's a pedantic point, but I think it better captures the original behaviour and also what I thought the intent was in terms of what `connect()` should be returning - i.e. a hook to allow async requests to be sent to the language server at (or the ability to block until) the point when the server is guaranteed to have been told about the document being opened.

It also goes to the heart of what we're trying to prevent with this change: client code trying to achieve a sequential ordering of interactions with the language server for a particular document, but doing it incorrectly.

If client code cares about ordering then it shouldn't interact with `lastChangeFuture` directly because that doesn't really offer any guarantees.

Possible sequence:

Document modified, with change 'A' sent to the server.

Client retrieves `lastChangeFuture` at time B.

Client async composes operation C on this future.

On another thread (or even on the same thread) something retrieves `lastChangeFuture`, gets the same object as at time B, and async composes operation D on it.

The only thing that is guaranteed is that the server will see change A before C or D is sent. Nothing is guaranteed about whether C or D is sent first, nor about which operation returns first.

Using `lastChangeFuture` doesn't really give any guarantees, other than that the document is (possibly) in a slightly later state than it would have been at `didOpenFuture`, which isn't really any use. If you care about sequencing then you need an API like the one we propose which offers 'after A, but before anything else', if not then just use `didOpenFuture` which just promises that the document has been loaded by the server.

Does that make any kind of sense?

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServerWrapper.java
ahmedneilhussain: Unfortunately I don't think you can guarantee that without blocking the UI thread completely and preventing further changes.

Let's suppose `lastChangeFuture` reflects a notification that has been sent to the server about a document change.

At this point, we compose `lastChangeFuture` with a call to the server to request formatting.

When this completes, it is guaranteed that the server knows about all the changes up to the point where we made the formatting request, so the formatting data returned will be consistent with the server's view of the document *at that point*. Unfortunately, it doesn't guarantee that the server's view of the document is up to date and no changes have been made - just that the server hasn't been told about them yet!

If we use the changes returned in the formatting request future `andThen` schedule an update on the UI thread, it is entirely possible corruption will take place, because the server's view of the document used to suggest the formatting edits is no longer accurate.

The change proposed here guards against this by allowing consuming code to detect when the document on which it wants to apply changes is inconsistent with the server's view of the document when computing those changes. Essentially this is an optimistic locking model - we can't easily prevent inconsistencies without locking threads and preventing changes. Instead we presume that mostly changes won't interfere with one another, but we can check when this is about to happen and avoid making a conflicting change (essentially a lost update).

org.eclipse.lsp4e/src/org/eclipse/lsp4e/VersionedResult.java
ahmedneilhussain: That's what this API is designed to do. It doesn't oblige consuming code to take version conflicts into account, it just allows them to do so.

The only thing that knows what version of the document the server will have seen when a request (e.g. a format request) is sent, is the `DocumentContentSynchronizer` for that document. We need a wrapper object like this to bundle the version and the result together to make the version check atomic and consistent. Otherwise client code would have to make two separate calls to `DocumentContentSynchronizer`.

If `T` is some edits, then when we schedule those on the event thread, we can read the document version (knowing it can only change on the event thread) and compare it to `VersionedResult.getStamp()`. If they're different, then the results are stale and the client code can either abandon the changes, apply them anyway (!), or restart the operation (the classic way of handling a write consistency clash).

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServerWrapper.java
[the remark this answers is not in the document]
mickaelistria: > The only thing that is guaranteed is that the server will see change A before C or D is sent

Right, but isn't it a good thing to guarantee? If I understand correctly this particular part of the proposal (using didOpen instead of lastChange) does remove this guarantee. I fail at seeing how it can improve some cases. Can you please provide insights about what is better with this?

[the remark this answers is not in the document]
ahmedneilhussain: Well, currently I think it is only invoked by `LanguageServerWrapper.connect(...)` which in turn is just used to ensure the document has been connected. If so, using `didOpen` allows this guarantee to be met earlier.

If client code needs a 'happens after' guarantee then `executeOnCurrentVersionAsync` also does that, but offers the much stronger ability to ensure point-in-time consistency (or rather, to detect point-in-time inconsistency!). Since we're not keen on exposing more API than we have to I thought it made sense to encourage controlled access via the new method.

I guess what exposing `lastChange` would do is allow someone to schedule a potentially long-running call without forcing change notifications to queue behind it, but as noted above nothing was actually using that.

What do you think? I guess I don't feel that strongly about it, so it's down to what we think needs exposing in an API...

org.eclipse.lsp4e/src/org/eclipse/lsp4e/DocumentContentSynchronizer.java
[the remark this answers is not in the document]
ahmedneilhussain: Done

org.eclipse.lsp4e/src/org/eclipse/lsp4e/DocumentContentSynchronizer.java
```diff
@@ -130,20 +136,20 @@ public DocumentContentSynchronizer(@NonNull LanguageServerWrapper languageServer
...
-		});
-		return resValueFuture.get();
+			this.lastChangeFuture = valueFuture.handle((value, error) -> this.languageServer);
+			return valueFuture;
+		}
 	}
 
-	CompletableFuture<LanguageServer> lastChangeFuture() {
-		return lastChangeFuture.get();
+	CompletableFuture<LanguageServer> didOpenFuture() {
```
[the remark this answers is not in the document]
ahmedneilhussain: Do we expose immutable fields anywhere else? I tend to fall into old habits of using getters other than for classes that are `record`-style immutable POJOs but that's just superstition I guess!

More importantly, do we want this to be `public` rather than package private?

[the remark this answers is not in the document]
mickaelistria: > Do we expose immutable fields anywhere else?

Yes. See for example how LanguageServerWrapper just has a public final field for the ServerDefinition and how this server definition also has final fields.

> More importantly, do we want this to be public rather than package private?

The most restricted, the better, so package would be ok.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServerWrapper.java
mickaelistria: OK, so that brings another question: do we really want to get the initialized language server for a particular document without having all the monitored document relatively up-to-date?
Basically is there a reason to prefer current `getInitializedServer` over some newer `getUpToDateServer` ?

org.eclipse.lsp4e/src/org/eclipse/lsp4e/VersionedResult.java
mickaelistria: If we have some `getUpToDateServer` as discussed above, would we still need such an API or would it be good enough?

ahmedneilhussain: Unfortunately you do need this kind of API to guarantee correctness. Without blocking the UI thread you can't ever guarantee that the server is up to date, only that it is consistent with all the changes it has received _so far_. Hence the need for an API which is effectively `computeBasedOnWhatYouKnowAndLetMeFigureOutWhetherItIsStale`.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServerWrapper.java
ahmedneilhussain: I don't think `getUpToDateServer` is possible without blocking updates from happening, and unfortunately relatively-up-to-date isn't enough to be bullet-proof against possible document corruption. See below.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/DocumentContentSynchronizer.java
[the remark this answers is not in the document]
ahmedneilhussain: OK, I can change to a package-level final field.

mickaelistria:
I'm still brainstorming a lot about this problem as I believe it's a core issue and an important pattern we'll need to setup, which I think deserves it.
Going back to the particular case of Format, what do you think would make the following in FormatHandler not sufficient:

```java
//  CURRENTLY IS
formatter.requestFormatting(document, textSelection).thenAcceptAsync(
	edits -> shell.getDisplay().asyncExec(() -> formatter.applyEdits(document, edits)));
// PROPOSAL
if (IDocument instanceof Document doc) {
	final int requestStamp = doc.getModificationStamp();
	formatter.requestFormatting(document, textSelection).thenAcceptAsync(
		edits -> shell.getDisplay().asyncExec(
			() -> if (doc.getModificationStamp() == requestStamp) {
				formatter.applyEdits(document, edits)
		})));
```

ahmedneilhussain:
> I'm still brainstorming a lot about this problem as I believe it's a core issue and an important pattern we'll need to setup, which I think deserves it. Going back to the particular case of Format, what do you think would make the following in FormatHandler not sufficient:
> 
> ```java
> //  CURRENTLY IS
> formatter.requestFormatting(document, textSelection).thenAcceptAsync(
> 	edits -> shell.getDisplay().asyncExec(() -> formatter.applyEdits(document, edits)));
> // PROPOSAL
> if (IDocument instanceof Document doc) {
> 	final int requestStamp = doc.getModificationStamp();
> 	formatter.requestFormatting(document, textSelection).thenAcceptAsync(
> 		edits -> shell.getDisplay().asyncExec(
> 			() -> if (doc.getModificationStamp() == requestStamp) {
> 				formatter.applyEdits(document, edits)
> 		})));
> ```

That isn't guaranteed to be correct as things stand because requestStamp could include a document modification that has been scheduled but not yet sent to the language server, i.e. the following ordering is still possible

document changed -> async notification enqueued on thread pool -> formathandler reads version and submits formatting request -> formatting request enqueued -> formatting request sent -> server computes formatting -> server receives update notification.

As things stand only thing that can track what version of a document the server knows about is the `DocumentContentSynchronizer`.

ahmedneilhussain:
Thanks @mickaelistria , very glad this is receiving careful scrutiny.... it's very easy to overlook subtle problems with this kind of change.

As you say it's a wide-ranging consideration, but I think that it's not actually causing many issues at the moment because of the types of server-initiated change that can currently arise and how they are exposed in practice. For example, refactoring/rename [which would potentially add problems of synchronising multiple documents!!!] is initiated via a modal dialog. That prevents the user from making any further changes and gives plenty of time for any pending notifications to be flushed to the server...

mickaelistria:
> the following ordering is still possible: document changed -> async notification enqueued on thread pool -> formathandler reads version and submits formatting request -> formatting request enqueued -> formatting request sent -> server computes formatting -> server receives update notification.

I believe we could fix that in LanguageServiceAccessor, in a way that would profit to every action; if we replace calls to `LanguageServerWrapper::getInitializedServer` by some calls to `LanguageServerWrapper::getUpToDateServer` that would take care of ensuring the `didChange` were sent by querying the various DocumentContentSynchronizer.

Something like
```java
getInitializedServer().thenRun(() -> CompletableFuture.allOf(connectedDocuments.values().stream().map(DocumentConstentSynchronizer::lastChangeWasSent))).thenApply(theVoid -> getInitializedServer())
```
And Document.lastChangeSent would be something like
```java
CompletableFuture<?> lastChangeWasSent() {
  return CompletableFuture.runAsync(() -> {
      do {
         synchronized (guard) {
           if (document.getModificationStamp() == lastModificationStamp) {
              return;
           }
           Thread.sleep(10); // active wait loop can be replace with something smarter involving locks/wait
         }
    };
}
```

ahmedneilhussain:
Unfortunately I don't think that will work. You can wait for all existing edits to have been dispatched but that won't prevent any *further* edits happening while the server is processing your request.

mickaelistria:
> You can wait for all existing edits to have been dispatched but that won't prevent any further edits happening while the server is processing your request.

But that we cannot prevent it at all...

ahmedneilhussain:
> > You can wait for all existing edits to have been dispatched but that won't prevent any further edits happening while the server is processing your request.
> 
> But that we cannot prevent it at all...

Exactly. (Well, not true: we could block edits/halt the event thread. But that would be counter to the whole async/future-based design of the plugin). But this is why I went for a design along the lines I did: given the architecture, trying to wait for the server's view to be completely up to date is doomed - we're chasing a moving target and with no way of making the target stop!

That's why I thought the solution is to make sure we detect when the server has supplied information that is no longer current. It's comparatively easy to do with the way LSP seems intended to work. The editor (us, on the client side), owns the Single Point of Truth about the contents of the document; the server passively observes the state and will answer queries (such as the edits required to format a document) based on what it knows to date. When the server suggests edits to accomplish some operation, there is no damage if we choose not to make them, because the server shouldn't be updating its view of the document based on the assumption they will be made, it should just wait and receive the updates eventually in the same way as if the user typed them in manually.

Remember that we're just trying to guarantee correctness in certain rare but possible scenarios. Or more specifically to prevent accidental corruption! Mostly things are fine because language servers can respond faster than most users can type.

mickaelistria:
I think a combination of https://github.com/eclipse/lsp4e/pull/174#issuecomment-1246942611 and https://github.com/eclipse/lsp4e/pull/174#issuecomment-1247774116 does cover your concerns. Indeed, if we have both, we are certain that the call to `document.getModificationStamp` before `lastChangeWasSent` will return a stamp/change that is no guaranteed to be received by server.

ahmedneilhussain:
Hi @mickaelistria, sorry for the delay in responding. I'm looking at a slightly different approach following a discussion with my boss, to see if I can address your concerns.

Just to clarify, the main objectives that have emerged in the above discussion would be:
1. To adhere to the original objective of this PR, which is to be able to have a robust ordering of requests so that we don't e.g. ask the server for formatting on regions it doesn't think exist, nor apply stale edits to a document that has changed since the server computed them.
2. To expose these guarantees in a simpler form, one which is easier to use and which the rest of the actions can profit from.

Is that a fair summary?

Having dug through the layers to see exactly how we interact with LSP4j and the way the current threading model works, I've observed that LSP4e is possibly more asynchronous than it needs to be in some places. A lot of the complexity in the various implementations results from trying to put an ordered layer on top of an unordered one, so we think we can make it simpler but non-blocking by switching this around a little.

I'll probably submit a sketch of this approach in a separate draft PR just to keep the discussion simpler.

ahmedneilhussain:
On a related point, there's quite a bit of logic that is based around the idea that there may be multiple language servers with different (possibly complementary, possibly overlapping) capabilities active for a given document. To what extent is this flexibility in use?

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LSPEclipseUtils.java
```diff
@@ -435,6 +437,19 @@ public static void applyEdits(IDocument document, List<? extends TextEdit> edits
 		}
 	}
 
+	public static void applyEditsWithVersionCheck(IDocument document, VersionedResult<List<? extends TextEdit>> versionedEdits) throws BadLocationException {
```
rubenporras, line 440: I think any call to applyEdits is affected by the concurrency problem, isn't it? If so, should we not change applyEdits to use versionedEdits?

ahmedneilhussain: Potentially anything that gets a list of edits from the server is vulnerable to the documents having changed by the time they receive and act on the response, yes. I added support for detecting and avoiding this as a general API that existing code can leverage, as you see, but didn't search through all existing uses of `applyEdits` to fix them as well.

As I mention above, I think in practice the risk is lessened in many cases because of the way the functionality is actually exposed to the user - e.g. rename uses a modal dialog, so the user is blocked from making further changes to the documents while the server is computing the rename delta.

ahmedneilhussain:
Much of this work has been generalised and improved as part of #344 and others, so this PR can be withdrawn as superseded by an implementation on top of #344  and #308

Issue #172, opened 2022-07-04
Improve logging of java.util.concurrent.TimeoutException?


rubenporras:
At the moment timeout exceptions are logged as any other exception with the stack trace and no other information. To the end user it could look a bit scary and as if the IDE would be broken, because if the server is not responsive enough, exceptions will rain.

I think we should handle this exceptions generally in a more user friendly way, and just report a warning/info, and include the timeout that was not met. As an example, instead of
```
									} catch (ExecutionException | TimeoutException e) {
										LanguageServerPlugin.logError(e);
										return null;
									}

```

As part of this issue I suggest we better do

```
									} catch (ExecutionException e) {
										LanguageServerPlugin.logError(e);
										return null;
									} catch (TimeoutException  e) {
                                                                                LanguageServerPlugin.logWarning("Could not detect hyperlinks due to timeout after 2 seconds in `textDocument/documentLink`");
```

In the end the timeout is not an error in the sense that usually there is nothing to be fixed on client, but rather in the server.

If you think it is a good idea to handle TimeoutException in this way I would submit a PR.

If you think it should just be an info message or a warning with an stacktrace, I wouldn't mind that either, I actually think the stack trace is not useful at all if the message already tells the endpoint and the timeout, but if you are unsure, we could report the stacktrace  for the time being, just not as an error.

What do you think?

martinlippert:
Without looking at the code to see if this is possible or not, I think it would be awesome to include which language server timed out. Sometimes we have several language servers working on the same file and for that it would be great to know which language server caused the timeout in the log output.

rubenporras:
@martinlippert , thanks, I can take a look if I work on this issue. 

And what about the main topic, that is not logging them as errors and include the timeout and the request endpoint?
What do you think about passing/not passing the stacktrace?

PR #176, opened 2022-07-06
Improve logging of java.util.concurrent.TimeoutException

rubenporras:
If the server is not responsive enough logging errors without any other
context than a stacktrace looks to the end user a bit scary, as if the
IDE would be broken.

Instead reporting a warning with a message that explains the context and
include the timeout that was not met.


rubenporras:
for https://github.com/eclipse/lsp4e/issues/172

sebthom:
@rubenporras these test failures don't look normal:
```python
Errors: 
  CompleteCompletionTest.testTriggerCharsNullList:233 » NullPointer
  CompleteCompletionTest.testTriggerCharsWithoutPreliminaryCompletion:224 » NullPointer
  ContextInformationTest.testTriggerChars:89 » NullPointer
  ContextInformationTest.testTriggerCharsNullList:98 » NullPointer
```

PR #180, opened 2022-07-08
Fix workspace warnings

rubenporras:


sebthom:
Do you think you could you rebase your branch instead of adding a merge commit?

rubenporras:
Hi @sebthom , is it good now?

sebthom:
I'm not sure, but I think the first commit should not appear in the list.

rubenporras:
Now it should be good.

sebthom:
Thanks!!

PR #181, opened 2022-07-08
Keep track of out of order notifications in the LSP Progress Manager

rubenporras:
WorkDoneProgressNotification that arrive before the job has been
started are kept in an orphan, which are checked by the jobs so that 
we do not have orphan jobs which run forever.

### DIFF
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/progress/LSPProgressManager.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/progress/LSPProgressManager.java
@@ -12,9 +12,11 @@
 package org.eclipse.lsp4e.progress;
 
 import java.util.Map;
+import java.util.Set;
 import java.util.concurrent.BlockingQueue;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentSkipListSet;
 import java.util.concurrent.LinkedBlockingDeque;
 import java.util.concurrent.TimeUnit;
 
@@ -38,10 +40,12 @@ public class LSPProgressManager {
 	private final Map<String, BlockingQueue<ProgressParams>> progressMap;
 	private final Map<IProgressMonitor, Integer> currentPercentageMap;
 	private LanguageServer languageServer;
+	private final Set<String> orphanSet;
 
 	public LSPProgressManager() {
 		this.progressMap = new ConcurrentHashMap<>();
 		this.currentPercentageMap = new ConcurrentHashMap<>();
+		this.orphanSet = new ConcurrentSkipListSet<>();
 	}
 
 	public void connect(final LanguageServer languageServer) {
@@ -63,7 +67,10 @@ public class LSPProgressManager {
 			LanguageServerPlugin.logInfo(
 					"Old progress with identifier " + jobIdentifier + " discarded due to new create progress request"); //$NON-NLS-1$//$NON-NLS-2$
 		}
+		return CompletableFuture.runAsync(() -> createJob(languageServer, queue, jobIdentifier));
+	}
 
+	private void createJob(final LanguageServer languageServer, final LinkedBlockingDeque<ProgressParams> queue, final String jobIdentifier) {
 		Job job = Job.create("Language Server Background Job", (ICoreRunnable) monitor -> { //$NON-NLS-1$
 			try {
 				while (true) {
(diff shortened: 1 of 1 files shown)


org.eclipse.lsp4e/src/org/eclipse/lsp4e/progress/LSPProgressManager.java
```diff
@@ -149,10 +157,15 @@ private void report(final WorkDoneProgressReport report, final IProgressMonitor
 	 *            the {@link ProgressParams} used for the progress notification
 	 */
 	public void notifyProgress(final @NonNull ProgressParams params) {
-		String jobName = params.getToken().getLeft();
-		BlockingQueue<ProgressParams> progress = progressMap.get(jobName);
+		String jobIdentifier = params.getToken().getLeft();
```
mickaelistria, line 160:
```suggestion
		String jobIdentifier = params.getToken().map(Function.identity(), Object::toString);
```
and also on line 61.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/progress/LSPProgressManager.java
```diff
@@ -63,7 +67,10 @@ public void connect(final LanguageServer languageServer) {
 			LanguageServerPlugin.logInfo(
 					"Old progress with identifier " + jobIdentifier + " discarded due to new create progress request"); //$NON-NLS-1$//$NON-NLS-2$
 		}
+		return CompletableFuture.runAsync(() -> createJob(languageServer, queue, jobIdentifier));
```
mickaelistria, line 70: I don't think calling `createJob` in a `runAsync()` is a good thing. Jobs are already async, so here we may wait for more time before actually running: the time it takes to the future to be run + the time it takes for the job to be scheduled.
As we do not need the server to wait before returning here (maybe we'll need the server to wait some day on the Job *completion* but not on the Job *creation*), it seems more efficient to call `createJob` and then return directly ` CompletableFuture.completedFuture(null)`

org.eclipse.lsp4e/src/org/eclipse/lsp4e/progress/LSPProgressManager.java
```diff
@@ -38,10 +40,12 @@ public class LSPProgressManager {
 	private final Map<String, BlockingQueue<ProgressParams>> progressMap;
```
mickaelistria, line 40: Is it worth keeping all the params or only the last one? Is there value in processing intermediary reports when we have multiple ones to process.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/progress/LSPProgressManager.java
```diff
@@ -38,10 +40,12 @@ public class LSPProgressManager {
 	private final Map<String, BlockingQueue<ProgressParams>> progressMap;
 	private final Map<IProgressMonitor, Integer> currentPercentageMap;
 	private LanguageServer languageServer;
+	private final Set<String> orphanSet;
```
mickaelistria, line 43: What about just calling it `done` ?

mickaelistria (commented):
Thanks.
I've added some unrelated command because I didn't have the time to review the previous commit properly and now have some more time to consider it. So some comments are not related to that particular change. You're free about how to deal with those, nothing is mandatory.
On a general note, I think it would be better to create a `LSPProgressJob` directly, that would store information about the progress instead of the manager. The manager would simply forward progress notifications to the right job. I believe this would make things simpler.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/progress/LSPProgressManager.java
rubenporras: That is nice, I did not know that API. Thanks.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/progress/LSPProgressManager.java
rubenporras: done

org.eclipse.lsp4e/src/org/eclipse/lsp4e/progress/LSPProgressManager.java
rubenporras: I thought about keeping all messages, but in the end if the progress is still running when the job starts, the job will pick some new report soon and then the work will be reflected. Thus for the time being I would to keep it simple. Maybe later someone  (or me I have time implements that)

rubenporras:
> On a general note, I think it would be better to create a LSPProgressJob directly, that would store information about the progress instead of the manager. The manager would simply forward progress notifications to the right job. I believe this would make things simpler.

@mickaelistria , I will consider that for another PR, for the moment I would like just to avoid endless progress bars.

PR #183, opened 2022-07-11
Fix #182 by changing the styling bits of the shell and text field

### LINKED ISSUE #182: Make the symbol dialogs blends in with rest of quick popups (abstract control styles)
Change both workspace and document symbol popups to looks as follows
Screen Shot 2022-07-11 at 20 32 15

gayanper:
The shell styling is changed to match AbstractInformationControl so the
same styles will get applied from themes. The text style bits are
updated to make it appear same as in different eclipse outline controls.

### DIFF
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/symbols/LSPSymbolInFileDialog.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/symbols/LSPSymbolInFileDialog.java
@@ -37,6 +37,7 @@ import org.eclipse.swt.SWT;
 import org.eclipse.swt.widgets.Composite;
 import org.eclipse.swt.widgets.Control;
 import org.eclipse.swt.widgets.Shell;
+import org.eclipse.swt.widgets.Text;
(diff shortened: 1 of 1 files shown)


sebthom:
@gayanper thanks for your contribution!
would you mind adding two screenshots (before/after) illustrating the effect of your change?

gayanper:
> @gayanper thanks for your contribution!
> would you mind adding two screenshots (before/after) illustrating the effect of your change?

Before



After

gayanper:
@sebthom Any feedback ?

sebthom:
@gayanper LGTM but I would like to leave the decision to @mickaelistria because of his comment at https://github.com/eclipse/lsp4e/issues/182#issuecomment-1188994413

sebthom:
@mickaelistria WDYT?

mickaelistria:
I usually prefer sticking to default settings, and consider those discussions in a more global way. I don't think it's (been) a good thing that each project implements slight tweaks over sticking to default. In the end, it brings more inconsistency and more maintenance effort.

mickaelistria:

Issue #178, opened 2022-07-07
Undo operation fails to correctly restore text 


vrubezhny:
When editing an element name, the Undo operation fails to correctly restore the text. Tested on XML/HTML LS Editors with WWD, 

Steps to reproduce:

- Create an XML or HTML file and open it in Generic Editor, for example:
```
<html>
	<body>
		<h1 class="test">
		</h1>
	</body>
</html>
```
- Place the text cursor after the name of tag `<body|>` (`"|"` char - shows the cursor position)
- Type-in some valid element name character or Paste some valid characters from the Clipboard into the current position. The element name is to be modified for both - the opening and the closing tag: 
```
<html>
	<bodyF>
		<h1 class="test">
		</h1>
	</bodyF>
</html>
```
- Press `Ctrl-Z` to undo the last modification. The result looks like the following: 
```
<html>
	<body
		<h1 class="test">
		</h1>
	</body>
</html>
```
(extra characters removed from the text: `F>` gets removed from the opening tag instead of removing just `F` character, the document structure gets broken as the result). 
- Pasting more than one character makes the `error` on `Undo` to be bigger (more extra characters resulted to be removed)

mickaelistria:
I can reproduce without pasting: from `<a></a>`, just type another char after the first `a` and linked edit turns it into eg `<ab></ab>`, Ctrl+Z; expected `<a></a>`, got `<a</a>`.
Do you think you can reproduce it in some automated test?

mickaelistria:
I fhink I found the cause: the undo text edit, when hitting Ctrl+Z is processed but the linked edit is still active. So the linkedEdit adds an extra event to remove some more chars. On undo, the linkedEdit should probably be ignored.

mickaelistria:
An interesting case is to reproduce is 
```
<a>
  <b></b>
</a>
```
if editing `a` and undoing immediately, file is corrupted (there are 3 TextEdits while undoing instead of 2)
When editing `a` and moving to `b` so linkedEdit ranges are not involved any more, then undoing is fine.

vrubezhny:
> I fhink I found the cause: the undo text edit, when hitting Ctrl+Z is processed but the linked edit is still active. So the linkedEdit adds an extra event to remove some more chars. On undo, the linkedEdit should probably be ignored.

[...]

PR #184, opened 2022-07-12
Undo operation fails to correctly restore text #178

vrubezhny:


vrubezhny:
@mickaelistria So, finally the test works, but unfortunately it succeeds, so it doesn't catch the issue

mickaelistria:
> so it doesn't catch the issue:

Can you reproduce the issue with the Mock Language Server manually? With Run As > Eclipse Application on the test fragment, and running the steps automated in this test?

vrubezhny:
> Can you reproduce the issue with the Mock Language Server manually? With Run As > Eclipse Application on the test fragment, and running the steps automated in this test?

Nope. I can't. maybe it's because the Linked Editing regions returned by Mock LS are static (not re-calculated on characters typing), but as of now it works correctly with JUnit test we made as well as when I try to test it manually with Mock LS.

mickaelistria:
Merged as https://github.com/eclipse/lsp4e/commit/336301738831ea681027fa5653e4d4a0889d0da2

PR #191, opened 2022-07-26
Do not crash if the initializeFuture is completed exceptionally

rubenporras:
in this case, the lspStreamProvider is set to null in stop() and
subsequent calls during the initialization will throw an NPE.


mickaelistria:
I'm not found of this many `if ...` being added, it kinds of violates the flow logic. If we want some parts of the future to only work in another future has completed properly, we should instead use a `thenCombineAsync(...)` which would wait for future successful a completion to allow further progress.

PR #192, opened 2022-07-26
Fix NPE when the server is stopped due to a failure during initialization

rubenporras:


rubenporras:
@mickaelistria , thanks, you suggestion is better. I did a mistake with the PR, so we lost the previous PR, I created a new one.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServerWrapper.java
```diff
@@ -243,7 +243,8 @@ public synchronized void start() throws IOException {
 					stop();
 				}
 				return null;
-			}).thenApply(unused -> {
+			});
+			initializeFuture.thenCombineAsync(initializeFuture, (unused1, unused2) -> {
```
mickaelistria, line 247: I don't think that `future.thenCombineAsync(future, ...)` is a good nor expected pattern.
I've a bit lost track about what you're willing to achieve. Maybe combining is not the right idea after all.
Maybe what you'd like to add here is just a line `initializeFuture.exceptionallyAsync(this::stop)` here before adding the other futures (which shouldn't be invoked anyway when previous future has exception)

rubenporras:
I think I got it right now :) Would you agree?

mickaelistria:
> I think I got it right now :) Would you agree?

Yes, thanks!

PR #193, opened 2022-07-26
feat: add support for `window/showDocument`

bastiandoetsch:
Hi all,

This PR adds support to navigate to a document when receiving the `window/showDocument` message.

Best,
Bastian

### DIFF
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServerWrapper.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServerWrapper.java
@@ -69,6 +69,7 @@ import org.eclipse.lsp4e.server.StreamConnectionProvider;
 import org.eclipse.lsp4e.ui.Messages;
 import org.eclipse.lsp4e.ui.UI;
 import org.eclipse.lsp4j.ClientCapabilities;
+import org.eclipse.lsp4j.ClientInfo;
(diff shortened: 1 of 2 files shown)


rubenporras (approved):
The change looks good to me. Shall we merge?

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServerWrapper.java
```diff
@@ -374,6 +383,20 @@ public synchronized void start() throws IOException {
...
+	private ClientInfo getClientInfo(String name) {
+		String pluginVersion = Platform.getBundle(LanguageServerPlugin.PLUGIN_ID).getVersion().toString();
+		ClientInfo clientInfo = new ClientInfo(name, pluginVersion);
+		return clientInfo;
+	}
+
+	private WindowClientCapabilities getWindowClientCapabilities() {
+		WindowClientCapabilities windowClientCapabilities = new WindowClientCapabilities();
+		windowClientCapabilities.setShowDocument(new ShowDocumentCapabilities(true));
+		windowClientCapabilities.setWorkDoneProgress(true);
```
rubenporras, line 395: good catch, this was indeed missing

rubenporras:
@mickaelistria , I think we should merge this before the next release, so if you plan to do it soon, let us merge this before. I think it is a good state.

PR #195, opened 2022-07-28
Fix suspected concurrency bug in debug logging

ahmedneilhussain:
We get random `ConcurrentModificationExceptions` when running with json wire logging (via setting `org.eclipse.lsp4e/debug` to true), e.g.:

image

We thought these might be the smoking gun for a bug we were investigating, but I think they're an artifact of the toString() method traversing collections while the ultimate endpoint message consumer may be mutating (e.g. by sorting) the message.

These turn up a fair bit when running our integration tests with this tracing turned on, and go away when these two lines are swapped (`logMessage()` is fully synchronous, while `consumer.consume()` might well not be).


sebthom:
LGTM, also from a troubleshooting point of view it makes sense to me to first log the message before it is consumed.

PR #196, opened 2022-07-29
Allow to navigate from a URI to a range within a document

bastiandoetsch:
This PR adds the ability to parse the fragment of a URI to determine the appropriate range in a document. This allows to navigate within a document from e.g. a hover that contains URIs.

The following formats are understood:
```
file://path/to/file#L34,1-L35,3
file://path/to/file#34,1-35,3
file://path/to/file#L34,1-35,3
file://path/to/file#L34
file://path/to/file#L34,1
```

### DIFF
--- a/org.eclipse.lsp4e.test/src/org/eclipse/lsp4e/test/edit/LSPEclipseUtilsTest.java
+++ b/org.eclipse.lsp4e.test/src/org/eclipse/lsp4e/test/edit/LSPEclipseUtilsTest.java
@@ -446,4 +446,51 @@ public class LSPEclipseUtilsTest {
 		IDE.openEditorOnFileStore(UI.getActivePage(), EFS.getStore(file.toURI()));
 		Assert.assertNotEquals(Collections.emptySet(), LSPEclipseUtils.findOpenEditorsFor(file.toURI()));
 	}
+
+	@Test
+	public void parseRange_shouldReturnRange_UriWithStartLineNo() {
+		Range actual = LSPEclipseUtils.parseRange("file:///a/b#L35");
+		assertEquals(34, actual.getStart().getLine());
+	}
+
+	@Test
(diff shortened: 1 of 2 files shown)


mickaelistria:
Where is `parseRange` called in the actual code?
I think it would be better to reuse FileLocationDetails and extend it to support this new form of URLs; so it would improve Platform as a whole and also improve LSP4E by reusing upstream stuff.

bastiandoetsch:
`ParseRange` is used in `LSPEclipseUtil.open()`

bastiandoetsch:
> Where is `parseRange` called in the actual code? I think it would be better to reuse FileLocationDetails and extend it to support this new form of URLs; so it would improve Platform as a whole and also improve LSP4E by reusing upstream stuff.

Which project would that actually be? I must admit, I'm not that familiar with the Eclipse projects we don't need directly for our Eclipse plugin 😁. Do you think, it would be integrated there quickly? 

Just thinking about change velocity, it may well be a good solution to do a two-pronged approach - one change in LSP4e that would be used until the other project (PDE?) would implement and release the other. Also, if it's done only in platform, wouldn't it force users to use the newest Eclipse version? Just thinking of our users, and some of them have longer and very restricted update cycles due to security within their companies.

bastiandoetsch:
Hey @mickaelistria , bumping this for resolution - how would you like to proceed with this? I'm fine with any solution, preferring the integration into lsp4e though. If upstream (which project?) contribution is not too complicated/bureaucratic, I'd be fine to add the same code there, too. Else, we'll see if we want to maintain this code ourselves in our plugin, which is of course the least preferred solution for me.

vrubezhny:
@bastiandoetsch Please fix compilation errors in JUnit Test and rebase your PR

bastiandoetsch:
@vrubezhny rebased and fixed the tests. Looks like it passes now.

mickaelistria:
Thanks!

PR #206, opened 2022-08-12
Update LSPRenameHandler to not NPE

GeraldMit:
It is possible to have an ITextEditor implementation that can return an ISelectionProvider from getSelectionProvider() that is null as the ITextEditor API specification doesn't forbid returning null  (though the preferred base of AbstractTextEditor doesn't happen to normally do so and if implementing AbstractTextEditor it makes a non-null assumption...)


GeraldMit:
For Issue #207

sebthom (approved):
LGTM Thanks!

Issue #147, opened 2022-06-03
Errors when creating untitled text files


PyvesB:
Hello 👋🏻 

When creating an untitled text file in Eclipse, the following error is now thrown by LSP4E, and the resulting file can no longer be saved:
```
org.eclipse.core.runtime.CoreException: File not found: <workspace>/.metadata/.plugins/org.eclipse.ui.editors/_1727503488.
	at org.eclipse.core.internal.filesystem.Policy.error(Policy.java:48)
	at org.eclipse.core.internal.filesystem.local.LocalFile.openInputStream(LocalFile.java:423)
	at org.eclipse.core.internal.filebuffers.FileStoreTextFileBuffer.getContentType(FileStoreTextFileBuffer.java:293)
	at org.eclipse.lsp4e.LSPEclipseUtils.getDocumentContentTypes(LSPEclipseUtils.java:934)
	at org.eclipse.lsp4e.LanguageServiceAccessor.getLSWrappers(LanguageServiceAccessor.java:348)
	at org.eclipse.lsp4e.LanguageServiceAccessor.getLanguageServers(LanguageServiceAccessor.java:601)
	at org.eclipse.lsp4e.operations.rename.LSPRenameHandler.setEnabled(LSPRenameHandler.java:113)
	at org.eclipse.ui.internal.handlers.HandlerProxy.setEnabled(HandlerProxy.java:229)
	at org.eclipse.ui.internal.handlers.E4HandlerProxy.canExecute(E4HandlerProxy.java:76)
	at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:104)
	at java.base/java.lang.reflect.Method.invoke(Method.java:577)
	at org.eclipse.e4.core.internal.di.MethodRequestor.execute(MethodRequestor.java:58)
	at org.eclipse.e4.core.internal.di.InjectorImpl.invokeUsingClass(InjectorImpl.java:317)
	at org.eclipse.e4.core.internal.di.InjectorImpl.invoke(InjectorImpl.java:251)
	at org.eclipse.e4.core.contexts.ContextInjectionFactory.invoke(ContextInjectionFactory.java:173)
	at org.eclipse.e4.core.commands.internal.HandlerServiceHandler.isEnabled(HandlerServiceHandler.java:62)
	at org.eclipse.core.commands.Command.isEnabled(Command.java:832)
[...]

PR #208, opened 2022-08-13
getDocumentContentTypes to ignore FNFE

GeraldMit:
for Errors when creating untitled text files #147

the content type may be based on path or file name pattern or another subsystem via the ContentTypeManager so that is not necessarily an error condition

### DIFF
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LSPEclipseUtils.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LSPEclipseUtils.java
@@ -935,7 +935,12 @@ public class LSPEclipseUtils {
 						contentTypes.add(contentType);
 					}
 				} catch (CoreException e) {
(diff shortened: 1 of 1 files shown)


org.eclipse.lsp4e/src/org/eclipse/lsp4e/LSPEclipseUtils.java
```diff
@@ -935,7 +935,12 @@ public static List<IContentType> getDocumentContentTypes(@NonNull IDocument docu
 						contentTypes.add(contentType);
 					}
 				} catch (CoreException e) {
-					LanguageServerPlugin.logError("Exception occurred while fetching the content type from the buffer", e); //$NON-NLS-1$;
+					if (!e.getCause() instanceof java.io.FileNotFoundException) {
```
GeraldMit: Needs to be 
`if (!(e.getCause() instanceof java.io.FileNotFoundException)) {`

GeraldMit (commented):
order of operations fix

mickaelistria:
I should have squashed and merged instead.
Thanks anyway!

PR #209, opened 2022-08-18
Fix lost error message checkInitialConditions

rubenporras:
The method checkInitialConditions was handling the error using an status
which was never returned. In addition the method handleError was
returning a value which was never used.

Both problem has been fixed by removing handleError and adding a method
that just gets the error message. This makes more clear where this error
message is attached. The error handling of checkInitialConditions has
been then fixed accordingly.

### DIFF
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/rename/LSPRenameProcessor.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/rename/LSPRenameProcessor.java
@@ -118,8 +118,7 @@ public class LSPRenameProcessor extends RefactoringProcessor {
 				}
 			}
 		} catch (Exception e) {
-			handleError(e, status);
-			return new RefactoringStatus();
+			status.addFatalError(getErrorMessage(e));
 		}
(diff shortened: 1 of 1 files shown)


org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/rename/LSPRenameProcessor.java
```diff
@@ -118,8 +118,8 @@ public RefactoringStatus checkInitialConditions(IProgressMonitor pm)
 				}
 			}
 		} catch (Exception e) {
-			handleError(e, status);
-			return new RefactoringStatus();
+			status = new RefactoringStatus();
```
mickaelistria, line 121: Why instantiating a new status here?

rubenporras: I also think it is weird, but the previous version of the code was doing it, since I do not understand why it was so, I decided to keep it that way. Since you also do not understand it, I submit a new version of the PR without the instantiation.

Issue #210, opened 2022-08-18
LSPCodeActionMarkerResolution.hasResolutions(IMarker) does not check current calculated values


rubenporras:
As a result, when one has a list of markers visible in the markers view, remote calls are constantly done to the language server, the previously calculated values are ignored. Such a stack trace looks like

Thread [main] (Suspended (breakpoint at line 120 in LSPCodeActionMarkerResolution))
owns: RunnableLock (id=10332)
LSPCodeActionMarkerResolution.checkMarkerResoultion(IMarker) line: 120
LSPCodeActionMarkerResolution.hasResolutions(IMarker) line: 203
MarkerHelpRegistry.hasResolution(IMarker, IConfigurationElement, IMarkerResolutionGenerator) line: 264
MarkerHelpRegistry.hasResolutions(IMarker) line: 234
MarkerProblemSeverityAndMessageField(MarkerField).annotateImage(MarkerItem, Image) line: 87
MarkerProblemSeverityAndMessageField.update(ViewerCell) line: 80
MarkerColumnLabelProvider.update(ViewerCell) line: 57
TreeViewerColumn(ViewerColumn).refresh(ViewerCell) line: 144
MarkersTreeViewer(AbstractTreeViewer).doUpdateItem(Item, Object) line: 970
MarkersTreeViewer.doUpdateItem(Item, Object) line: 70
AbstractTreeViewer$UpdateItemSafeRunnable.run() line: 126
SafeRunner.run(ISafeRunnable) line: 45
1263455953.run(ISafeRunnable) line: not available
SafeRunnable.run(ISafeRunnable) line: 174
MarkersTreeViewer(AbstractTreeViewer).doUpdateItem(Widget, Object, boolean) line: 1048
StructuredViewer$UpdateItemSafeRunnable.run() line: 427
SafeRunner.run(ISafeRunnable) line: 45
1263455953.run(ISafeRunnable) line: not available
SafeRunnable.run(ISafeRunnable) line: 174
MarkersTreeViewer(StructuredViewer).updateItem(Widget, Object) line: 2109
MarkersTreeViewer(AbstractTreeViewer).updateChildren(Widget, Object, Object[], boolean) line: 2791
MarkersTreeViewer(AbstractTreeViewer).internalRefreshStruct(Widget, Object, boolean) line: 1971
MarkersTreeViewer(TreeViewer).internalRefreshStruct(Widget, Object, boolean) line: 684
[...]

PR #211, opened 2022-08-18
Check current calculated remediation in hasResolutions

rubenporras:
and only check for resolutions if there is no remediation.


rubenporras:
For https://github.com/eclipse/lsp4e/issues/210

Issue #216, opened 2022-08-23
Missing user feedback when on save actions timeout and timeout is too small


rubenporras:
We currently have no user feedback when on save actions timeout and the timeout is too small (just 2secs).

The proposal is to:
* increase the timeout to 10secs 
* add a pop-up dialog if the timeout is reached explaining what happen and pointing the user to the preference page where he can
* change the timeout or suppress the pop-up

mickaelistria:
Do save action freeze the editor? If so a 10s timeout would be too extreme.
A popup is extremely annoying.
A notification (vanishing popup on the bottom right corner) would be a good cimpromise.

rubenporras:
I also think pop-ups are very annoying but I did not know we could do vanishing notifications. That is much better. 
I have set the default timeout to five secs

PR #218, opened 2022-08-23
Improve user feedback when on save actions timeout

rubenporras:
by showing a notification (vanishing pop-up on the bottom right corner) and
increasing the default timeout from two to five seconds

### DIFF
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/DocumentContentSynchronizer.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/DocumentContentSynchronizer.java
@@ -39,10 +39,13 @@ import org.eclipse.jface.text.BadLocationException;
 import org.eclipse.jface.text.DocumentEvent;
 import org.eclipse.jface.text.IDocument;
 import org.eclipse.jface.text.IDocumentListener;
+import org.eclipse.lsp4e.ui.Messages;
 import org.eclipse.lsp4j.DidChangeTextDocumentParams;
 import org.eclipse.lsp4j.DidCloseTextDocumentParams;
 import org.eclipse.lsp4j.DidOpenTextDocumentParams;
 import org.eclipse.lsp4j.DidSaveTextDocumentParams;
+import org.eclipse.lsp4j.MessageParams;
+import org.eclipse.lsp4j.MessageType;
 import org.eclipse.lsp4j.Range;
 import org.eclipse.lsp4j.ServerCapabilities;
 import org.eclipse.lsp4j.TextDocumentContentChangeEvent;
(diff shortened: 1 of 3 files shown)


rubenporras:
for https://github.com/eclipse/lsp4e/issues/216

org.eclipse.lsp4e/src/org/eclipse/lsp4e/DocumentContentSynchronizer.java
```diff
@@ -238,29 +240,29 @@ public void documentAboutToBeSaved() {
...
 			Integer timeoutCount = WILL_SAVE_WAIT_UNTIL_TIMEOUT_MAP.compute(uri,
 					(k, v) -> v == null ? 1 : Integer.valueOf(v + 1));
-			String message = "WillSaveWaitUntil timeouted out after " + Integer.valueOf(WILL_SAVE_WAIT_UNTIL_TIMEOUT_IN_SECONDS).toString() +" seconds for " + uri;  //$NON-NLS-1$//$NON-NLS-2$
+			String message = "On-Save action timeouted out after " + Integer.valueOf(WILL_SAVE_WAIT_UNTIL_TIMEOUT_IN_SECONDS).toString() +" seconds for " + uri;  //$NON-NLS-1$//$NON-NLS-2$
 			if (timeoutCount > WILL_SAVE_WAIT_UNTIL_COUNT_THRESHOLD) {
-				message = message + ", it will no longer be called for this document"; //$NON-NLS-1$
+				message = message + ". On-Save actions will no longer be called for this document"; //$NON-NLS-1$
 			}
-			LanguageServerPlugin.logWarning(message, e);
+			ServerMessageHandler.showMessage("On Save Action Timeout", new MessageParams(MessageType.Error, message)); //$NON-NLS-1$
```
mickaelistria, line 261: Shouldn't the title feature a warning sign ⚠ to get some attention?
Also, as those becomes UI elements and not only logging message, those should be internationalized in the Messages class.

rubenporras: Since it is an error message, it already shows an error sign. I will move the message to the Messages class

rubenporras: done

rubenporras:
@mickaelistria 
there is a problem with the notification, it is cut on the right margin, instead of being wrap:

image

The problem should be somewhere in `org.eclipse.lsp4e.ServerMessageHandler.LSPNotification` but I cannot figure it out, I would expect that using `SWT.WRAP` in `org.eclipse.lsp4e.ServerMessageHandler.LSPNotification.createContentArea(Composite)` would be enough, but obviously it is not.

Do you have an idea what the problem with the wrapping is?

mickaelistria:
I'm going to merge this. Please track the rendering problem in another deficated report

Issue #220, opened 2022-08-23
LSPNotification cuts message on the right margin, instead of wrapping it


rubenporras:
See https://github.com/eclipse/lsp4e/pull/218#issuecomment-1223965590

@akurtakov, any ideas what is the problem with wrapping? I think you initially introduced the SWT.WRAP style.

akurtakov:
Sorry but nothing comes to my mind. Can you test whether behavior is same on at least two windowing system e.g. Win32+ GTK or Win32+Cocoa or Cocoa+GTK ? This is to rule out possible WS specific bug in SWT?

bastiandoetsch:
I see it also on macOS :)

PR #221, opened 2022-08-23
Make sure LSP notification message is wrapped

akurtakov:
Caused by long standing SWT issue (
https://bugs.eclipse.org/bugs/show_bug.cgi?id=194758 ) but easily
workaroundable as done in this patch.
Some minor SonarLint warnings are also fixed.
Fixes https://github.com/eclipse/lsp4e/issues/220 .


akurtakov:
@rubenporras Please verify this fixes the issue for you, it was good enough in my tests.

rubenporras:
@akurtakov , yes, it fixes it for me as well. Thanks. I merge it since I think the test is a wrong positive.

PR #222, opened 2022-08-24
Add property to allow custom value of the timeout for willSaveWaitUntil

rubenporras:


rubenporras:
for https://github.com/eclipse/lsp4e/issues/216

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServerWrapper.java
```diff
@@ -711,7 +711,15 @@ public LanguageServer getServer() {
...
 	}
 
-	/**
+  /**
+   * Returns the language server ID.
+   */
+  @NonNull
+  public String getServerId() {
+      return this.serverDefinition.id;
+  }
```
mickaelistria, line 720: Could we avoid adding API method here? you can directly call `wrapper.serverDefinition.id` from DocumentContentSynchronizer.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/DocumentContentSynchronizer.java
```diff
@@ -225,10 +232,27 @@ private boolean serverSupportsWillSaveWaitUntil() {
...
 	private static final int WILL_SAVE_WAIT_UNTIL_COUNT_THRESHOLD = 3;
 	private static final Map<String, Integer> WILL_SAVE_WAIT_UNTIL_TIMEOUT_MAP = new ConcurrentHashMap<>();
 
+	/**
+	 * Converts a language server ID to the preference ID to define a timeout
+	 * for willSaveWaitUntil
+	 *
+	 * @return language server's preference ID to define a timeout for willSaveWaitUntil
+	 */
+	public static @NonNull String lsToWillSaveWaitUntilTimeoutKey(String serverId) {
```
mickaelistria, line 246: Instead of a serverId as parameter, the method should directly take the languageServerWrapper as argument here. Stronger typing helps.
Also, it shouldn't have to be public, should it?

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServerWrapper.java
rubenporras: thanks, done.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/DocumentContentSynchronizer.java
rubenporras: removed

mickaelistria: I'm still seeing it ;)

rubenporras: now :)

PR #223, opened 2022-08-25
Implement preference page for configurable timeouts

rubenporras:


mickaelistria:
I'm not so enthusiast about providing those details to end-uesrs. It's quite complex and very technical internal stuff. Most users wouldn't be comfortable with it. Do you already have a real use-case of some end-users wanting configurable timeouts?

rubenporras:
@mickaelistria , how would you then suggest to set the preference, on the ini file? Is this what you have in mind. That would be doable but just for advanced users. We could do with that for the moment.

mickaelistria:
> on the ini file?

Yes, that's also what I had in mind. This preference would basically be set by the person who integrates the LS, and this is usually the plugin developer, or the RCP application provider, that can set them via plugin_customization.ini and other APIs.

rubenporras:
Okay, lets go with that.

PR #229, opened 2022-08-30
UNC paths: Support preferred UNC URI syntax on Windows

ahmedneilhussain:
This change allows LSP4e to cope with a language server that sends URIs for UNC paths on an unmapped network drive in the modern [Microsoft preferred] way, using the authority to convey the server name.

I.e. `\\myserver\myshare\path\to\file` can be accepted in the newer, preferred syntax `file://myserver/myshare/path/to/file` rather than `file:////myserver/myshare/path/to/file`. The traditional `java.io.File` class will only generate and accept URIs in the legacy format, but `java.nio.file.Paths` will accept either and generate the former.

Changing LSP4e to generate the newer syntax would risk breaking any language servers built in Java, and furthermore caused some internal breakages where we convert `File`s to `URI`s and back again. There are, however, only a couple of places that consume URIs from the server, which are trivial to fix. LSP4e doesn't use `URI`s directly to key any lookup tables or dictionaries (it mostly uses eclipse `IFile/IResource/IPath` objects in internal APIs) so there don't seem to be any canonicalisation problems.

See https://docs.microsoft.com/en-us/archive/blogs/ie/file-uris-in-windows
https://en.wikipedia.org/wiki/File_URI_scheme#Windows
https://wiki.eclipse.org/Eclipse/UNC_Paths#Programming_with_UNC_paths
https://bugs.openjdk.org/browse/JDK-4723726

And the notes on https://docs.oracle.com/javase/7/docs/api/java/io/File.html and https://docs.oracle.com/javase/7/docs/api/java/nio/file/Path.html#toUri()

### DIFF
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LSPEclipseUtils.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LSPEclipseUtils.java
@@ -1098,12 +1098,8 @@ public class LSPEclipseUtils {
 	 * @param uri A file URI, possibly for a UNC path in the newer syntax with the server encoded in the authority
 	 * @return A file
 	 */
-	public static File fromUri(URI uri) {
+	private static File fromUri(URI uri) {
 		return Paths.get(uri).toFile();
-	}
-
-	public static File fromUri(String uri) {
-		return fromUri(URI.create(uri));
 	}
 
 	public static boolean hasCapability(Either<Boolean, ? extends Object> eitherCapability) {
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LSPEclipseUtils.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LSPEclipseUtils.java
@@ -28,6 +28,7 @@ import java.lang.reflect.Method;
 import java.net.URI;
 import java.net.URISyntaxException;
 import java.net.URL;
+import java.nio.file.Paths;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
@@ -491,7 +492,7 @@ public class LSPEclipseUtils {
 		if (resource != null) {
 			return getDocument(resource);
 		}
-		if (!new File(uri).isFile()) {
+		if (!fromUri(uri).isFile()) {
 			return null;
 		}
 
@@ -731,7 +732,7 @@ public class LSPEclipseUtils {
 					if (resourceOperation instanceof CreateFile) {
 						CreateFile createOperation = (CreateFile) resourceOperation;
 						URI targetURI = URI.create(createOperation.getUri());
-						File targetFile = new File(targetURI);
+						File targetFile = fromUri(targetURI);
 						if (targetFile.exists() && createOperation.getOptions() != null) {
 							if (!createOperation.getOptions().getIgnoreIfExists()) {
 								if (createOperation.getOptions().getOverwrite()) {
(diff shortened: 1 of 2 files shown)


org.eclipse.lsp4e/src/org/eclipse/lsp4e/LSPEclipseUtils.java
```diff
@@ -1081,6 +1082,30 @@ public static URI toUri(String uri) {
...
+	 * complaint. Going via nio.Path allows us to accept either syntax. LSP4e does not use URIs directly e.g. as
+	 * keys in lookup dictionaries so we don't have to worry about canonicalisation problems
+	 *
+	 * See https://bugs.openjdk.org/browse/JDK-4723726
+	 * https://docs.microsoft.com/en-us/archive/blogs/ie/file-uris-in-windows
+	 *
+	 * @param uri A file URI, possibly for a UNC path in the newer syntax with the server encoded in the authority
+	 * @return A file
+	 */
+	public static File fromUri(URI uri) {
```
mickaelistria, line 1101: This should be private if it's not used in other classes.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LSPEclipseUtils.java
```diff
@@ -1081,6 +1082,30 @@ public static URI toUri(String uri) {
...
+		return Paths.get(uri).toFile();
+	}
+
+	public static File fromUri(String uri) {
```
mickaelistria, line 1105: This should be private if it's not used in other classes.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LSPEclipseUtils.java
ahmedneilhussain: Fair point - it's a hang over from when I experimentally tried coaxing LSP4e to work exclusively in terms of the newer syntax (just to see how difficult it would be, ignoring the fact it would probably break some java-based lang servers). We already had a few conversion methods in this utils class so I wanted it to be a public One True Way to convert URIs to Files. If we only want to be tolerant of the newer syntax, then you're right there is no consumer outside this class itself.

ahmedneilhussain: What do you think about testing? Unfortunately the java APIs themselves are platform dependent so you can't just even write a pure unit test that will work cross platform. I can use JUnit Assume.that() but it won't be much use if there is no build infrastructure that ever runs on Windows.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LSPEclipseUtils.java
ahmedneilhussain: I'll delete - it's not actually used unless we make further changes to both produce and consume newer URIs. Apologies.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LSPEclipseUtils.java
mickaelistria: It's up to you. If you prefer adding a test, that would be welcome. And it's very likely that someday someone contributes an automated test execution on windows (it's not that hard), so those tests may be helpful in the future.
But if you prefer, we can merge without those tests; it will help already also things would be more likely to see regressions in the future.

mickaelistria:
Thanks!

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LSPEclipseUtils.java
ahmedneilhussain: I added a unit test but that meant making fromURI public again as package-private isn't enough for test accessibility due to the code structure...

I looked into an integration test but that's going to be very difficult given the way we're set up - can't realistically set up a network share mid test, can't use the local admin share if the workspace isn't mounted in that way in the first place, and can't use mocking very easily because of the prevalence of static methods in the Eclipse APIs used :-(

PR #231, opened 2022-09-01
textDocument/documentLink should be renderered with hyperlink style.

### LINKED ISSUE #126: textDocument/documentLink should be renderered with hyperlink style.
textDocument/documentLink should be renderered with hyperlink style.

Given this HTML file;

```html
<h2>Absolute URLs</h2>
<p><a href="https://www.w3.org/">W3C</a></p>
<p><a href="https://www.google.com/">Google</a></p>

<h2>Relative URLs</h2>
<p><a href="html_images.asp">HTML Images</a></p>
<p><a href="/css/default.asp">CSS Tutorial</a></p>
```

The HTML LS returns documentLink for a/@ref. Those document link should be underlined like this:

image

angelozerr:
textDocument/documentLink should be renderered with hyperlink style.

Fixes #126

### DIFF
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/documentLink/LSPDocumentLinkPresentationReconcilingStrategy.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/documentLink/LSPDocumentLinkPresentationReconcilingStrategy.java
@@ -1,0 +1,169 @@
+/*******************************************************************************
+ * Copyright (c) 2022 Red Hat Inc. and others.
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *******************************************************************************/
+package org.eclipse.lsp4e.operations.documentLink;
+
+import java.net.URI;
+import java.util.List;
+import java.util.concurrent.CompletableFuture;
+
+import org.eclipse.core.runtime.IProgressMonitor;
+import org.eclipse.jdt.annotation.Nullable;
+import org.eclipse.jface.text.BadLocationException;
+import org.eclipse.jface.text.IDocument;
+import org.eclipse.jface.text.IRegion;
+import org.eclipse.jface.text.ITextViewer;
+import org.eclipse.jface.text.Region;
+import org.eclipse.jface.text.TextPresentation;
+import org.eclipse.jface.text.reconciler.DirtyRegion;
+import org.eclipse.jface.text.reconciler.IReconcilingStrategy;
+import org.eclipse.jface.text.reconciler.IReconcilingStrategyExtension;
+import org.eclipse.lsp4e.LSPEclipseUtils;
+import org.eclipse.lsp4e.LanguageServerPlugin;
+import org.eclipse.lsp4e.LanguageServiceAccessor;
+import org.eclipse.lsp4j.DocumentLink;
+import org.eclipse.lsp4j.DocumentLinkParams;
+import org.eclipse.lsp4j.TextDocumentIdentifier;
+import org.eclipse.swt.custom.StyleRange;
+import org.eclipse.swt.widgets.Control;
+
+/**
+ * Reconciling strategy used to display links coming from LSP
+ * 'textDocument/documentLink' with underline style.
+ *
+ * @author Angelo ZERR
+ *
+ */
+public class LSPDocumentLinkPresentationReconcilingStrategy
+		implements IReconcilingStrategy, IReconcilingStrategyExtension {
+
+	/** The target viewer. */
+	private ITextViewer viewer;
+
+	private CompletableFuture<Void> request;
+
+	private IDocument document;
+
+	public void install(@Nullable ITextViewer viewer) {
+		this.viewer = viewer;
+	}
+
+	public void uninstall() {
+		this.viewer = null;
+		cancel();
+	}
+
+	private void underline() {
+		if (viewer == null)
+			return;
+
+		final IDocument document = viewer.getDocument();
+		if (document == null) {
+			return;
+		}
+
+		URI uri = LSPEclipseUtils.toUri(document);
+		if (uri == null) {
+			return;
+		}
+		cancel();
+		final DocumentLinkParams params = new DocumentLinkParams(new TextDocumentIdentifier(uri.toString()));
+		request = LanguageServiceAccessor
+				.getLanguageServers(document, capabilities -> capabilities.getDocumentLinkProvider() != null)
+				.thenAcceptAsync(languageServers -> CompletableFuture.allOf(languageServers.stream()
+						.map(languageServer -> languageServer.getTextDocumentService().documentLink(params))
+						.map(request -> request.thenAcceptAsync(links -> {
+							final Control control = viewer.getTextWidget();
+							if (control != null) {
+								control.getDisplay().asyncExec(() -> underline(links));
+							}
+						})).toArray(CompletableFuture[]::new)));
+	}
+
+	private void underline(List<DocumentLink> links) {
+		if (document == null) {
+			return;
+		}
+		for (DocumentLink link : links) {
+			try {
+				// Compute link region
+				int start = LSPEclipseUtils.toOffset(link.getRange().getStart(), document);
+				int end = LSPEclipseUtils.toOffset(link.getRange().getEnd(), document);
+				int length = end - start;
+				IRegion linkRegion = new Region(start, length);
+
+				// Update existing style range with underline or create a new style range with
+				// underline
+				StyleRange styleRange = null;
+				StyleRange[] styleRanges = viewer.getTextWidget().getStyleRanges(start, length);
+				if (styleRanges != null && styleRanges.length > 0) {
+					// It exists some styles for the range of document link, update just the
+					// underline style.
+					for (StyleRange s : styleRanges) {
+						s.underline = true;
+					}
+					TextPresentation presentation = new TextPresentation(linkRegion, 100);
+					presentation.replaceStyleRanges(styleRanges);
+					viewer.changeTextPresentation(presentation, false);
+
+				} else {
+					// No styles for the range of document link, create a style range with underline
+					styleRange = new StyleRange();
+					styleRange.underline = true;
+					styleRange.start = start;
+					styleRange.length = length;
+
+					TextPresentation presentation = new TextPresentation(linkRegion, 100);
+					presentation.replaceStyleRange(styleRange);
+					viewer.changeTextPresentation(presentation, false);
+				}
+
+			} catch (BadLocationException e) {
+				LanguageServerPlugin.logError(e);
+			}
+		}
+	}
+
+	@Override
+	public void initialReconcile() {
+		underline();
+	}
+
+	/**
+	 * Cancel the last call of 'documenLink'.
+	 */
+	private void cancel() {
+		if (request != null) {
+			request.cancel(true);
+			request = null;
+		}
+	}
+
+	@Override
+	public void setDocument(IDocument document) {
+		this.document = document;
+	}
+
+	@Override
+	public void setProgressMonitor(IProgressMonitor monitor) {
+		// Do nothing
+	}
+
+	@Override
+	public void reconcile(DirtyRegion dirtyRegion, IRegion subRegion) {
+		// Do nothing
+	}
+
+	@Override
+	public void reconcile(IRegion partition) {
+		// Underline document by using textDocument/documentLink with some delay as
+		// reconcile method is called in a Thread background.
+		underline();
+	}
+
+}
--- a/org.eclipse.lsp4e/plugin.xml
+++ b/org.eclipse.lsp4e/plugin.xml
@@ -646,4 +646,21 @@
          </enabledWhen>
       </reconciler>
    </extension>
+   <extension
+         point="org.eclipse.ui.genericeditor.reconcilers">
+      <reconciler
+            class="org.eclipse.lsp4e.operations.documentLink.LSPDocumentLinkPresentationReconciler"
+            contentType="org.eclipse.core.runtime.text">
+         <enabledWhen>
+           <or>
+             <with variable="editorInput">
+               <test property="org.eclipse.lsp4e.hasLanguageServer"/>
+             </with>
+             <with variable="viewer">
+               <test property="org.eclipse.lsp4e.hasLanguageServer"/>
+             </with>
+           </or>
+         </enabledWhen>
+      </reconciler>
+   </extension>   
 </plugin>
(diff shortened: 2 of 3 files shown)


angelozerr:
Please don't see the quality of the ugly code of this draft PR, it is just a POC that I did quickly, but it seems it is working. The main idea is that a IReconciler is registered to consume documentLink and display the DocumentLink as underline style:

image

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/documentLink/LSPPresentationReconcilingStrategy.java
```diff
@@ -0,0 +1,316 @@
+package org.eclipse.lsp4e.operations.documentLink;
```
mickaelistria, line 1: Missing copyright header.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/documentLink/LSPPresentationReconciler.java
```diff
@@ -0,0 +1,42 @@
...
+
+import org.eclipse.jface.text.IDocument;
+import org.eclipse.jface.text.ITextViewer;
+import org.eclipse.jface.text.reconciler.IReconciler;
+import org.eclipse.jface.text.reconciler.MonoReconciler;
+
+/**
+ * {@link IReconciler} implementation to Highlight Symbol (mark occurrences like).
+ */
+public class LSPPresentationReconciler extends MonoReconciler {
```
mickaelistria, line 23: Would you mind renaming to `LSPDocumentLinkPresentationReconciler` ?

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/documentLink/LSPPresentationReconcilingStrategy.java
```diff
@@ -0,0 +1,316 @@
...
+import org.eclipse.lsp4e.LSPEclipseUtils;
+import org.eclipse.lsp4e.LanguageServiceAccessor;
+import org.eclipse.lsp4j.DocumentLink;
+import org.eclipse.lsp4j.DocumentLinkParams;
+import org.eclipse.lsp4j.TextDocumentIdentifier;
+import org.eclipse.lsp4j.services.LanguageServer;
+import org.eclipse.swt.custom.StyleRange;
+import org.eclipse.swt.widgets.Control;
+
+public class LSPPresentationReconcilingStrategy implements IReconcilingStrategy, IReconcilingStrategyExtension {
```
mickaelistria, line 33: Would you mind renaming it to LSPDocumentLinkPresentationReconcilingStrategy ?

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/documentLink/LSPPresentationReconcilingStrategy.java
```diff
@@ -0,0 +1,316 @@
...
+				//
+				// if (localGrammar != null) {
+				// LSPPresentationReconciler.this.grammar = localGrammar;
+				// } else if (isThrowError()) {
+				// throw new TMException("Cannot find TextMate grammar for the given document");
+				// }
+
+				// Update the theme
+				// if (localGrammar != null) {
+				// final String scopeName = localGrammar.getScopeName();
```
mickaelistria, line 109: Can all those comments be removed?

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/documentLink/LSPPresentationReconcilingStrategy.java
```diff
@@ -0,0 +1,316 @@
...
+					.getLanguageServers(document, capabilities -> capabilities.getDocumentLinkProvider() != null)
+					.thenApplyAsync(languageServers -> {
+						List<CompletableFuture<List<DocumentLink>>> futures = new ArrayList<>();
+						for (LanguageServer languageServer : languageServers) {
+							futures.add(languageServer.getTextDocumentService().documentLink(params));
+						}
+						allOf(futures).thenAccept(links -> {
+							final Control control = viewer.getTextWidget();
+							if (control != null) {
+								control.getDisplay().asyncExec(() -> colorize(links));
```
mickaelistria, line 174: I believe all the computation can be run in any thread, and only the `viewer.changeTextPresentation()` should be in the UI Thread.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/documentLink/LSPPresentationReconciler.java
angelozerr: LSPDocumentLinkPresentationReconciler means that it will take care only about document link. My idea is was to manage sematntic higlighting to avoid creating a new Thread reconciler for that. What do you think about that?

mickaelistria: I would rather have different reconcilers for different operations.

rubenporras:
@angelozerr , thanks. It is indeed interesting to me :)

One thing I do not understand is that, if we have different reconcilers (that is the one by TexMate, and then this one which does hyperlinking and semantic highlight), how do we make sure that they do not conflict? Or would we just say that because the LSP Reconciler does a remote call, we assume it will always be slower than the Texmate one? I do not think we can take that assumption. The problem I want to avoid is that the LSP reconciler creates some very elaborate highlight, which is then later undo by a more generic syntax highlight.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/documentLink/LSPPresentationReconcilingStrategy.java
angelozerr: fixed

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/documentLink/LSPPresentationReconcilingStrategy.java
angelozerr: fixed

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/documentLink/LSPPresentationReconciler.java
angelozerr: ok fixed

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/documentLink/LSPPresentationReconcilingStrategy.java
angelozerr: fixed

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/documentLink/LSPPresentationReconcilingStrategy.java
angelozerr: I need to do that otherwise I have an invalid SWT Thread access.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/documentLink/LSPDocumentLinkPresentationReconcilingStrategy.java
```diff
@@ -0,0 +1,229 @@
...
+					StyleRange styleRange = null;
+					StyleRange[] styleRanges = viewer.getTextWidget().getStyleRanges(start, length);
+					if (styleRanges != null) {
+						// It exists some styles for the range of document link, update just the
+						// underline style.
+						for (StyleRange s : styleRanges) {
+							s.underline = true;
+						}
+						TextPresentation presentation = new TextPresentation(linkRegion, 100);
+						presentation.replaceStyleRanges(styleRanges);
```
mickaelistria, line 172: Shouldn't it be `mergeStyleRanges` so the existing other styles (eg colors) should be kept?

angelozerr:
@mickaelistria I think the PR is good to merge even if it doesn't take care of cancel like other operations.

I think we should provide cancel support to improve performance for all operations. As document link is generally quick I think the PR could be merged if you are agree.

Once this PR will be merged I will create an another PR with cancel support. It is not perfect but I think it will start some discussion how to 

 * cancel request (ex : cancel textDocument/documentLink  request) 
 * cancel the process of the result request (ex : render the document  linkas underline).

angelozerr:
> @angelozerr , thanks. It is indeed interesting to me :)

Great!

> One thing I do not understand is that, if we have different reconcilers (that is the one by TexMate, and then this one which does hyperlinking and semantic highlight), how do we make sure that they do not conflict? 

At first you will need to implement an IReconciler (as I did) and not an IPresentationReconclier because Genereic editor can support multiple IPresentationReconclier. But we don't need really an IPresentationReconclier (for textmate too) because we don't use the partition support of Eclipse Platform Text.

> Or would we just say that because the LSP Reconciler does a remote call, we assume it will always be slower than the Texmate one? I do not think we can take that assumption. The problem I want to avoid is that the LSP reconciler creates some very elaborate highlight, which is then later undo by a more generic syntax highlight.

If you try my PR with HTML LS, you will see a/@href are green (coming from textmate) and underlined(coming from my PR):

image

To do that I get the current styles at https://github.com/eclipse/lsp4e/pull/231/files#diff-82ab357a4743f7ecbfe9b3520d818c17ebed3ce81ac9169850a12c2aa251636dR164 andI update it with underline at https://github.com/eclipse/lsp4e/pull/231/files#diff-82ab357a4743f7ecbfe9b3520d818c17ebed3ce81ac9169850a12c2aa251636dR169

I suggest that you try like I did but perhaps it is not the proper solution and it will require to study how Java editor do that with semantic color https://github.com/eclipse-jdt/eclipse.jdt.ui/blob/master/org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/javaeditor/SemanticHighlightingPresenter.java

rubenporras:
@mickaelistria: Just for the record, I have been looking at the code and using it as a basis for my semantic highlighting, and I think the change looks good and would be ready to merge.

mickaelistria:
@rubenporras OK, feel free to merge ;)

angelozerr:
Please note that there is a bug with this PR. If textmate applies styles AFTER this reconciler, it will remove the underlined.

angelozerr:
> @angelozerr , so still the problem with the order of reconcilers exist. 

Yes.

> Is there maybe code to make sure that the IPresentationReconciler runs before the rest of the IReconcilers?

JDT editor have the same problematic, so we should study how it support syntax and semantic syntax coloration. I know that JST extends the default TextPresentation class which manages damage, repairer feature. TMPresentationReconciler should perhaps manage repairer and damage feature? To be honnest I don't know and I hope someone could have timeto study this problematic.

> On my tests, it was always like that for the semantic highlight, but I do not know if it was luck or something that is guaranteed.

It is easy to reporduce the problem with my PR:

 * create an HTML file like this:

```html
<h2>Absolute URLs</h2>
<p><a href="https://www.w3.org/">W3C</a></p>
<p><a href="https://www.google.com/">Google</a></p>

<h2>Relative URLs</h2>
<p><a href="html_images.asp">HTML Images</a></p>
<p><a href="/css/default.asp">CSS Tutorial</a></p>
```

 * at this step you should see the proper coloration (from textmate and from LSP4e documentLink)

image

* Select All text (Ctr+A) and do a Cut (Ctrl+X) to copy the editor content and remove all content.
* Do a Copy (Ctrl+V ), you will see that textmate color is done correctly but not the hyperlink (because textmate process colorization after the documentLink)
 * type a character, all documentLink are displayed as underlined.

Here a demo with this problem:

HTMLUnderlinedProblem

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/documentLink/LSPDocumentLinkPresentationReconcilingStrategy.java
rubenporras: @angelozerr , I think the change would be ready to be merged, but there is still this pending question on why use replace and not `mergeStyleRanges`. If that is clear, I would merge it ;)

angelozerr: TextPresentation is created only here so add, replace, merge will do the same thing I think.

There are several problems with this PR:

 * the request is not canceled properly on language server side.
 * the result of request is not canceled. You can have a lot of error logs with BadLocation (like Outline or other operations) and you will update the UI to many times althouth it should be updated just with last changes.

Those problems can be improved in https://github.com/eclipse/lsp4e/issues/236

It create an extra background Thread with reconciler although is is never used. This issue can be fixed like https://github.com/eclipse/lsp4e/pull/241

angelozerr:
@rubenporras I did a new commit and it works better, indeed now I do the underline on reconcile method (it means that there is some delay) and it works good now with the HTML sample as underlined is delayed:

HTMLDocumentLink

mickaelistria:
This looks good to me and I'm going to merge it.
However, while reading the code in my IDE, I thought that in the future, we should probably rework the whole documentLink story to implement it with some IHyperlinkDetector, which comes with a bunch of well integrated capacities.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/documentLink/LSPDocumentLinkPresentationReconcilingStrategy.java
```diff
@@ -0,0 +1,168 @@
...
+	}
+
+	public void uninstall() {
+		this.viewer = null;
+		cancel();
+	}
+
+	private void underline() {
+		final var viewer = LSPDocumentLinkPresentationReconcilingStrategy.this.viewer;
+		if (viewer == null)
```
rubenporras, line 63: why not just `this.viewer`?

mickaelistria (approved):
Actually, I'll leave it to you @rubenporras to merge it when you're happy with it. The current state is good enough to me.

rubenporras (approved):

rubenporras:
@mickaelistria , thanks, I will merge it then when Jenkins approve it

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/documentLink/LSPDocumentLinkPresentationReconcilingStrategy.java
angelozerr: Good catch, fixed.

rubenporras:
@angelozerr , thanks a lot!

angelozerr:
@rubenporras @mickaelistria thank a lot for your reviews!

PR #227, opened 2022-08-26
Support for textDocument/inlayHint

angelozerr:
Support for textDocument/inlayHint

Fixes #125

Signed-off-by: azerr <azerr@redhat.com>

PR #237, opened 2022-09-06
Use full document range for inlay hints request

ahmedneilhussain:
The new inlay hints implementation is really cool, but our language server respects the viewport range in the request so as things stand doesn't return any data.

The protocol is designed (a la Vscode) to support optimising traffic for large files by only requesting hints for the visible portion of the document, i.e. the viewport. I don't think Eclipse can support this - I stuck some breakpoints/logging in and only got requests for the code mining when the document was loaded or changed, not when I scrolled an unmodified document. If that's correct, then the best and only thing we can do is to send the full document range.

What do people think? @angelozerr @mickaelistria 

Cf https://github.com/eclipse/lsp4e/pull/227#discussion_r956560920 and https://github.com/eclipse/lsp4e/pull/227#issuecomment-1229200677

### DIFF
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/inlayhint/InlayHintProvider.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/inlayhint/InlayHintProvider.java
@@ -37,7 +37,15 @@ public class InlayHintProvider extends AbstractCodeMiningProvider {
 	private CompletableFuture<List<? extends ICodeMining>> provideCodeMinings(@NonNull IDocument document) {
 		URI docURI = LSPEclipseUtils.toUri(document);
 		if (docURI != null) {
-			Range viewPortRange = new Range(new Position(0,0), new Position(0,0));
+			// Eclipse seems to request minings only when the document is loaded (or changed), rather than
+			// whenever the viewport [displayed area] changes, so request minings for the whole document in one go.
+			Position end = new Position(0,0);
+			try {
+				end = LSPEclipseUtils.toPosition(document.getLength(), document);
+			} catch (BadLocationException e) {
+				LanguageServerPlugin.logWarning("Unable to compute end of document", e); //$NON-NLS-1$
+			}
+			Range viewPortRange = new Range(new Position(0,0), end);
 			InlayHintParams param = new InlayHintParams(new TextDocumentIdentifier(docURI.toString()), viewPortRange);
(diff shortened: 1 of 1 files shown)


angelozerr:
> The new inlay hints implementation is really cool, 

Glad it pleases you!

> but our language server respects the viewport range in the request so as things stand doesn't return any data.

Indeed it is a bug, but I think we should compute the visible viewport instead of the full document.

angelozerr:
To compute the visible viewport, I suggest that you do like https://github.com/eclipse-platform/eclipse.platform.text/blob/150da4b2675d18f53a2087b74f6e6bed65ce0472/org.eclipse.jface.text/src/org/eclipse/jface/text/source/inlined/InlinedAnnotationSupport.java#L141

But it means that compute of visible lines will be done twice. It should be nice if platform text provide a mean to get the visible line viewport to avoid computing those information twice. @mickaelistria what do you think about that?

mickaelistria:
I believe this case will require new API in Platform to instruct it to refresh codeMinings on viewport change (the CodeMiningProvider or the extension point would need to declare the restrained scope). I'd rather avoid LSP4E to host workarounds for a problem that seems rather generic.

ahmedneilhussain:
> I believe this case will require new API in Platform to instruct it to refresh codeMinings on viewport change (the CodeMiningProvider or the extension point would need to declare the restrained scope). I'd rather avoid LSP4E to host workarounds for a problem that seems rather generic.

Do we think it's not possible to implement this solely in LSP4e? It would require changes to the underlying `jface` classes?

ahmedneilhussain:
> > Do we think it's not possible to implement this solely in LSP4e?
> 
> That would probably be a bad idea: this problem is not only for LSP4E but affects Platform code mining in general. Workarounds in LSP4E would lead to delay a proper resolution on Platform. I'd rather avoid LSP4E to be a cause of not-fixed-issue on Platform because it hosts too many workarounds.

In that case is this PR as it stands acceptable as an incremental improvement? If we can't meaningfully pass through the viewport then in order to see anything the server has to send inlay hints for the whole document anyway.

angelozerr:
Indeed by waiting an update from platform text this PR should be merged since it sends to thée server a correct range even if it il not optimized

mickaelistria:
Thanks!

angelozerr:
Indeed by waiting for a platform text update I think we should mère this PR
which send a correct range even il is not optimized.

Le ven. 9 sept. 2022 à 18:21, ahmedneilhussain ***@***.***> a
écrit :

PR #241, opened 2022-09-08
Don't create a reconciler Thread for higlight and linked editing range.

angelozerr:
Don't create a reconciler Thread for higlight and linked editing range.

### DIFF
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/NoThreadReconciler.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/NoThreadReconciler.java
@@ -1,0 +1,121 @@
+/*******************************************************************************
+ * Copyright (c) 2022 Red Hat Inc. and others.
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *   Victor Rubezhny (Red Hat Inc.) - initial implementation
+ *******************************************************************************/
+package org.eclipse.lsp4e.operations;
+
+import org.eclipse.jface.text.DocumentEvent;
+import org.eclipse.jface.text.IDocument;
+import org.eclipse.jface.text.IDocumentListener;
+import org.eclipse.jface.text.ITextInputListener;
+import org.eclipse.jface.text.ITextViewer;
+import org.eclipse.jface.text.reconciler.IReconciler;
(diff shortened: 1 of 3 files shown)


mickaelistria:
I think it makes sense in LSP4E because it already uses async operations/futures, so the extra thread is useless. This should probably be documented on `NoThreadReconciler` to explain why a different pattern then the usual AbstractReconciler can be used here.

angelozerr:
> I think it makes sense in LSP4E because it already uses async operations/futures, so the extra thread is useless. 

It is indeed useless because all reconcile methods are implemented with empty code.

mickaelistria:
One thing is that a documentListener will cause the requests to be sent on every change, while if we stick to a reconciler, there is IIRC a delay before running the desired work to not invalidate/execute it too often while user is typing. I think we should honor that an only send requests after the usual delay.

angelozerr:
> One thing is that a documentListener will cause the requests to be sent on every change, while if we stick to a reconciler, there is IIRC a delay before running the desired work to not invalidate/execute it too often while user is typing.

Indeed but for higlight and linked editing range LSP4e implement reconcile mthod with empty, so we don't need that and more we don't need for those operations a document listener. I will remove this usnused document listener.

mickaelistria:
As mentioned, I don't think we should go for a reconciler without thread. We should instead use some MonoReconciler, drop the document listener, and make the LS request as part of the reconcilingStrategy.
The 500ms is useful, and reconciler already has a document listeners, we should just embrace those for a better UX.

angelozerr:
> The 500ms is useful, and reconciler already has a document listeners, we should just embrace those for a better UX.

The 500ms is used only when you implemnt reconcile methods. In our case we don't implement it.

mickaelistria:
> The 500ms is used only when you implemnt reconcile methods. In our case we don't implement it.

Exactly, and I think this is an issue. We should instead remove the document listener and implement those methods.

angelozerr:
> Exactly, and I think this is an issue. We should instead remove the document listener and implement those methods.

Higlight reacts on selection changes and linked editing range don't need to react on some reconcile or document change. So we dont need to react on some document change (on reconcile or document change lsitener).

mickaelistria:
Sorry, I was mixing highlights and hyperlinks. So you're entirely right (again!) that both those operations do not require a thread. However, what value does NoThreadReconcilier bring over plain `IReconciler`?

angelozerr:
@mickaelistria I think we should manage with clean mean this topic of Thread with https://github.com/eclipse-platform/eclipse.platform.text/issues/85

angelozerr:
Closed for https://github.com/eclipse/lsp4e/pull/272

PR #251, opened 2022-09-28
Experimentally use a single dispatcher thread per LS

ahmedneilhussain:
DRAFT - For design discussion only!!

This PR illustrates a possible solution to the concurrency issues that https://github.com/eclipse/lsp4e/pull/174 was attempting to fix in an ad-hoc manner for a single use case. The present approach is more general and offers an API that should allow many places in the plugin to benefit from stronger ordering guarantees whilst also simplifying their own code.

Motivation:
Following the API of LSP4j on which LSP4e is layered, the code makes heavy use of `CompletableFuture` to interact with the server asynchronously, specifically without blocking the UI thread.

The `CompletableFuture` pattern allows the code to schedule asynchronous blocks of work to be done when the server responds to requests. If the server is multithreaded then response processing need not correlate to the order requests are sent: long-running requests need not block short-running ones. However in order to achieve non-blocking behaviour, the current pattern of usage can cause the code to *dispatch* requests to the server in an unpredictable order.

E.g. if `a()`, `b()` and `c()` are LSP4j-style calls to the language server that each return a `CompletableFuture`, then the code
`a().thenComposeAsync(b);
a().thenComposeAsync(c);`
guarantees that `b` and `c` are seen by the server after `a` *completes*, but can be seen by the server in either order wrt one another.

We think this me be the course of some general erratic behaviour in the code: e.g. the server being sent a `HoverRequest` for a `Location` that does not exist in the server's view of the document, because the server has yet to receive the modification that created it.

Chaining futures together, e.g. `a().thenComposeAsync(b()).thenComposeAsync(c())` avoids this problem, but at the expense of throughput - each operation has to wait for the server to *respond* to the previous request, not just for it to have received it.

Solution:
1. We note that the effect of ordered dispatch could be achieved by simply calling the language server directly rather than via chaining to the completable futures currently exposed by `LanguageServerAccessor` and `LanguageServerWrapper`. This follows when considering that ultimately a call to the exposed proxy is just writing JSON messages in the current thread to a stream, protected by a lock: see https://github.com/eclipse/lsp4j/blob/3b83dafdd0c9ed15e86fbfe6c464f83aaf757505/org.eclipse.lsp4j.jsonrpc/src/main/java/org/eclipse/lsp4j/jsonrpc/json/StreamMessageConsumer.java#L66
2. However, while we expect message dispatch to generally be fast, occasionally large messages might take non-negligible time to be serialised to JSON, and it is also possible that writing to the output stream could block if piped to an unresponsive server. Therefore synchronous dispatch could be problematic, particularly if invoked from the UI thread. It is also desirable to oblige requests to the server to at least wait for initialisation to complete.
3. Therefore, a single-threaded executor is created for each language server, using the Java convenience API. This has a single unbounded queue.
4. Once the server is started, document change notifications and any requests made using the new API will be dispatched on this queue in the order in which they are made.
5. Calls using the new API on `LanguageServerWrapper` will be dispatched in the order that they are made. Specifically, any request made from the UI thread will result in the server receiving the request after all previous document updates, and before any new ones. The server's view of a given document should therefore match that of the client *at the point the request was made*. Essentially the server's timeline of modifications should be the same as the client's, just with a possible lag.
6. A convenience API is provided on `LanguageServiceAccessor` to handle the common case of aggregating results over all applicable language servers. This also guarantees that client code invoking it cannot block any server responses or inbound requests/notifications: this is important when considering that https://github.com/eclipse/lsp4j/blob/3b83dafdd0c9ed15e86fbfe6c464f83aaf757505/org.eclipse.lsp4j.jsonrpc/src/main/java/org/eclipse/lsp4j/jsonrpc/RemoteEndpoint.java#L184 dispatches synchronously from the thread reading from the server's output stream: if code calls `languageServerFunction().andThen( < arbitrary code block > )` the chained code block will execute on the foreground in the inbound dispatch thread when the future completes. This would block further inbound messages from being read and potentially even lead to deadlock.
7. Document hover, colour mining and hyperlink detection have been ported over to demonstrate this new API.

Note that while the ordering guaranteed in 5 allows the client to predict what the server's view of a document will be when it receives a request, there is still the problem that the document may have changed by the time the response arrives. However, this can be guarded against using an optimistic locking approach. E.g. document formatting can use the following pseudocode:

image

Without the dispatch ordering, it would not be safe to assume that the server would be up to date with `v1` when receiving the format request.

### DIFF
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/DocumentContentSynchronizer.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/DocumentContentSynchronizer.java
@@ -18,14 +18,10 @@ import java.net.URI;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
-import java.util.concurrent.CompletableFuture;
-import java.util.concurrent.CompletionStage;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
-import java.util.concurrent.atomic.AtomicReference;
-import java.util.function.Function;
 
 import org.eclipse.core.filesystem.EFS;
 import org.eclipse.core.filesystem.IFileStore;
@@ -72,7 +68,7 @@ final class DocumentContentSynchronizer implements IDocumentListener {
 	private int version = 0;
 	private DidChangeTextDocumentParams changeParams;
 	private long modificationStamp;
-	private final AtomicReference<CompletableFuture<LanguageServer>> lastChangeFuture;
+//	private final AtomicReference<CompletableFuture<LanguageServer>> lastChangeFuture;
 	private LanguageServer languageServer;
 	private IPreferenceStore store;
 
@@ -115,36 +111,37 @@ final class DocumentContentSynchronizer implements IDocumentListener {
 
 		textDocument.setLanguageId(languageId);
 		textDocument.setVersion(++version);
-		lastChangeFuture = new AtomicReference<>(languageServerWrapper.getInitializedServer().thenApplyAsync(ls -> {
-			this.languageServer = ls;
-			ls.getTextDocumentService().didOpen(new DidOpenTextDocumentParams(textDocument));
-			return ls;
-		}));
+//		lastChangeFuture = new AtomicReference<>(languageServerWrapper.getInitializedServer().thenApplyAsync(ls -> {
+//			this.languageServer = ls;
+//			ls.getTextDocumentService().didOpen(new DidOpenTextDocumentParams(textDocument));
+//			return ls;
+//		}));
+		this.languageServerWrapper.notifyOnLatestVersion(ls -> ls.getTextDocumentService().didOpen(new DidOpenTextDocumentParams(textDocument)));
 	}
 
-	/**
- 	 * Submit an asynchronous call (i.e. to the language server) that be inserted into the current position w.r.t.
- 	 * document change events
- 	 *
- 	 * @param <U> Computation return type
- 	 * @param fn Asynchronous computation on the language server
- 	 * @return Asynchronous result object.
- 	 */
-	private <U> @NonNull CompletableFuture<U> executeOnCurrentVersionAsync(
-			Function<LanguageServer, ? extends CompletionStage<U>> fn) {
-		AtomicReference<CompletableFuture<U>> resValueFuture = new AtomicReference<>();
-		lastChangeFuture.updateAndGet(f -> {
-			CompletableFuture<U> valueFuture = f.thenComposeAsync(fn);
-			resValueFuture.set(valueFuture);
-			// We ignore any exceptions that happen when executing the given future
-			return valueFuture.handle((value, error) -> this.languageServer);
-		});
-		return resValueFuture.get();
-	}
+//	/**
+// 	 * Submit an asynchronous call (i.e. to the language server) that be inserted into the current position w.r.t.
+// 	 * document change events
+// 	 *
+// 	 * @param <U> Computation return type
+// 	 * @param fn Asynchronous computation on the language server
+// 	 * @return Asynchronous result object.
+// 	 */
+//	private <U> @NonNull CompletableFuture<U> executeOnCurrentVersionAsync(
+//			Function<LanguageServer, ? extends CompletionStage<U>> fn) {
+//		AtomicReference<CompletableFuture<U>> resValueFuture = new AtomicReference<>();
+//		lastChangeFuture.updateAndGet(f -> {
+//			CompletableFuture<U> valueFuture = f.thenComposeAsync(fn);
+//			resValueFuture.set(valueFuture);
+//			// We ignore any exceptions that happen when executing the given future
+//			return valueFuture.handle((value, error) -> this.languageServer);
+//		});
+//		return resValueFuture.get();
+//	}
 
-	CompletableFuture<LanguageServer> lastChangeFuture() {
-		return lastChangeFuture.get();
-	}
+//	CompletableFuture<LanguageServer> lastChangeFuture() {
+//		return lastChangeFuture.get();
+//	}
 
 	@Override
 	public void documentChanged(DocumentEvent event) {
@@ -158,10 +155,11 @@ final class DocumentContentSynchronizer implements IDocumentListener {
 			changeParams = null;
 
 			changeParamsToSend.getTextDocument().setVersion(++version);
-			lastChangeFuture.updateAndGet(f -> f.thenApplyAsync(ls -> {
-				ls.getTextDocumentService().didChange(changeParamsToSend);
-				return ls;
-			}));
+//			lastChangeFuture.updateAndGet(f -> f.thenApplyAsync(ls -> {
+//				ls.getTextDocumentService().didChange(changeParamsToSend);
+//				return ls;
+//			}));
+			this.languageServerWrapper.notifyOnLatestVersion(ls -> ls.getTextDocumentService().didChange(changeParamsToSend));
 		}
 	}
 
@@ -269,7 +267,7 @@ final class DocumentContentSynchronizer implements IDocumentListener {
 
 
 		try {
-			List<TextEdit> edits = executeOnCurrentVersionAsync(ls -> ls.getTextDocumentService().willSaveWaitUntil(params))
+			List<TextEdit> edits = this.languageServerWrapper.executeOnLatestVersion(ls -> ls.getTextDocumentService().willSaveWaitUntil(params))
 				.get(lsToWillSaveWaitUntilTimeout(), TimeUnit.SECONDS);
 			try {
 				LSPEclipseUtils.applyEdits(document, edits);
@@ -304,10 +302,12 @@ final class DocumentContentSynchronizer implements IDocumentListener {
 		}
 		TextDocumentIdentifier identifier = new TextDocumentIdentifier(fileUri.toString());
 		DidSaveTextDocumentParams params = new DidSaveTextDocumentParams(identifier, document.get());
-		lastChangeFuture.updateAndGet(f -> f.thenApplyAsync(ls -> {
-  			ls.getTextDocumentService().didSave(params);
-  			return ls;
-  		}));
+//		lastChangeFuture.updateAndGet(f -> f.thenApplyAsync(ls -> {
+//  			ls.getTextDocumentService().didSave(params);
+//  			return ls;
+//  		}));
+		this.languageServerWrapper.notifyOnLatestVersion(ls -> ls.getTextDocumentService().didSave(params));
+
 	}
 
 	public void documentClosed() {
@@ -318,7 +318,7 @@ final class DocumentContentSynchronizer implements IDocumentListener {
 		if (languageServerWrapper.isActive()) {
 			TextDocumentIdentifier identifier = new TextDocumentIdentifier(uri);
 			DidCloseTextDocumentParams params = new DidCloseTextDocumentParams(identifier);
-			lastChangeFuture.get().thenAcceptAsync(ls -> ls.getTextDocumentService().didClose(params));
+			languageServerWrapper.notifyOnLatestVersion(ls -> ls.getTextDocumentService().didClose(params));
 		}
 	}
 
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/DocumentContentSynchronizer.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/DocumentContentSynchronizer.java
@@ -68,8 +68,7 @@ final class DocumentContentSynchronizer implements IDocumentListener {
 	private int version = 0;
 	private DidChangeTextDocumentParams changeParams;
 	private long modificationStamp;
-//	private final AtomicReference<CompletableFuture<LanguageServer>> lastChangeFuture;
-	private LanguageServer languageServer;
+
 	private IPreferenceStore store;
 
 	public DocumentContentSynchronizer(@NonNull LanguageServerWrapper languageServerWrapper,
@@ -111,37 +110,10 @@ final class DocumentContentSynchronizer implements IDocumentListener {
 
 		textDocument.setLanguageId(languageId);
 		textDocument.setVersion(++version);
-//		lastChangeFuture = new AtomicReference<>(languageServerWrapper.getInitializedServer().thenApplyAsync(ls -> {
-//			this.languageServer = ls;
-//			ls.getTextDocumentService().didOpen(new DidOpenTextDocumentParams(textDocument));
-//			return ls;
-//		}));
+
 		this.languageServerWrapper.notifyOnLatestVersion(ls -> ls.getTextDocumentService().didOpen(new DidOpenTextDocumentParams(textDocument)));
 	}
 
-//	/**
-// 	 * Submit an asynchronous call (i.e. to the language server) that be inserted into the current position w.r.t.
-// 	 * document change events
-// 	 *
-// 	 * @param <U> Computation return type
-// 	 * @param fn Asynchronous computation on the language server
-// 	 * @return Asynchronous result object.
-// 	 */
-//	private <U> @NonNull CompletableFuture<U> executeOnCurrentVersionAsync(
-//			Function<LanguageServer, ? extends CompletionStage<U>> fn) {
-//		AtomicReference<CompletableFuture<U>> resValueFuture = new AtomicReference<>();
-//		lastChangeFuture.updateAndGet(f -> {
-//			CompletableFuture<U> valueFuture = f.thenComposeAsync(fn);
-//			resValueFuture.set(valueFuture);
-//			// We ignore any exceptions that happen when executing the given future
-//			return valueFuture.handle((value, error) -> this.languageServer);
-//		});
-//		return resValueFuture.get();
-//	}
-
-//	CompletableFuture<LanguageServer> lastChangeFuture() {
-//		return lastChangeFuture.get();
-//	}
 
 	@Override
 	public void documentChanged(DocumentEvent event) {
@@ -155,10 +127,6 @@ final class DocumentContentSynchronizer implements IDocumentListener {
 			changeParams = null;
 
 			changeParamsToSend.getTextDocument().setVersion(++version);
-//			lastChangeFuture.updateAndGet(f -> f.thenApplyAsync(ls -> {
-//				ls.getTextDocumentService().didChange(changeParamsToSend);
-//				return ls;
-//			}));
 			this.languageServerWrapper.notifyOnLatestVersion(ls -> ls.getTextDocumentService().didChange(changeParamsToSend));
 		}
 	}
@@ -302,10 +270,7 @@ final class DocumentContentSynchronizer implements IDocumentListener {
 		}
 		TextDocumentIdentifier identifier = new TextDocumentIdentifier(fileUri.toString());
 		DidSaveTextDocumentParams params = new DidSaveTextDocumentParams(identifier, document.get());
-//		lastChangeFuture.updateAndGet(f -> f.thenApplyAsync(ls -> {
-//  			ls.getTextDocumentService().didSave(params);
-//  			return ls;
-//  		}));
+
 		this.languageServerWrapper.notifyOnLatestVersion(ls -> ls.getTextDocumentService().didSave(params));
 
 	}
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/DocumentContentSynchronizer.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/DocumentContentSynchronizer.java
@@ -55,7 +55,6 @@ import org.eclipse.lsp4j.TextEdit;
 import org.eclipse.lsp4j.VersionedTextDocumentIdentifier;
 import org.eclipse.lsp4j.WillSaveTextDocumentParams;
 import org.eclipse.lsp4j.jsonrpc.messages.Either;
-import org.eclipse.lsp4j.services.LanguageServer;
 import org.eclipse.osgi.util.NLS;
 
 final class DocumentContentSynchronizer implements IDocumentListener {
@@ -127,7 +126,7 @@ final class DocumentContentSynchronizer implements IDocumentListener {
 			changeParams = null;
 
 			changeParamsToSend.getTextDocument().setVersion(++version);
-			this.languageServerWrapper.notifyOnLatestVersion(ls -> ls.getTextDocumentService().didChange(changeParamsToSend));
+			languageServerWrapper.notifyOnLatestVersion(ls -> ls.getTextDocumentService().didChange(changeParamsToSend));
 		}
 	}
 
@@ -235,7 +234,7 @@ final class DocumentContentSynchronizer implements IDocumentListener {
 
 
 		try {
-			List<TextEdit> edits = this.languageServerWrapper.executeOnLatestVersion(ls -> ls.getTextDocumentService().willSaveWaitUntil(params))
+			List<TextEdit> edits = languageServerWrapper.executeOnLatestVersion(ls -> ls.getTextDocumentService().willSaveWaitUntil(params))
 				.get(lsToWillSaveWaitUntilTimeout(), TimeUnit.SECONDS);
 			try {
 				LSPEclipseUtils.applyEdits(document, edits);
@@ -271,7 +270,7 @@ final class DocumentContentSynchronizer implements IDocumentListener {
 		TextDocumentIdentifier identifier = new TextDocumentIdentifier(fileUri.toString());
 		DidSaveTextDocumentParams params = new DidSaveTextDocumentParams(identifier, document.get());
 
-		this.languageServerWrapper.notifyOnLatestVersion(ls -> ls.getTextDocumentService().didSave(params));
+		languageServerWrapper.notifyOnLatestVersion(ls -> ls.getTextDocumentService().didSave(params));
 
 	}
 
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/DocumentContentSynchronizer.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/DocumentContentSynchronizer.java
@@ -55,6 +55,7 @@ import org.eclipse.lsp4j.TextEdit;
 import org.eclipse.lsp4j.VersionedTextDocumentIdentifier;
 import org.eclipse.lsp4j.WillSaveTextDocumentParams;
 import org.eclipse.lsp4j.jsonrpc.messages.Either;
+import org.eclipse.lsp4j.services.LanguageServer;
 import org.eclipse.osgi.util.NLS;
 
 final class DocumentContentSynchronizer implements IDocumentListener {
@@ -71,6 +72,7 @@ final class DocumentContentSynchronizer implements IDocumentListener {
 	private IPreferenceStore store;
 
 	public DocumentContentSynchronizer(@NonNull LanguageServerWrapper languageServerWrapper,
+			@NonNull LanguageServer languageServer,
 			@NonNull IDocument document, TextDocumentSyncKind syncKind) {
 		this.languageServerWrapper = languageServerWrapper;
 		URI uri = LSPEclipseUtils.toUri(document);
@@ -109,8 +111,7 @@ final class DocumentContentSynchronizer implements IDocumentListener {
 
 		textDocument.setLanguageId(languageId);
 		textDocument.setVersion(++version);
-
-		this.languageServerWrapper.notifyOnLatestVersion(ls -> ls.getTextDocumentService().didOpen(new DidOpenTextDocumentParams(textDocument)));
+		languageServer.getTextDocumentService().didOpen(new DidOpenTextDocumentParams(textDocument));
 	}
 
 
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServerWrapper.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServerWrapper.java
@@ -646,10 +646,10 @@ public class LanguageServerWrapper {
 			return null;
 		}
 		final IDocument theDocument = document;
-		return initializeFuture.thenComposeAsync(theVoid -> {
+		return initializeFuture.thenApplyAsync(theVoid -> {
 			synchronized (connectedDocuments) {
 				if (this.connectedDocuments.containsKey(uri)) {
-					return CompletableFuture.completedFuture(null);
+					return null;
 				}
 				Either<TextDocumentSyncKind, TextDocumentSyncOptions> syncOptions = initializeFuture == null ? null
 						: this.serverCapabilities.getTextDocumentSync();
@@ -661,12 +661,13 @@ public class LanguageServerWrapper {
 						syncKind = syncOptions.getLeft();
 					}
 				}
-				DocumentContentSynchronizer listener = new DocumentContentSynchronizer(this, theDocument, syncKind);
+
+				DocumentContentSynchronizer listener = new DocumentContentSynchronizer(this, languageServer, theDocument, syncKind);
 				theDocument.addDocumentListener(listener);
 				LanguageServerWrapper.this.connectedDocuments.put(uri, listener);
-				return CompletableFuture.completedFuture(languageServer);
+				return languageServer;
 			}
-		}, dispatcher).thenApply(theVoid -> languageServer);
+		}).thenApply(theVoid -> languageServer);
 	}
 
 	public void disconnect(URI uri) {
@@ -792,7 +793,7 @@ public class LanguageServerWrapper {
 	 * null if the server does not support the requested capabilities or could not be started.
 	 */
 	@NonNull CompletableFuture<@Nullable LanguageServerWrapper> connectIf(@NonNull IDocument document, @NonNull Predicate<ServerCapabilities> filter) {
-		return getInitializedServer().thenComposeAsync(server -> {
+		return getInitializedServer().thenCompose(server -> {
 			if (server != null && (filter == null || filter.test(getServerCapabilities()))) {
 				try {
 					return connect(document);
@@ -801,7 +802,7 @@ public class LanguageServerWrapper {
 				}
 			}
 			return CompletableFuture.completedFuture(null);
-		}, dispatcher).thenApply(server -> server == null ? null : this);
+		}).thenApply(server -> server == null ? null : this);
 	}
 
 	/**
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServerWrapper.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServerWrapper.java
@@ -768,7 +768,7 @@ public class LanguageServerWrapper {
 	 * @return Async result
 	 */
 	<T> CompletableFuture<T> executeOnLatestVersion(Function<LanguageServer, ? extends CompletionStage<T>> fn) {
-		return getInitializedServer().thenComposeAsync(fn, this.dispatcher);
+ 		return getInitializedServer().thenComposeAsync(fn, this.dispatcher);
 	}
 	/**
 	 * Sends a notification to the LS on a single executor thread tied to this server. Use of this guarantees that the server
@@ -800,7 +800,7 @@ public class LanguageServerWrapper {
 				}
 			}
 			return CompletableFuture.completedFuture(null);
-		}, this.dispatcher).thenApply(server -> this);
+		}, this.dispatcher).thenApply(server -> server == null ? null : this);
 	}
 
 	/**
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServerWrapper.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServerWrapper.java
@@ -767,7 +767,8 @@ public class LanguageServerWrapper {
 	 * @param fn LS method to invoke
 	 * @return Async result
 	 */
-	<T> CompletableFuture<T> executeOnLatestVersion(Function<LanguageServer, ? extends CompletionStage<T>> fn) {
+	@NonNull
+	<T> CompletableFuture<T> executeOnLatestVersion(@NonNull Function<LanguageServer, ? extends CompletionStage<T>> fn) {
  		return getInitializedServer().thenComposeAsync(fn, this.dispatcher);
 	}
 	/**
@@ -775,7 +776,7 @@ public class LanguageServerWrapper {
 	 * will have seen all previous such requests/notifications (and document updates).
 	 * @param fn LS notification to send
 	 */
-	void notifyOnLatestVersion(Consumer<LanguageServer> fn) {
+	void notifyOnLatestVersion(@NonNull Consumer<LanguageServer> fn) {
 		getInitializedServer().thenAcceptAsync(fn, this.dispatcher);
 	}
 
@@ -790,7 +791,7 @@ public class LanguageServerWrapper {
 	 * @return Async result that guarantees the wrapped server will be active and connected to the document. Wraps
 	 * null if the server does not support the requested capabilities or could not be started.
 	 */
-	CompletableFuture<LanguageServerWrapper> connectIf(IDocument document, Predicate<ServerCapabilities> filter) {
+	@NonNull CompletableFuture<@Nullable LanguageServerWrapper> connectIf(@NonNull IDocument document, @NonNull Predicate<ServerCapabilities> filter) {
 		return getInitializedServer().thenComposeAsync(server -> {
 			if (server != null && (filter == null || filter.test(getServerCapabilities()))) {
 				try {
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServerWrapper.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServerWrapper.java
@@ -666,7 +666,7 @@ public class LanguageServerWrapper {
 				LanguageServerWrapper.this.connectedDocuments.put(uri, listener);
 				return CompletableFuture.completedFuture(languageServer);
 			}
-		}, this.dispatcher).thenApply(theVoid -> languageServer);
+		}, dispatcher).thenApply(theVoid -> languageServer);
 	}
 
 	public void disconnect(URI uri) {
@@ -801,7 +801,7 @@ public class LanguageServerWrapper {
 				}
 			}
 			return CompletableFuture.completedFuture(null);
-		}, this.dispatcher).thenApply(server -> server == null ? null : this);
+		}, dispatcher).thenApply(server -> server == null ? null : this);
 	}
 
 	/**
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServerWrapper.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServerWrapper.java
@@ -31,6 +31,7 @@ import java.util.Timer;
 import java.util.TimerTask;
 import java.util.concurrent.CancellationException;
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionStage;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
@@ -38,6 +39,9 @@ import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
 import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.function.Consumer;
+import java.util.function.Function;
+import java.util.function.Predicate;
 import java.util.function.UnaryOperator;
 import java.util.stream.Collectors;
 
@@ -187,6 +191,8 @@ public class LanguageServerWrapper {
 	private Timer timer;
 	private AtomicBoolean stopping = new AtomicBoolean(false);
 
+	private final ExecutorService dispatcher;
+
 	/**
 	 * Map containing unregistration handlers for dynamic capability registrations.
 	 */
@@ -210,6 +216,14 @@ public class LanguageServerWrapper {
 		this.initialPath = initialPath;
 		this.serverDefinition = serverDefinition;
 		this.connectedDocuments = new HashMap<>();
+		String threadNameFormat = "LS-" + serverDefinition.id + "-dispatcher-%d"; //$NON-NLS-1$ //$NON-NLS-2$
+		this.dispatcher = Executors
+				.newSingleThreadExecutor(new ThreadFactoryBuilder().setNameFormat(threadNameFormat).build());
+
+	}
+
+	void stopDispatcher() {
+		this.dispatcher.shutdownNow();
 	}
 
 	/**
@@ -650,9 +664,9 @@ public class LanguageServerWrapper {
 				DocumentContentSynchronizer listener = new DocumentContentSynchronizer(this, theDocument, syncKind);
 				theDocument.addDocumentListener(listener);
 				LanguageServerWrapper.this.connectedDocuments.put(uri, listener);
-				return listener.lastChangeFuture();
+				return CompletableFuture.completedFuture(languageServer);
 			}
-		}).thenApply(theVoid -> languageServer);
+		}, this.dispatcher).thenApply(theVoid -> languageServer);
 	}
 
 	public void disconnect(URI uri) {
@@ -743,6 +757,50 @@ public class LanguageServerWrapper {
 			}
 		}
 		return CompletableFuture.completedFuture(this.languageServer);
+	}
+
+	/**
+	 * Dispatches a LS request on a single executor thread tied to this server. Use of this guarantees that the server
+	 * will have seen all previous such requests (and document updates), and will not receive any further updates
+	 * prior to receiving this one.
+	 * @param <T> LS response type
+	 * @param fn LS method to invoke
+	 * @return Async result
+	 */
+	<T> CompletableFuture<T> executeOnLatestVersion(Function<LanguageServer, ? extends CompletionStage<T>> fn) {
+		return getInitializedServer().thenComposeAsync(fn, this.dispatcher);
+	}
+	/**
+	 * Sends a notification to the LS on a single executor thread tied to this server. Use of this guarantees that the server
+	 * will have seen all previous such requests/notifications (and document updates).
+	 * @param fn LS notification to send
+	 */
+	void notifyOnLatestVersion(Consumer<LanguageServer> fn) {
+		getInitializedServer().thenAcceptAsync(fn, this.dispatcher);
+	}
+
+	/**
+	 * Test whether this server supports the requested <code>ServerCapabilities</code>, and ensure
+	 * that it is connected to the document if so.
+	 *
+	 * NB result is a future on this <emph>wrapper</emph> rather than the wrapped language server directly,
+	 * to support accessing the server on the single-threaded dispatch queue.
+	 * @param document Document to connect
+	 * @param filter Constraint on server capabilities
+	 * @return Async result that guarantees the wrapped server will be active and connected to the document. Wraps
+	 * null if the server does not support the requested capabilities or could not be started.
+	 */
+	CompletableFuture<LanguageServerWrapper> connectIf(IDocument document, Predicate<ServerCapabilities> filter) {
+		return getInitializedServer().thenComposeAsync(server -> {
+			if (server != null && (filter == null || filter.test(getServerCapabilities()))) {
+				try {
+					return connect(document);
+				} catch (IOException ex) {
+					LanguageServerPlugin.logError(ex);
+				}
+			}
+			return CompletableFuture.completedFuture(null);
+		}, this.dispatcher).thenApply(server -> this);
 	}
 
 	/**
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServiceAccessor.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServiceAccessor.java
@@ -28,8 +28,10 @@ import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionStage;
 import java.util.concurrent.CopyOnWriteArraySet;
 import java.util.concurrent.ExecutionException;
+import java.util.function.Function;
 import java.util.function.Predicate;
 import java.util.stream.Collectors;
 
@@ -626,6 +628,73 @@ public class LanguageServiceAccessor {
 			LanguageServerPlugin.logError(e);
 		}
 		return CompletableFuture.completedFuture(Collections.emptyList());
+	}
+
+	static void shutdownAllDispatchers() {
+		startedServers.forEach(s -> s.stopDispatcher());
+	}
+
+	/**
+	 * Make a call to the language server(s), aggregating the responses into a single CompletableFuture if
+	 * more than one language server is applicable. The call will be enqueued such that it will be sent to the
+	 * language servers after any previous calls using this framework and before any subsequent calls. More
+	 * specifically, if called from the UI thread then all document updates to date are guaranteed to have
+	 * been seen by any server before it sees this request, and before it sees any further
+	 * document updates.
+	 *
+	 * <p/>The returned result will complete when all language servers have responded,
+	 * and will return its result in a pool thread to avoid blocking the inbound message readers.
+	 *
+	 * @param <T> Return type of method being called on lang server
+	 * @param document Document on which the request is being made
+	 * @param filter Restriction on capabilities of the language servers we're interested in
+	 * @param fn A single method invocation on <code>LanguageServer</code>
+	 * @return Async result aggregated over all applicable lang servers, filtering out nulls.
+	 */
+	public static <T> CompletableFuture<List<T>> computeOnServers(@NonNull IDocument document, Predicate<ServerCapabilities> filter,
+			Function<LanguageServer, ? extends CompletionStage<T>> fn) {
+
+		// Out-of-line so we can declare it as List rather than ArrayList to avoid type errors below
+		CompletableFuture<List<T>> init = CompletableFuture.completedFuture(new ArrayList<T>());
+
+		return getLSWrappers(document).stream()
+			// Ensure wrappers are started, connected to the document, and filter for capabilities
+			.map(wrapper -> wrapper.connectIf(document, filter)
+				// Call fn on lang servers, excluding null servers (that failed to start/connect or do not have the required capability)
+				.thenCompose(w -> w == null ? CompletableFuture.completedFuture((T)null) : w.executeOnLatestVersion(fn)))
+
+			// Transform individual async results into a single async with the aggregate result
+			.reduce(init, LanguageServiceAccessor::combine, LanguageServiceAccessor::concatResults)
+
+			// Ensure any subsequent computation added by caller does not block further incoming messages from language servers
+			.thenApplyAsync(t -> t);
+	}
+
+	/**
+	 * Accumulator that appends the result of an async computation onto an async aggregate result. Nulls will be excluded.
+	 * @param <T> Result type
+	 * @param result Async aggregate result
+	 * @param next Pending result to include
+	 * @return
+	 */
+	static <T> CompletableFuture<List<T>> combine(CompletableFuture<? extends List<T>> result, CompletableFuture<T> next) {
+		return result.thenCombine(next, (List<T> a, T b) -> {
+			if (b != null) {
+				a.add(b);
+			}
+			return a;
+		});
+	}
+
+	/**
+	 * Merges two async sets of results into a single async result
+	 * @param <T> Result type
+	 * @param a First async result
+	 * @param b Second async result
+	 * @return Async combined result
+	 */
+	public static <T> CompletableFuture<List<T>> concatResults(CompletableFuture<List<T>> a, CompletableFuture<List<T>> b) {
+		return a.thenCombine(b, (c, d) -> { c.addAll(d); return c; });
 	}
 
 	public static boolean checkCapability(LanguageServer languageServer, Predicate<ServerCapabilities> condition) {
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServiceAccessor.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServiceAccessor.java
@@ -140,6 +140,16 @@ public class LanguageServiceAccessor {
 			return this.wrapper.getServerCapabilities();
 		}
 
+		/**
+		 * Make a request to the language server such that it will be received only after any pending document updates, and before any subsequent changes
+		 * @param <T> Result type
+		 * @param fn language server operation
+		 * @return Async result
+		 */
+		public @NonNull <T> CompletableFuture<@Nullable T> computeOnLatestVersion(@NonNull Function<LanguageServer, ? extends CompletionStage<T>> fn) {
+			return this.wrapper.executeOnLatestVersion(fn);
+		}
+
 		public boolean isActive() {
 			return this.wrapper.isActive();
 		}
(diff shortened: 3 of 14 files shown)


ahmedneilhussain:
@mickaelistria food for thought! Please note that this PR is not intended to be consumed as-is, as I have not even started looking at e.g. the testing, or even tried the code! It is intended to be the basis for discussing a better, more general solution to https://github.com/eclipse/lsp4e/pull/174

When you have time, perhaps you could pick through the analysis and see if you concur with our observations about ordered dispatch vs synchronous dispatch vs asynchronous response..!

mickaelistria:
My concern with this proposed approach is that it doesn't seem very transparent: consumers would have to adapt a fair amount of code to leverage the guarantee that the pending change notifications are sent before sending the request. (or maybe I misunderstood something?)
I'm still under impression that https://github.com/eclipse/lsp4e/pull/174#issuecomment-1247774116 can actually be a solution (your reaction is that we cannot guarantee more than what is offered there) which would be easier to leverage for consumers. We may even be able to merge it in `getInitializedLanguageServers` so no new consumer API is necessary.

tomgr:
The problem with https://github.com/eclipse/lsp4e/pull/174#issuecomment-1247774116 is that doing `andThen(o)` on the `lastChangeWasSent()` result would allow a subsequent edit to overtake `o`, meaning that `o` gets executed on the wrong version of the document. Your suggestion of also checking `getModificationStamp` is sufficient to fix the formatting but not other commands unless they are also adapted to check the modification stamp. It would be good to fix this properly for all commands, as we do see odd requests from LSP4e sometimes where the server considers the positions to be out of bounds, presumably because an edit has happened in the wrong order versus, e.g. a hover request.

If a command uses the approach outlined in this PR, this will guarantee that the request is sent to the server in exactly the correct position: all edits will have been applied, and further no more edits (or other commands) can be sent until the request has been sent. And what's also nice about this approach is that we don't unnecessarily block the client-side. The commands are sent to the server in the correct order, but we do not wait for the server to complete the first command before sending the second. This is useful for more complex requests, such as the workspace-wide requests like rename.

rubenporras:
Hi @tomgr , for what is worth, I can confirm that we see very often that the server considers a the positions to be out of bounds. I never analyzed the problem but your assumption makes sense.

tomgr:
> Hi @tomgr , for what is worth, I can confirm that we see very often that the server considers a the positions to be out of bounds. I never analyzed the problem but your assumption makes sense.

Ahh that's interesting. I'm glad we're not the only ones!

> Just as information. I plan to take a look at this and the related PR in detail next week.

Great thanks :) I think the approach @ahmedneilhussain worked on in this PR is a better way to go as, with a bit of effort, we could fix all commands.

mickaelistria:
@tomgr I don't get how this proposal here make the modification stamps check useless. eg it's still very possible that the `format` request takes a long time to process and that document gets changed in the meantime, making the result possibly invalid. I don't see that checked in the proposal, did I miss something?

Maybe we could have a specific implementation of `LanguageServer` in LSP4E which would delegate to the underlying language server, but adding extra checks for some operations (eg wait for all change event to be sent before sending a request, verify the stamp before/after to validate the result and if necessary cancel/resend...).

tomgr:
> @tomgr I don't get how this proposal here make the modification stamps check useless. eg it's still very possible that the `format` request takes a long time to process and that document gets changed in the meantime, making the result possibly invalid. I don't see that checked in the proposal, did I miss something?

No you've not missed anything. It is definitely still required for the reason you outline. What this PR does is fix the issue from the server's perspective. If all commands were updated to use this API, then the server will never be sent an invalid request. At the moment, LSP4e can issue requests to format/display a hover/find references/etc on locations that are out of bounds because the edits are sent in the wrong sequence versus the requests. Any command that alters a user's buffer would definitely still need a modification stamp check.

> Maybe we could have a specific implementation of `LanguageServer` in LSP4E which would delegate to the underlying language server, but adding extra checks for some operations (eg wait for all change event to be sent before sending a request, verify the stamp before/after to validate the result and if necessary cancel/resend...).

I don't think this is enough. There is a fundamental problem with using futures to send requests to the language server, which is that the extra `then*Async` calls that appear in the various client methods mean that there is no guarantee about the order in which these futures are executed. For example, consider `initiateHoverRequest` which contains:
https://github.com/eclipse/lsp4e/blob/114422a698342ec6b300411674c75c4884faf55f/org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/hover/LSPTextHover.java#L197-L212
The `thenApplyAsync` at the start of this makes this unfixable, I think. Specifically, it means that by the time the hover request is sent (via the languageServer.getTextDocumentService().hover() call), arbitrary edits may have been sent to the server. To fix this, we would have to ensure that `.thenApplyAsync(f)` would insert `f` into a FIFO queue. Ideally we would do this automatically, but I don't think there is a way of doing this within the Java Futures API. Hence the approach in this PR.

mickaelistria:
OK, so if the goal is to ensure that the order of the future is respected on client side, we may just be able to replace the CacheThredPool in LanguageServerWrapper, line 254 by a `Executors.singleThreadExecutor` ?

tomgr:
`thenComposeAsync` doesn't propagate the executor, so if you do `f1.thenComposeAsync(f2)`, `f2` could be executed with a different executor than `f1`. So unfortunately that approach wouldn't work for anything apart from the initial future.

martinlippert:
I am not sure I understand all the details here, but I find https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#implementationConsiderations very interesting within this context.

It feels to me like lsp4e needs to make sure that document change events are sent to the server using a single threaded executor with an unbounded queue to guarantee that those changes are sent to the server in the right order. The exact order of requests are not really important - as long as the client keeps track of the document state/version for which the request was initiated, cancels requests if additional document edits happen before the request is actually being send or the response arrives, and checks the current document state/version against the one for which the request was initiated when receiving the response - and ignoring the response when it is outdated.

Whether to re-send the request or not in such a case sounds like a difficult decision and probably depends on the individual request.

mickaelistria:
> thenComposeAsync doesn't propagate the executor, so if you do f1.thenComposeAsync(f2), f2 could be executed with a different executor than f1

If we change the executor, then is gives the gurantee than any `languageServer.request()` is on the same single thread executor; even if you compose it, as long as `languageServer` is dereferenced, the single thread executor is used. In practice, it means that the request are received by the LS in the same order as the request are requested in LSP4E code. The requests flow becomes more sequential. Isn't it most of what is required here?

mickaelistria:
The thing is that we do not want to make eg handing a hover blocking for eg updating outline or even processing a further edit. If both the "show hover" and "update outline" hover happen on the same single thread executor, then it means they cannot happen in parallel and completion of the former will block the start of the later; and imagine if the users edit the file in the meantine, even the document change event would be waiting, leading to a state where the IDE and the LS are more frequently out of sync about document content.
What you have in mind with your example is actually more about cancellation, which is also under discussion at #238

tomgr:
> The thing is that we do not want to make eg handing a hover blocking for eg updating outline or even processing a further edit. 

I agree that this would be bad, but I'm not proposing that the *completion* of the hover would block the outline request. I'm proposing that the *sending* of the hover request would block the sending of the outline request. For example, I would expect the following interleaving to be allowed:
```
send(hover), send(outline), receive(outline), receive(hover)
```
This PR does ensure this.

mickaelistria:
>  I'm proposing that the sending of the hover request would block the sending of the outline request. 

That's what using the singleThreadExecutor does: however they are invoked, `languageServer.hover(...)` and `languageServer.documentSymbols(...)` would end up running sequentially on the same thread, in the order they're invoked.

ahmedneilhussain:
> > thenComposeAsync doesn't propagate the executor, so if you do f1.thenComposeAsync(f2), f2 could be executed with a different executor than f1
> 
> If we change the executor, then is gives the gurantee than any `languageServer.request()` is on the same single thread executor; even if you compose it, as long as `languageServer` is dereferenced, the single thread executor is used. In practice, it means that the request are received by the LS in the same order as the request are requested in LSP4E code. The requests flow becomes more sequential. Isn't it most of what is required here?

I initially looked at this very idea as a possible solution, but unfortunately as Tom says it won't work because `CompletableFuture` doesn't work that way :-( If you don't specifically supply an executor to an `thenComposeAsync` chaining then it will use the default thread pool. An async composition doesn't in any way remember the supplied executor for future chainings... So if you supply the single thread executor for the first document update or whatever then it will use that, but anything that chains onto the resultant future will not!

mickaelistria:
>  An async composition doesn't in any way remember the supplied executor for future chainings... 

The executor that is used to send LS request is never the default one or another contextual one, but always the one configured in the Language Server launcher.

ahmedneilhussain:
> > I'm proposing that the sending of the hover request would block the sending of the outline request.
> 
> That's what using the singleThreadExecutor does: however they are invoked, `languageServer.hover(...)` and `languageServer.documentSymbols(...)` would end up running sequentially on the same thread, in the order they're invoked.

Indeed, but don't forget that calling `languageServer.hover(...)` and `languageServer.documentSymbols(...)` return `CompletableFuture` objects. Calling them sequentially one after another means that the requests will be *dispatched* one after another, in order, but doesn't mean the client is blocked - it can carry on processing, unless it explicitly calls `get()` or `join()` on the completable future.

In order to visualise how things interleave I found it quite helpful to drill down into the LSP4j layer to see what was going on:

1. If you pick your way through https://github.com/eclipse/lsp4j/blob/main/org.eclipse.lsp4j.jsonrpc/src/main/java/org/eclipse/lsp4j/jsonrpc/services/EndpointProxy.java you can see that if you directly call a method on the `LanguageServer` object, you are calling `notfy()` or `request()` in https://github.com/eclipse/lsp4j/blob/main/org.eclipse.lsp4j.jsonrpc/src/main/java/org/eclipse/lsp4j/jsonrpc/RemoteEndpoint.java
2. This method just writes the messages as serialised JSON to a stream, https://github.com/eclipse/lsp4j/blob/main/org.eclipse.lsp4j.jsonrpc/src/main/java/org/eclipse/lsp4j/jsonrpc/json/StreamMessageConsumer.java
3. Provided the JSON serialisation is fast, and provided the server consumes the stream as fast as we write it (in practice something that might not be true and so should be guarded against), then this is instant and synchronous - if you call the language server proxy directly on the SWT event thread, then you're writing JSON messages to the stream on the SWT event thread!
4. If you're making a request, rather than sending a notification, then at the point of sending the request, `RemoteEndpoint` constructs a CompletableFuture to hold the eventual result, and stores it in a map indexed by request ID, and returns it to the caller. This is of course fast and synchronous - because constructing a completable future isn't the same as waiting for it to complete!
5. Independently, a dedicated thread listens for inbound messages from the language server. If you look at https://github.com/eclipse/lsp4j/blob/main/org.eclipse.lsp4j.jsonrpc/src/main/java/org/eclipse/lsp4j/jsonrpc/json/ConcurrentMessageProcessor.java and https://github.com/eclipse/lsp4j/blob/main/org.eclipse.lsp4j.jsonrpc/src/main/java/org/eclipse/lsp4j/jsonrpc/json/StreamMessageProducer.java it just sits in a loop reading json messages off a stream and forwarding them on to RemoteEndpoint. The dedicated thread is from the thread pool ultimately at https://github.com/eclipse/lsp4e/blob/114422a698342ec6b300411674c75c4884faf55f/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServerWrapper.java#L254
6. If the inbound message is a response to an earlier request, then it retrieves the previously-created `CompletableFuture` placeholder associated with the request ID and calls `complete(...)` with the result. A consequence is that any further work that has been composed onto that future by client code using `f.andThen(... < work > ...)` happens in the foreground of this dedicated listening thread, hence it is desirable for clients to call `f.andThenAsync( <work> )` if doing anything complicated.

So if the client has a block of code like this on the event thread

`
ls.hover().thenApply(<work1>);
ls.documentSymbols().thenApply(<work2>);
otherStuff();
`
otherStuff() will happen as soon as the two outgoing messages have been written to the stream. <work1> and <work2> will happen at some arbitrary point in the future, and never on the event thread - they will happen on the dedicated client listening thread. They could happen in either order, depending on how the ls behaves. As currently written, <work1> and <work2> could happen in either order, but not simultaneously, unless we replace `thenApply` with `thenApplyAsync`.

In this example, we are calling the LS on the event thread. This could be problematic if JSON serialisation is slow or the LS does not consume the stream as fast as we produce it. Hence using a dedicated single thread, as per this PR.

tomgr:
> The executor that is used to send LS request is never the default one or another contextual one, but always the one configured in the Language Server launcher.

As far as I can tell, the configured executor is used by the MessageProducer. That means the configured executor will be used to handle messages sent from the server -> client, not from the client -> server.

ahmedneilhussain:
> > An async composition doesn't in any way remember the supplied executor for future chainings...
> 
> The executor that is used to send LS request is never the default one or another contextual one, but always the one configured in the Language Server launcher.

I don't think that's quite correct. I did pick through the LSP4j code quite carefully and stuck some breakpoints in to understand it ;-) The executor you refer to is used for *inbound* messages - this can be responses from the server, or requests and notifications that are initiated server-side.

mickaelistria:
> As far as I can tell, the configured executor is used by the MessageProducer. That means the configured executor will be used to handle messages sent from the server -> client, not from the client -> server.

> I don't think that's quite correct. I did pick through the LSP4j code quite carefully and stuck some breakpoints in to understand it ;-) The executor you refer to is used for inbound messages - this can be responses from the server, or requests and notifications that are initiated server-side.

OK, if it is so, isn't it a bug in LSP4J then?

tomgr:
> OK, if it is so, isn't it a bug in LSP4J then?

I don't think so. LSP4j can't be responsible for creating threads that are need even prior to invoking LSP4j. The issue here is the use of futures prior to creating the LSP4j request. Once the LSP4j request is created, we're ok.

mickaelistria:
> I don't think so. LSP4j can't be responsible for creating threads that are needed even prior to invoking LSP4j.

LSP4J is the API that does create the CompletableFuture instances, so it should be the API responsible of allowing to customize how those CompletableFuture are dispatched.

tomgr:
> > I don't think so. LSP4j can't be responsible for creating threads that are needed even prior to invoking LSP4j.
> 
> LSP4J is the API that does create the CompletableFuture instances, so it should be the API responsible of allowing to customize how those CompletableFuture are dispatched.

I'm not sure about this. LSP4J should be responsible for sending/receiving requests to the language server. The only reason that LSP4J ultimately creates the CompletableFuture instances for LSP4E is because LSP4E takes the result of calling `initialize()` and chains the creation of most requests on top of this. So far as I can see from the documentation of LSP4J, it doesn't promise that calling `thenApplyAsync` or any other method will result in re-using the launcher-supplied LSP4J executor. In fact, that would be directly contrary to the Java Futures API where such methods are documented as using some default executor.

I just want to add that reusing the executor is not enough to fix the concurrency bugs. You also need to be careful about thenApplyAsync versus thenApply prior to the sending of the request.

ahmedneilhussain:
> > I don't think so. LSP4j can't be responsible for creating threads that are needed even prior to invoking LSP4j.
> 
> LSP4J is the API that does create the CompletableFuture instances, so it should be the API responsible of allowing to customize how those CompletableFuture are dispatched.

I think the natural model for LSP4j and indeed LSP, is the javascript threading model - single-threaded, but asynchronous. If one considers that the LSP originated from VSCode, that's not so surprising ;-)

As discussed above, LSP4j doesn't do anything about dispatch by default - that's completely synchronous. LSP4e only achieves async dispatch due to our practice of using `thenComposeAsync` on top of earlier `CompletableFuture`

rubenporras:
I am catching up on this PR today. It looks to me that @martinlippert hits the nail with his comment on https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#implementationConsiderations

```
... The exact order of requests are not really important - as long as the client keeps track of the document state/version for which the request was initiated, cancels requests if additional document edits happen before the request is actually being send or the response arrives, and checks the current document state/version against the one for which the request was initiated when receiving the response - and ignoring the response when it is outdated.

Whether to re-send the request or not in such a case sounds like a difficult decision and probably depends on the individual request.
```

After reading the link, I had the same impression as him, the recommendation is not to impose any ordering of request or responses, but rather keep track of the document state/version and at a minimum ignore responses which no longer apply. 

So far he did not get any response, did he? I think it would be interesting to see how such an approach would work for the formatter, code actions, and other calls that modify the document. I will not have time this week to try it out, but if someone can do a draft, that would be nice.

In addition to that, even if we have a single thread for the requests, since the server can respond out of order, we would still need such a document state/version check, wouldn't we? If that is true, having a single dispatcher thread per LS will not help.

PD: I still need to catch up on the other related PR.

Regards.

rubenporras:
So actually https://github.com/eclipse/lsp4e/pull/174 follows the approach of checking document state/version. I will go through it in detail tomorrow, but so far it looks like the approach that follow the implementation recommendation.

ahmedneilhussain:
> I am catching up on this PR today. It looks to me that @martinlippert hits the nail with his comment on https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#implementationConsiderations
> 
> ```
> ... The exact order of requests are not really important - as long as the client keeps track of the document state/version for which the request was initiated, cancels requests if additional document edits happen before the request is actually being send or the response arrives, and checks the current document state/version against the one for which the request was initiated when receiving the response - and ignoring the response when it is outdated.

Thanks. There are really two issues at play, here:

1. Making sure that the client only sends meaningful requests to the server. That is to say, we only ask for computations that are consistent with the state of the system that we have informed the server of to date.
2. Making sure that if the document state changes while the server computes its response, we are able to detect this and avoid making modifications on the basis of stale data (which would corrupt the document).

We were originally motivated by the second point, specifically for formatting - see the other PR - since it can actually lead to corruption (and we have seen this in the wild). You are right that this PR does not address this problem *as such*. It does, however, address a much deeper problem that makes the other problem much harder to fix.

The way to avoid the corruption with async processing is essentially an optimistic locking approach, as we mention in the other PR. Although I don't talk about it so much in this PR, in the preamble I did show how it can be done with some pseudocode:
1. Read doc version on event thread
2. Dispatch format request
3. In `thenApply` action on format response, check doc version on event thread and apply formatting if unchanged.

Whilst it might indeed by nice to have direct API support for it (such as the `VersionedEdit` I propose in the other PR), the above algorithm will work fine *if this PR is implemented*. Without this PR, the above pseudocode is not safe: say we read the version as `N` and send the format request: it may still be the case that the `documentChanged` event for version `N` has been scheduled but not yet dispatched, and if we are unlucky with thread pool scheduling, our format request will arrive at the server whilst it still only knows about `N - 1`. The other PR demonstrates that it is still possible to avoid corrupting the document, but at the expense of greater complexity: we cannot read the current version ourselves directly prior to dispatch, we need the `documentContentSynchronizer` to keep track of the version as it dispatches the edits.

So no, this PR doesn't directly solve the 'optimistic locking' problem, but it makes addressing the problem much more straightforward.

ahmedneilhussain:
> It feels to me like lsp4e needs to make sure that document change events are sent to the server using a single threaded executor with an unbounded queue to guarantee that those changes are sent to the server in the right order.
Yes, we initially addressed this in https://github.com/eclipse/lsp4e/pull/138 but actually the present PR simplifies things greatly.

> The exact order of requests are not really important - as long as the client keeps track of the document state/version for which the request was initiated, cancels requests if additional document edits happen before the request is actually being send or the response arrives, and checks the current document state/version against the one for which the request was initiated when receiving the response - and ignoring the response when it is outdated.

It's not quite true to say the order of requests does not matter - as you say above obviously it does for document updates! For other things, it is a little more debatable, but the single-threaded approach dispatch approach means avoiding a lot of 'accidental BadLocatation errors' that would otherwise arise from the vagaries of scheduling - which we do see, as verified by @rubenporras 

E.g. at present the following is possible
1. User edits
2. User hovers
3. hover request sent
4. edit request sent
5. server returns error for the hover request since location doesn't exist yet
6. server processes edit

The error in this case is entirely down to the unpredictable client-side dispatch, not because the user has done something that invalidates an earlier operation:
1. User hovers
2. hover request sent
3. user edits
4. stale hover response received
5. edit notification sent
It might be nice for the framework at step 3 to cancel the request in 2 before/as well as scheduling 5, but this would be quite a lot of extra mechanism. I believe there is a proposal for cancel support pending, but I think cancelling requests because the user has invalidated them before the server could respond is a different problem from sending the server invalid requests because the order in which we send them is not deterministic.

ahmedneilhussain:
> After reading the link, I had the same impression as him, the recommendation is not to impose any ordering of request or responses, but rather keep track of the document state/version and at a minimum ignore responses which no longer apply.

Well, the Microsoft documentation is really talking about request/response ordering, it's not really discussing whether or not it is a good idea to send requests in order w.r.t. other requests!

As I discuss above, without having a strong ordering of requests, it's actually hard to even keep track of the document state version, which is the reason I originally needed to change the `documentContentSynchronizer` to take care of it for me in the other PR.

Yes, as we and the Microsoft document discuss, the user making changes between a request being sent and the response being received is something we have to be able to cope with. That's an intrinsic feature of having an asynchronous design with the server 'passively observing' the document with a possible arbitrary delay and no right of veto.

The problem of requests being rejected by the server because they have arrived out of order is entirely self-inflicted, and one we can avoid by being a little more careful about how we schedule dispatch of requests.

Whilst it's still possible with the present design to avoid making e.g. formatting changes that are based on stale data - as we show in the other PR, it's harder than it otherwise would be. And it's completely impossible to avoid the spurious `BadLocation` that both you (@rubenporras ) and we observe.

It's a bit like the way network stacks have to cope with out-of-order or lossy packet transmission: you can layer a whole bunch of coordinating logic over the top in order to give the illusion of reliability, but never completely transparently and at the price of a lot of really complicated logic. What we're arguing in this PR is that the randomness is something we have created for ourselves in LSP4e by some of our implementation patterns, not something imposed on us by LSP4j//LSP, so the better solution is to remove the randomness!

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServerWrapper.java
```diff
@@ -745,6 +759,50 @@ protected IStatus run(IProgressMonitor monitor) {
...
+	 * that it is connected to the document if so.
+	 *
+	 * NB result is a future on this <emph>wrapper</emph> rather than the wrapped language server directly,
+	 * to support accessing the server on the single-threaded dispatch queue.
+	 * @param document Document to connect
+	 * @param filter Constraint on server capabilities
+	 * @return Async result that guarantees the wrapped server will be active and connected to the document. Wraps
+	 * null if the server does not support the requested capabilities or could not be started.
+	 */
+	CompletableFuture<LanguageServerWrapper> connectIf(IDocument document, Predicate<ServerCapabilities> filter) {
```
rubenporras, line 793: could we annotate all arguments and return with @Nullable, @NonNull?

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServerWrapper.java
```diff
@@ -745,6 +759,50 @@ protected IStatus run(IProgressMonitor monitor) {
...
+	 */
+	<T> CompletableFuture<T> executeOnLatestVersion(Function<LanguageServer, ? extends CompletionStage<T>> fn) {
+		return getInitializedServer().thenComposeAsync(fn, this.dispatcher);
+	}
+	/**
+	 * Sends a notification to the LS on a single executor thread tied to this server. Use of this guarantees that the server
+	 * will have seen all previous such requests/notifications (and document updates).
+	 * @param fn LS notification to send
+	 */
+	void notifyOnLatestVersion(Consumer<LanguageServer> fn) {
```
rubenporras, line 778: could we annotate all arguments with @Nullable, @NonNull?

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServerWrapper.java
```diff
@@ -745,6 +759,50 @@ protected IStatus run(IProgressMonitor monitor) {
...
 
+	/**
+	 * Dispatches a LS request on a single executor thread tied to this server. Use of this guarantees that the server
+	 * will have seen all previous such requests (and document updates), and will not receive any further updates
+	 * prior to receiving this one.
+	 * @param <T> LS response type
+	 * @param fn LS method to invoke
+	 * @return Async result
+	 */
+	<T> CompletableFuture<T> executeOnLatestVersion(Function<LanguageServer, ? extends CompletionStage<T>> fn) {
```
rubenporras, line 770: could we annotate all arguments and return with @Nullable, @NonNull?

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServiceAccessor.java
```diff
@@ -628,6 +630,73 @@ protected static LanguageServerDefinition getLSDefinition(@NonNull StreamConnect
...
+	 * <p/>The returned result will complete when all language servers have responded,
+	 * and will return its result in a pool thread to avoid blocking the inbound message readers.
+	 *
+	 * @param <T> Return type of method being called on lang server
+	 * @param document Document on which the request is being made
+	 * @param filter Restriction on capabilities of the language servers we're interested in
+	 * @param fn A single method invocation on <code>LanguageServer</code>
+	 * @return Async result aggregated over all applicable lang servers, filtering out nulls.
+	 */
+	public static <T> CompletableFuture<List<T>> computeOnServers(@NonNull IDocument document, Predicate<ServerCapabilities> filter,
```
rubenporras, line 654: could we annotate all arguments and return with @Nullable, @NonNull?

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServiceAccessor.java
```diff
@@ -628,6 +630,73 @@ protected static LanguageServerDefinition getLSDefinition(@NonNull StreamConnect
...
+	}
+
+	/**
+	 * Accumulator that appends the result of an async computation onto an async aggregate result. Nulls will be excluded.
+	 * @param <T> Result type
+	 * @param result Async aggregate result
+	 * @param next Pending result to include
+	 * @return
+	 */
+	static <T> CompletableFuture<List<T>> combine(CompletableFuture<? extends List<T>> result, CompletableFuture<T> next) {
```
rubenporras, line 680: could we annotate all arguments with @Nullable, @NonNull?

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServiceAccessor.java
```diff
@@ -628,6 +630,73 @@ protected static LanguageServerDefinition getLSDefinition(@NonNull StreamConnect
...
+	}
+
+	/**
+	 * Merges two async sets of results into a single async result
+	 * @param <T> Result type
+	 * @param a First async result
+	 * @param b Second async result
+	 * @return Async combined result
+	 */
+	public static <T> CompletableFuture<List<T>> concatResults(CompletableFuture<List<T>> a, CompletableFuture<List<T>> b) {
```
rubenporras, line 696: could we annotate all arguments and return with @Nullable, @NonNull?

org.eclipse.lsp4e/src/org/eclipse/lsp4e/DocumentContentSynchronizer.java
```diff
@@ -158,10 +155,11 @@ public void documentChanged(DocumentEvent event) {
...
 
 			changeParamsToSend.getTextDocument().setVersion(++version);
-			lastChangeFuture.updateAndGet(f -> f.thenApplyAsync(ls -> {
-				ls.getTextDocumentService().didChange(changeParamsToSend);
-				return ls;
-			}));
+//			lastChangeFuture.updateAndGet(f -> f.thenApplyAsync(ls -> {
+//				ls.getTextDocumentService().didChange(changeParamsToSend);
+//				return ls;
+//			}));
```
rubenporras, line 161: remove commented code?

org.eclipse.lsp4e/src/org/eclipse/lsp4e/DocumentContentSynchronizer.java
```diff
@@ -115,36 +111,37 @@ public DocumentContentSynchronizer(@NonNull LanguageServerWrapper languageServer
...
-		lastChangeFuture = new AtomicReference<>(languageServerWrapper.getInitializedServer().thenApplyAsync(ls -> {
-			this.languageServer = ls;
-			ls.getTextDocumentService().didOpen(new DidOpenTextDocumentParams(textDocument));
-			return ls;
-		}));
+//		lastChangeFuture = new AtomicReference<>(languageServerWrapper.getInitializedServer().thenApplyAsync(ls -> {
+//			this.languageServer = ls;
+//			ls.getTextDocumentService().didOpen(new DidOpenTextDocumentParams(textDocument));
+//			return ls;
+//		}));
```
rubenporras, line 118: remove commented code?

org.eclipse.lsp4e/src/org/eclipse/lsp4e/DocumentContentSynchronizer.java
```diff
@@ -115,36 +111,37 @@ public DocumentContentSynchronizer(@NonNull LanguageServerWrapper languageServer
...
+//			resValueFuture.set(valueFuture);
+//			// We ignore any exceptions that happen when executing the given future
+//			return valueFuture.handle((value, error) -> this.languageServer);
+//		});
+//		return resValueFuture.get();
+//	}
+
+//	CompletableFuture<LanguageServer> lastChangeFuture() {
+//		return lastChangeFuture.get();
+//	}
```
rubenporras, line 144: remove commented code?

org.eclipse.lsp4e/src/org/eclipse/lsp4e/DocumentContentSynchronizer.java
```diff
@@ -304,10 +302,12 @@ public void documentSaved(long timestamp) {
...
 		TextDocumentIdentifier identifier = new TextDocumentIdentifier(fileUri.toString());
 		DidSaveTextDocumentParams params = new DidSaveTextDocumentParams(identifier, document.get());
-		lastChangeFuture.updateAndGet(f -> f.thenApplyAsync(ls -> {
-  			ls.getTextDocumentService().didSave(params);
-  			return ls;
-  		}));
+//		lastChangeFuture.updateAndGet(f -> f.thenApplyAsync(ls -> {
+//  			ls.getTextDocumentService().didSave(params);
+//  			return ls;
+//  		}));
```
rubenporras, line 308: remove commented code?

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServiceAccessor.java
rubenporras: I would name the arguments `first` and `second` as in the javadoc comment.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/DocumentContentSynchronizer.java
```diff
@@ -72,7 +68,7 @@ final class DocumentContentSynchronizer implements IDocumentListener {
 	private int version = 0;
 	private DidChangeTextDocumentParams changeParams;
 	private long modificationStamp;
-	private final AtomicReference<CompletableFuture<LanguageServer>> lastChangeFuture;
+//	private final AtomicReference<CompletableFuture<LanguageServer>> lastChangeFuture;
```
rubenporras, line 71: remove commented code?

rubenporras:
@ahmedneilhussain , I understand now. It makes sense to me. Could you add also the proposed code for the formatting?

rubenporras:
Two tests are failing, it looks like the failures could be related to your changes. Did you look at them?

rubenporras:
@mickaelistria , the change makes sense to me, and the new APIs look actually easier to use than the previous ones (as seen by the adapted code). If the test pass I would be for merging a change like this (maybe we should adapt all consumers in this PR if we agree that it is the way to move forward, to be sure we did not miss something), as it solves a bunch of problems, which are at least a release blocker for our implementation using LSP4E, and both Ahmed and Thomas seem to have put a lot of thought on the topic.

ahmedneilhussain:
> Two tests are failing, it looks like the failures could be related to your changes. Did you look at them?

I haven't even actually *tried* the code! I submitted the PR to have something concrete to discuss about the design we were proposing - if people didn't like the approach or had a better idea then there wasn't any point in polishing it up.

If there's a consensus that this approach will be acceptable then I will do a proper tidy up and make sure the tests work :-)

rubenporras:
Maybe looking at the test would reveal some flaw in the proposal?

ahmedneilhussain:
Hi @rubenporras I think I have addressed all of your 'tidy-up' review comments. I also went so far as to actually try the code which did lead to fixing a logic bug (the code for filtering on server capabilities was broken!). I think this shows that I should probably add some direct tests for the new API - would you agree?

ahmedneilhussain:
@rubenporras I've implemented version-safe formatting as you requested. I've gone for the minimal change, which meant enhancing `DocumentInfo` to allow controlled access to the single threaded executor. I'm not sure whether this is the best change, but it is one that preserves the current semantics (other than concurrency safety!).

Another way of doing it might have been to change LanguageServerWrapper.computeOnServers to take a bifunction of `(LanguageServer, ServerCapabilities)-> CompletableFuture<T>` and then do a `filter(list -> !list.isEmpty()).findFirst().ifPresent(<apply formatting with version check>)` on the overall result. This would change the current semantics, however, since it would actually ask all the applicable servers to compute the formatting and then just pick one result rather than asking the first server that had the formatting capability. It depends how much one agonises over the comment at https://github.com/eclipse/lsp4e/blob/c574d8653a453f4224e4807303f383520d429fc5/org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/format/LSPFormatter.java#L58

More generally, still need to give thoughts as to how to port the existing API/code to access, or at least encourage access of the LS on the dispatcher queue. The multi-language-server use case really complicates the process of exposing an API that is easy to use and safe. `LanguageServerAccessor.computeOnServers` handles a widely-used common case, but doesn't cover everything, as the change to formatting shows!

Similarly with detecting 'optimistic locking' violations - it would be nice to have API support for it. Is the added `LSPEclipseUtils.applyEditsWithVersionCheck` enough?

Obviously we don't want a patchwork of ad-hoc special case methods exposed on the API, but I'm not familiar enough with the whole of the plugin code to be sure what should be a minimal API that captures the required use cases.

I think porting code to the API can be done incrementally, though: it is now possible to get deterministic ordering of requests wrt document updates. Code that still uses the LS directly won't be any worse off than it currently is.

ahmedneilhussain:
> Hi @rubenporras I think I have addressed all of your 'tidy-up' review comments. I also went so far as to actually try the code which did lead to fixing a logic bug (the code for filtering on server capabilities was broken!). I think this shows that I should probably add some direct tests for the new API - would you agree?

Added some tests for LanugageServiceAccessor.computeOnServers. I'll add some for the DocumentInfo methods if we decide they are to be kept.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/DocumentContentSynchronizer.java
```diff
@@ -269,7 +235,7 @@ public void documentAboutToBeSaved() {
 
 
 		try {
-			List<TextEdit> edits = executeOnCurrentVersionAsync(ls -> ls.getTextDocumentService().willSaveWaitUntil(params))
+			List<TextEdit> edits = this.languageServerWrapper.executeOnLatestVersion(ls -> ls.getTextDocumentService().willSaveWaitUntil(params))
```
rubenporras, line 238: remove `this`?

org.eclipse.lsp4e/src/org/eclipse/lsp4e/DocumentContentSynchronizer.java
```diff
@@ -304,10 +270,9 @@ public void documentSaved(long timestamp) {
 		}
 		TextDocumentIdentifier identifier = new TextDocumentIdentifier(fileUri.toString());
 		DidSaveTextDocumentParams params = new DidSaveTextDocumentParams(identifier, document.get());
-		lastChangeFuture.updateAndGet(f -> f.thenApplyAsync(ls -> {
-  			ls.getTextDocumentService().didSave(params);
-  			return ls;
-  		}));
+
+		this.languageServerWrapper.notifyOnLatestVersion(ls -> ls.getTextDocumentService().didSave(params));
```
rubenporras, line 274: remove `this`?

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServerWrapper.java
```diff
@@ -650,9 +664,9 @@ private CompletableFuture<LanguageServer> connect(@NonNull URI uri, IDocument do
 				DocumentContentSynchronizer listener = new DocumentContentSynchronizer(this, theDocument, syncKind);
 				theDocument.addDocumentListener(listener);
 				LanguageServerWrapper.this.connectedDocuments.put(uri, listener);
-				return listener.lastChangeFuture();
+				return CompletableFuture.completedFuture(languageServer);
 			}
-		}).thenApply(theVoid -> languageServer);
+		}, this.dispatcher).thenApply(theVoid -> languageServer);
```
rubenporras, line 669: remove `this`?

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServerWrapper.java
```diff
@@ -745,6 +759,51 @@ protected IStatus run(IProgressMonitor monitor) {
...
+		return getInitializedServer().thenComposeAsync(server -> {
+			if (server != null && (filter == null || filter.test(getServerCapabilities()))) {
+				try {
+					return connect(document);
+				} catch (IOException ex) {
+					LanguageServerPlugin.logError(ex);
+				}
+			}
+			return CompletableFuture.completedFuture(null);
+		}, this.dispatcher).thenApply(server -> server == null ? null : this);
```
rubenporras, line 804: remove `this`?

rubenporras:
> > I don't think so. LSP4j can't be responsible for creating threads that are needed even prior to invoking LSP4j.
> 
> LSP4J is the API that does create the CompletableFuture instances, so it should be the API responsible of allowing to customize how those CompletableFuture are dispatched.

Hi @mickaelistria, you raised some concerns some days ago, @tomgr  and @ahmedneilhussain commented on those but as far as I can see there is no reply. Does it mean that their answers were satisfactory or do you still have the same concerns?

From my point of view, the proposed changes look sensible so I think we could remove the draft status, see if there are any remarks left, and merge, would that be fine with you or do you sill think the issue is still to be solved within LSP4j?

org.eclipse.lsp4e/src/org/eclipse/lsp4e/DocumentContentSynchronizer.java
ahmedneilhussain: Sure. Force of habit from codebases that have the opposite convention :-)

mickaelistria:
I was off for the last 10 days and will go through the discussion ASAP. I still have concerns; not telling that I believe it's wrong, but the complexity of the problem -and its solution- make me "concerned by default".

One first question:

> There are really two issues at play, here:
> 1.  Making sure that the client only sends meaningful requests to the server. That is to say, we only ask for computations that are consistent with the state of the system that we have informed the server of to date.
>  2. Making sure that if the document state changes while the server computes its response, we are able to detect this and avoid making modifications on the basis of stale data (which would corrupt the document).

Does this patch also intend to fix the 1st item? If yes, I would appreciate if there is a separate PR about it which would be only internal as IIRC this doesn't require API (I remember we discuss that earlier). If not, then can you please open a dedicated issue about it?

ahmedneilhussain:
> I was off for the last 10 days and will go through the discussion ASAP. I still have concerns; not telling that I believe it's wrong, but the complexity of the problem -and its solution- make me "concerned by default".
> 
> One first question:
> 
> > There are really two issues at play, here:
> > 
> > 1. Making sure that the client only sends meaningful requests to the server. That is to say, we only ask for computations that are consistent with the state of the system that we have informed the server of to date.
> > 2. Making sure that if the document state changes while the server computes its response, we are able to detect this and avoid making modifications on the basis of stale data (which would corrupt the document).
> 
> Does this patch also intend to fix the 1st item? If yes, I would appreciate if there is a separate PR about it which would be only internal as IIRC this doesn't require API (I remember we discuss that earlier). If not, then can you please open a dedicated issue about it?

This PR is *directly* concerned almost entirely with the first issue. As I discuss with @rubenporras above it is motivated by the second as well, but I originally didn't address the second, just mentioned how it made the second easier to address safely. At Ruben's request, I added actual code rather than pseudocode for the formatting version check problem!

Concerns are good - this sort of change could have wide and unpleasant ramifications if it were flawed! Hopefully if you pick your way through the explanation above it will be OK: this change is intended to be a *simplification* of the way we interact with LSP4j, as we are removing some asynchronous patterns that it turns out we don't need and introduce problems...

mickaelistria:
The LSPEclipseUtils.applyEditsWithVersionCheck gives the impression it covers more the 2nd point (checking sync before processing responses)
This would be sufficient to avoid issues for the case of formatting.

About the 1st one (prioritize documentChanges over other requests and wait for those before sending further requests); 
Imagine a user makes a change, then LSP4E sends new request for hover, for documentSymbols/outline, for documentLink... Those can be sent in parallel without need for one to lock the other. The only thing we can verify is when accessing the `getInitializedLanguageServer`, the last documentChanged notification that was sent matches the current documentTimestamp, and wait in case it doesn't. But I don't see why a new API is needed to achieve that. Maybe there is a very small delay between getting the initialized/up-to-date server and sending the request where the doc can be changed; but there will always be such delay unless we lock the document whenever we make a request; which is clearly something we don't want as user actions that are out of LSP4E should never be blocked. And the delay here is most likely neglect-able compared to the delay between request and response, so putting much effort here would have a very small impact on the general overall problem of "response coming from the past".

And then, sure, document can change when an edit is applied between request being sent and response, but then it's the 2nd topic.

tomgr:
> About the 1st one (prioritize documentChanges over other requests and wait for those before sending further requests); Imagine a user makes a change, then LSP4E sends new request for hover, for documentSymbols/outline, for documentLink... Those can be sent in parallel without need for one to lock the other. 

Note that at the core, the LSP4J dispatch is single-threaded: there is a single stream down which the requests can be sent. All this PR does is to ensure that the requests can be sent down this stream in the correct order. It does not make any request wait for the response of a previous request. So in your example, if LSP4E wants to send a change, a hover, an outline, a documentLink, then those will be written to the stream as fast as it is possible to do so, but in a correct order. It absolutely does not require the server to respond to any of those requests before issuing the next one. In that sense, this PR does not introduce any change in latency versus the current implementation.

> The only thing we can verify is when accessing the `getInitializedLanguageServer`, the last documentChanged notification that was sent matches the current documentTimestamp, and wait in case it doesn't. But I don't see why a new API is needed to achieve that. 

The reason for the new API is that the Java futures API, and specifically `thenApplyAsync`, does not give the properties required to ensure that requests are sent to the server in the correct order. As discussed further up in the thread, such as https://github.com/eclipse/lsp4e/pull/251#issuecomment-1270620189, this isn't fixable within the current API.

> Maybe there is a very small delay between getting the initialized/up-to-date server and sending the request where the doc can be changed; but there will always be such delay unless we lock the document whenever we make a request; which is clearly something we don't want as user actions that are out of LSP4E should never be blocked. 

As discussed above, this PR should not result in any additional delay. But it will make dispatch order correct which is a highly desirable thing! Personally I think this new API is also easier to use, and it encapsulates the fiddly details of composing all the futures together (see the changes to the DocumentColorProvider for example), but of course that's subjective.

mickaelistria:
>  So in your example, if LSP4E wants to send a change, a hover, an outline, a documentLink, then those will be written to the stream as fast as it is possible to do so, but in a correct order.

There is actually no such "correct order" for requests. There is only a correct order when edits or other synchronization events are at play. For requests, we shouldn't try to enforce any order, but just ensure those requests are sent after the last synchronization event is sent.
I don't think any solution that is built on the idea of sequential processing of requests is overall desired according to LSP implementation recommendation.

At this stage, what would be interesting would be the we get in a separate PR the actual functional case we're trying to resolve, to just ensure we agree on the expectations; which seems less certain to me now.

>  But it will make dispatch order correct which is a highly desirable thing!

The only thing desirable is what was mentioned by LSP spec: the synchronization notifications needs to be ordered -but that might be already the case-; and then we want that the request do wait for the last known necessary notification to be sent -which is currently missing-. We do not care about the dispatch order as long as no modification happens.

ahmedneilhussain:
> > So in your example, if LSP4E wants to send a change, a hover, an outline, a documentLink, then those will be written to the stream as fast as it is possible to do so, but in a correct order.
> 
> There is actually no such "correct order" for requests. There is only a correct order when edits or other synchronization events are at play.
That's true for read-only computations wrt one another - so a hover request and a hyperlink request can be sent in either order wrt one another. But it's obviously desirable that they don't get sent before the server has been told about the location they refer to.

> For requests, we shouldn't try to enforce any order, but just ensure those requests are sent after the last synchronization event is sent. I don't think any solution that is built on the idea of sequential processing of requests is overall desired according to LSP implementation recommendation.

Yeah, but as already discussed LSP4j would naturally achieve dispatch of requests in the order they are initiated - it is only LSP4e logic layered over the top that introduces the unpredictability.

Although a hover request and a hyperlink request (for example) don't need to happen in a particular order wrt one another, in practice they both need to be able to wait for the previous document changes they depend on. That can't happen without changing them to use a different API (we did this in the alternative approach to formatting, where we added API onto document content synchroniser to enable this).

If they both need to change to use a different API, or at least to use the API differently, then we are resigned to needing to make intrusive changes to achieve correctness.

It would indeed be possible to make them able to dispatch in either order wrt one another while still wait for any preceding document updates, but it would be more complicated to do this and what would be the benefit? It wouldn't increase throughput or latency, given writing a lambda onto a single thread executor is essentially instant anyway. It wouldn't alter the ability of the server to *process* the requests in either order (that is true anyway).

Yes, making the dispatch sequential is stricter than it needs to be (note that @tomgr did refer to 'a' correct order, not 'the' correct order!), but it's actually a simpler implementation. And having it sequential makes it easier to understand from the point of view of debugging and tracing. It's always better not to have nondeterministic behaviour in code if it's not actually needed, no?

ahmedneilhussain:
> At this stage, what would be interesting would be the we get in a separate PR the actual functional case we're trying to resolve, to just ensure we agree on the expectations; which seems less certain to me now.

Well, it boils down to this, as we say above!

1. Making sure that the client only sends meaningful requests to the server. That is to say, we only ask for computations that are consistent with the state of the system that we have informed the server of to date.
2. Making sure that if the document state changes while the server computes its response, we are able to detect this and avoid making modifications on the basis of stale data (which would corrupt the document).

AIUI your chief concern with this PR is that it is not automatically retrofittable to existing code, and you would prefer changes to the language client that 'silently just worked' to fix up the existing codebase. We don't think that's possible, unfortunately. 

If you can propose a simpler, more elegant change that is transparent to the existing codebase and meets the above two criteria then we'd be delighted to see it and will happily review it :-)

Cheers
Ahmed

ahmedneilhussain:
Added a torture test to show the problem that exists on `master` and discovered this PR didn't quite fix it, so further tweaked the code. Now passes reliably for me!

ahmedneilhussain:
Hi @mickaelistria @rubenporras and others. Do you think this solution can be tidied up and merged? Should I split the formatting changes (including the extra API needed for it) back out?

Thanks,
Ahmed

rubenporras:
Hi, 
from my side, that would be welcome. What do you think @mickaelistria ?
Regards

gerdleon:
Hi @mickaelistria,
Based on your comment on the torture test, I gather that you see the topic as a problem as well. Can we support you in some way?

mickaelistria:
See #318 . I now believe that instead of adding new API to enforce usage of a single thread; just review internal usage of `then*Async` to keep those mininal could do the job. The policy as been to get to `then*Async` almost everytime because it does the job without having much to care about; but this policy has shown its limits: there are block of operations that we do want to enforce sequential executions and for those, plain `then*` does the job.

ahmedneilhussain:
@mickaelistria @rubenporras @gerdleon I've ported the test from https://github.com/eclipse/lsp4e/pull/323 into this branch to show that the dedicated dispatch thread approach doesn't block/interfere with the UI.

My (environment-dependent!) numbers for this test are 0.014s dispatch time, 20.013s test time, 0 invocations on the UI thread.

mickaelistria:
OK, I now agree that using a single threaded executor to dispatch client message to LS is the best path forward and am now short of ideas about anything else that even partially improve things. I believe the overall idea of a dispatcher thread has to be pursued.

However, instead of providing a dedicated entry point in LanguageServerWrapper and more methods in LanguageServerAccesor, I wonder whether we could instead try to have a `LSRequestExecutor.for(languageServer)` utility that would take care of finding the wrapper that matches the LS and then return the associated executor so when using `thenAsync` we could just add the extra `LSRequestExecutor.for(languageServer)` argument when we want to stay in this Language Server thread to run further requests?
That would IMO be simpler to adopt and better emphasize the issue and its solution. What do you think?

ahmedneilhussain:
> OK, I now agree that using a single threaded executor to dispatch client message to LS is the best path forward and am now short of ideas about anything else that even partially improve things. I believe the overall idea of a dispatcher thread has to be pursued.
Thanks, @mickaelistria , welcome to the Dark Side ;-)

> 
> However, instead of providing a dedicated entry point in LanguageServerWrapper and more methods in LanguageServerAccesor, I wonder whether we could instead try to have a `LSRequestExecutor.for(languageServer)` utility that would take care of finding the wrapper that matches the LS and then return the associated executor so when using `thenAsync` we could just add the extra `LSRequestExecutor.for(languageServer)` argument when we want to stay in this Language Server thread to run further requests? That would IMO be simpler to adopt and better emphasize the issue and its solution. What do you think?

When I came up with my solution I did initially toy with `LanguageServerWrapper` exposing a `getDispatchExecutorService()` method along those lines.

However I found that the requirement to support multiple language servers for a given document *really* complicated the act of trying to fix up the dispatch ordering. Hence my rather forlorn question https://github.com/eclipse/lsp4e/pull/174#issuecomment-1255256937 as to whether this was really a must-have!! Not for the first time in software, what is probably the minority use-case really complicates the expected case...

ahmedneilhussain:
On a broader point, @tomgr and I have concerns about exposing too low-level an API, in which the caller has to assemble the bits themselves.

I think that's an excellent idea when the individual components are orthogonal and can be usefully and correctly combined in arbitrary ways - in that case it's the most helpful and general API, in which the user can 'mix and match'.

However, in this case, there are only a handful of correct ways of combining the pieces, and a lot of incorrect ways. In that case, we think a high level 'do it for me' API is more appropriate. We'd rather client code didn't have to deal with the details of when to use async and how to make things go on the dispatch thread, because it's hard to get right. If you see https://github.com/eclipse/lsp4e/pull/251/commits/f0af6d9bb948d13cfe71b2c076b11512912c8d72 you will note that, even having picked through it all quite carefully and had it reviewed by @tomgr in earlier commits, I had still got it wrong!

I think it should be possible to have two or three high-level API methods that handle all the needs of domain code without exposing the low level details, which callers probably would rather not have to worry about. We can then document what the API does and does not guarantee so that callers can use it correctly without caring how those guarantees are achieved.

ahmedneilhussain:
I know one of your concerns about this PR is that it isn't very transparent to calling code, in the sense that calling code will have to change to a different API However if you pick through the reasoning above, you will maybe agree that calling code is inevitably going to be affected if we are to achieve our goal of fixing the concurrency problems. If so, then we should perhaps endeavour to provide an API that is as convenient as possible going forwards.

I'm no doubt biased, but I think the revised API *is* transparent to calling code, in the broader sense that it does more of the work and so the caller's job is simplified, see e.g. org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/hover/LSPTextHover.java 

As I mentioned, having to cater for the possibility of multiple language servers complicates things. From the examples I ported, the core use case is
1. Starting on the event thread, I want to dispatch an operation
2. When processing the result, I will need to handle the possibility of multiple results from multiple servers
3. In the case of `Hover`, I want to aggregate the results and show them all
4. In other cases, I might want just the first one to return a result. This is true for format (and probably everything else that returns an edit, because you don't want to apply multiple overlapping edits!!)

3 is currently implemented by the client code, but as you can see I designed `computeOnServers` to just handle this directly, as it is such a common use-case - if it's a frequent requirement I think it makes sense for the API to handle this out of the box, as the bespoke aggregation in the existing TextHover implementation adds quite a bit of complexity to the code (and is hard to get right) that obscures what the code is really doing.

I didn't implement 4 directly, but if using `Stream::findFirst` was insufficient then `computeOnServers` could probably have an extra overload or two to cater for this.

For example, `Format` is a bit harder to implement with this API, because although it uses a top level predicate to only invoke servers that support formatting, it then invokes format differently on each language server depending on whether or not the server supports section-based or only whole-document formatting.

I didn't do it this way in the example code, but this could be gracefully supported by having an overload of `computeOnServers` that took a function that had *two* arguments, e.g. it would look like

```
LanguageServiceAccessor.computeOnServers(myDoc, canFormat(), (ls, cap) -> {
   if (cap.supportsSectionFormatting) {
     return ls.format(<paragraph>);
  else {
   return ls.format(<whole document>);
 });
```

ahmedneilhussain:
This is the question that I was originally trying to provoke discussion on with this PR (!) - what would be a minimal high-level API that supported the basic use-cases?

I know there are already quite a few methods on `LanguageServiceAccessor` etc, so providing revised implementations using the new threading model on the lines I suggest above, would potentially be a lot of work.

In practice, however, I think many of those methods are quite sparsely used - I have code lenses turned on in my IDE and https://github.com/eclipse/lsp4e/blob/3a3e5341da505bc823f105a39218abd705e3a229/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServiceAccessor.java#L616 is by far the most used. Many of the others are little-used, or turn out to be used mostly in tests!

ahmedneilhussain:
Food for thought, anyway! Thanks for your patience, @mickaelistria  @rubenporras @gerdleon I think we're getting closer to finding a solution we're all happy with.

ahmedneilhussain:
> What I found was that, given the current API, it was very hard to expose a dispatcher such that the calling code could make effective use of it.

PS I ought to mention that I also considered using a single dispatcher thread for the whole plugin, i.e. for all language servers. It didn't really help overcome some of the awkward problems I found in finding the right way to expose an API, and has some fairly unpleasant implications (one rogue language server could potentially block up all others, even on totally unrelated document types/languages).

mickaelistria:
OK, I get that the current API is basically not fixable; and I see the proposal you've implemented.
Do you think it is the ideal? Imagine we could just abandon LanguageServerAccessor and not care about backward compatibility and so on; what do you think would be the minimal yet sufficient set of operations to provide?
I like the idea of providing better API and this case is a great opportunity to do so; so if we can design some better APIs and ditch/deprecate previous ones by the way. This will show a lot of benefits.

So imagine we have a fresh new `LanguageServers` service that is meant as a replacement of LanguageServiceAccessor, what method would be necessary in it in your opinion? Here is what I see as "must-have"

Mainly
* `List<CompletableFuture<T>> LanguageServers.computeAll(Predicate<ServerCapability> capabilityFilter, Function<LanguageServer, CompletableFuture<T>>)` (or `List<LanguageServerExecutor> getAll(Predicate<ServerCapability>)`, see below)

And some utility methods:
* `void LanguageServers.startLanguageServer(ServerDefinition)`
* isStillRunning
* enable/disableLanguageServerContentType
* ...

Some other questions would be
* Do we want to keep the LSDocumentInfo (which seems to be very convenient)? If so, should we remove direct access to the language server and have an `execute(Function<LangaugeServer, CompletableFuture<T>>` method instead?
* Should we expose LanguageSeverWrapper? All of it or only a subset exposed in an interface? For example the `execute<LanugaeServer, CompletableFuture<T>>` mentioned above would seem like the main operation we'd need; so we could have it in a common `LanguageServerExecutor` interface.
* and if we have such interface, couldn't we just access its instances from the `LanguageServers` entry point? So instead of `LanguageServers.executeAll(filter, ls -> ls.doSomething(params))`, it would be `LanguageServers.getLanguageServers(filter).stream().map(executor -> executor.execute(ls -> ls.doSomething(params)))` which is more verbose by narrows the possible entry points.

ahmedneilhussain:
@mickaelistria  I'm going through the current API trying to summarise what the existing methods are and how they are used: that will help provide an overview so we can more easily see what a revised API needs to support (maybe more for the benefit of myself and others: you probably know the majority of the plugin better than anyone else!)

I think we could probably just add methods to LanguageServerAccessor and deprecate the others (some are already deprecated as the API has evolved over time). It might be cleaner to create a new class and deprecate the old one, but something has to e.g. maintain the list of started servers.

One wider point about the API - what is its target audience? Do we expect that the accessor methods are mostly used by people adding functionality to the core LSP4e plugin, or do we need to worry about plugins built on top of LSP4e that use the API for their own stuff?

From what I can see looking at e.g. the corrosion plugin and others, consuming plugins mostly just e.g. subclass `LanguageClientImpl` and deliver the rest of their functionality via some extension points and so on. I guess the whole point of the LSP is that the server provides all the brains and the client side implementation should be very thin!

I also don't know anything about the JDT plugins as we don't use them ourselves - is there much fine-grained dependency on LSP4e in those?

rubenporras:
> I think we could probably just add methods to LanguageServerAccessor and deprecate the others (some are already deprecated as the API has evolved over time). It might be cleaner to create a new class and deprecate the old one, but something has to e.g. maintain the list of started servers.

I also think that is the best course of action. I hope I get some time in the next days myself, but I would nevertheless be worried about doing something bigger than that. Just speaking for myself, I think I lack the experience in the codebase to make big bang breaking changes.

ahmedneilhussain:
Initial fruits of my investigation
[API Summary.pdf](https://github.com/eclipse/lsp4e/files/10155890/API.Summary.pdf)

mickaelistria:
It is definitely a must-have to support multiple LSs per document. Wild Web Developer for example has TypeScript Language Server and ESLint language server working on JS and TS files. I know SonarQube for instance also hooks its diagnostic reports as language server. The MicroProfile Language Server contributes extra analysis and content (hover info, content-assist, code minings...) to .java files, those can already be powered by JDT-LS instead of JDT); same goes to STS language server.
Basically every linter/analyzer/extra integration can be an additional language server on the same file.

> As an Eclipse expert, @mickaelistria , can you explain the reasons between exposing API methods that work in terms of IDocument, IFile and URI?

IFile is a file is part of a project imported in the workspace. IFile integration in the IDE is strong and powerful already, so when we can have an IFile, it's good to use it. But we want LSP4E to work without need for IFile, with files just being directly opened in the IDE without being imported.
The URI is what you expect, a URI ;) The URI can then be used to resolve to an IFile or an IDocument or just a filesystem File...
The IDocument is the object that Eclipse creates when opening a textual editor. It is the working copy of the file. From an editor, we can easily get access to the IDocument; and it is the common object you have in all cases (IFile or external file, or whatever).

> Obviously there won't be an IDocument if the file hasn't been loaded into Eclipse yet, 

You can do File > Open File and you'll get an IDocument for this external file although it's not loaded in the workspace. LSP4E works in such case as well.

> but are there any circumstances in which IFile and URI aren't implicitly interconvertible?

You can have a URI that points to another file which is not part of the Eclipse workspace, for which you won't have an IFile.

> E.g. LanguageServerWrapper has an overload of connect() that takes an IDocument, and another than takes both an IDocument and an IFile.

The IFile one is only meant as a convenience method. From LSP4E perspective, what's needed is mostly the IDocument (to monitor file changes and react). From the IDocument, it should be possible to resolve to an URI and -when available- to an IFile. There are all the necessary methods to do that in LSPEclipseUtils.

rubenporras:
> It is definitely a must-have to support multiple LSs per document. ...  Basically every linter/analyzer/extra integration can be an additional language server on the same file.

Thanks, I understand now. This makes sense (As a side not, I am not sure we handle in resilient and correct way such cases all over the code base, but that would be for another discussion). In that case I think the design should be such that a rouge server does not slow down the other ones.

ahmedneilhussain:
> > It is definitely a must-have to support multiple LSs per document. ...  Basically every linter/analyzer/extra integration can be an additional language server on the same file.
> 
> Thanks, I understand now. This makes sense (As a side not, I am not sure we handle in resilient and correct way such cases all over the code base, but that would be for another discussion). In that case I think the design should be such that a rouge server does not slow down the other ones.

Yeah, thanks @mickaelistria that use case makes a lot more sense now. I struggled to visualise how it might work with multiple language servers all fighting over the same document, as it were, but if in practice it would be one primary language server doing the bulk of the work, and some linters and analysers supplementing this with some extra information, then that feels a lot more intuitive.

I think @rubenporras it will be in practice difficult to prevent language servers that serve the same content types from interfering with one another: most of the API usage gets a set of servers that can satisfy a capability and a content type [usually specified by passing in an `IDocument`] and aggregates the results in one of a handful of ways. In that case you need them all to finish to get the complete result, and if one of them takes a long time, then too bad, you can't get the full result until they all have responded.

However it should be possible to make sure a rogue server doesn't mess things up for unrelated content types.

ahmedneilhussain:
@mickaelistria just a quick question about the project-level methods on `LanguageServiceAccessor`:
https://github.com/eclipse/lsp4e/blob/2c3d0d7ae9096462b96c5eea918e525bafff6ace/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServiceAccessor.java#L552
https://github.com/eclipse/lsp4e/blob/2c3d0d7ae9096462b96c5eea918e525bafff6ace/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServiceAccessor.java#L567

Do I understand the way these work correctly: there is no real concept of e.g. a language server being keyed by a project nature, so if we're looking for the appropriate language server via an `IProject` rather than an `IDocument` then the best we can do is scan through servers that we have thus far instantiated. These methods are thus reliant on a document having been previously opened in the project and caused the language server for that content type to be started. If no such document has been opened, then these methods will return nothing, even if they might have under other circumstances.

Is that right?

rubenporras:
> > > It is definitely a must-have to support multiple LSs per document. ...  Basically every linter/analyzer/extra integration can be an additional language server on the same file.
> > 
> > 
> > Thanks, I understand now. This makes sense (As a side not, I am not sure we handle in resilient and correct way such cases all over the code base, but that would be for another discussion). In that case I think the design should be such that a rouge server does not slow down the other ones.
> 
> Yeah, thanks @mickaelistria that use case makes a lot more sense now. I struggled to visualise how it might work with multiple language servers all fighting over the same document, as it were, but if in practice it would be one primary language server doing the bulk of the work, and some linters and analysers supplementing this with some extra information, then that feels a lot more intuitive.

Actually, after thinking again about it, yes, I understand that it does make sense for a linter, but not in general. As it happens, markers are shown either in marker view which the user can reorder to his liking, or on the sidebar, and in both cases ordering does not matter at all, one can just add all the markers to some list and that is fine. 

But that is not the general case. You cannot have two servers giving you a document outline because those cannot be combined just by processing results as they come, or the same for syntax highlight (you would need to define which one is the main one and how to resolve conflicts), or format (what to do if two LS process the same format requests? you can for sure not send them in parallel, probably one needs to send one, apply result, then ask the second formatter to do its job), or folding (how to fold according to two different criteria?). 

The code in LSP4E assumes overall that one can just process answers to requests of different servers by looping through them, and applying some logic without any particular order or conflict resolution when two servers have a different opinion on something.  Since I doubt this actually works in many cases, and API returning an optional API like

```
	@NonNull
	public static CompletableFuture<Optional<@NonNull LanguageServer>> getLanguageServer(@NonNull IDocument document,
			Predicate<ServerCapabilities> filter) {

```
in addition to

```
	@NonNull
	public static CompletableFuture<List<@NonNull LanguageServer>> getLanguageServers(@NonNull IDocument document,
			Predicate<ServerCapabilities> filter) {

```

would be nice. Then it could be more explicit about if several servers are handled safely or not. I am not sure if that API should just filter all but the first LS or throw an exception. Any opinions?

ahmedneilhussain:
Hi @rubenporras I was just coming to this point: I've opened a new draft PR (hooray, another one!) with a proposed skeleton API https://github.com/eclipse/lsp4e/pull/333

I think I have found out how to handle the case where you just want one response, so that we can just get the result from whichever language server responds first with something non-null.

The method would return a `CompletableFuture<Optional<T>>` - this is slightly different to your example but I propose trying to make as much as possible of the boilerplate for handling multiple results part of the core API, so that the consuming code can focus on the core business logic.

What do you think?

sebthom:
@ahmedneilhussain Is this PR still relevant?

ahmedneilhussain:
> @ahmedneilhussain Is this PR still relevant? I am just wondering because of the title "use a single dispatcher thread per LS". Because this is what we currently have with the SingleThreadExecutor, no?

No @sebthom the code here is completely superseded by that in #344 / #347 (and proposed in draft form in #333 )
This PR was the original draft to show the approach/try out the idea. It is of historic relevance as the place where a lot of the discussion of the concurrency issues with the existing API took place (the comments and discussion are a lot more extensive than the proposed code changes 😆 !)

Closing.

PR #255, opened 2022-10-07
Use singleThreadExecutor for LS requests

mickaelistria:
This should guarantee that the LS receives the requests in the same order as LSP4E code invokes them.

### DIFF
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServerWrapper.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServerWrapper.java
@@ -252,7 +252,7 @@ public class LanguageServerWrapper {
 				LanguageClientImpl client = serverDefinition.createLanguageClient();
 				String theardNameFormat = "LS-" + serverDefinition.id + "-launcher-%d"; //$NON-NLS-1$ //$NON-NLS-2$
 				ExecutorService executorService = Executors
-						.newCachedThreadPool(new ThreadFactoryBuilder().setNameFormat(theardNameFormat).build());
+						.newSingleThreadExecutor(new ThreadFactoryBuilder().setNameFormat(theardNameFormat).build());
 				initParams.setProcessId((int) ProcessHandle.current().pid());
 
 				if (rootURI != null) {


mickaelistria:
cc @rubenporras @ahmedneilhussain @tomgr @martinlippert WDYT? Maybe it doesn't solve all issues about concurrency, but could still solve a lot of them, if not most.

tomgr:
Sorry I think our replies crossed! Just to cross-reference the key point from my other message: my understanding is that this won't propagate the executor across thenComposeAsync, which is used quite a lot around LSP4e for building the requests:

https://github.com/openjdk/jdk/blob/85569e966c22e5b5314a157d05c43c0f6e8c2b28/src/java.base/share/classes/java/util/concurrent/CompletableFuture.java#L2344-L2347

Personally I'm not that keen on a solution that doesn't solve all the issues. My experience of concurrency bugs is that if there is a possibility that it can go wrong, it will go wrong!

mickaelistria:
> my understanding is that this won't propagate the executor across thenComposeAsync

If you want ot use the same executor, then one can already use `thenCompose` instad of `thenComposeAsync`, but that would use the executor and prevent block further parallel executions in case of a singleThreadExecutor, basically reducing the perceived performance.

tomgr:
> > my understanding is that this won't propagate the executor across thenComposeAsync
> 
> If you want ot use the same executor, then one can already use `thenCompose` instad of `thenComposeAsync`, but that would use the executor and prevent block further parallel executions in case of a singleThreadExecutor, basically reducing the perceived performance.

Assuming all the future does is to serialise the JSON request and write it to the pipe to the LS, I don't think parallel execution will help on the request side, as only one request can be written to the pipe anyway, and I would imagine the JSON overhead is negligible. The response side is a different matter: I imagine it would be helpful for the responses to be processed in parallel.

I don't know enough about the Java futures API, but if all calls to `thenComposeAsync` were replaced by `thenCompose`, and a single-threaded executor was used then I think this would fix the concurrency issues. That said, the guarantees are still quite weak here, as CompletableFuture only states:

> Actions supplied for dependent completions of non-async methods **may** be performed by the thread that completes the current CompletableFuture, or by any other caller of a completion method.

I think we need the may to be a must to avoid any races (i.e. we need the thread that completes the current CompletableFuture to execute the action).

`thenComposeAsync` could still be used after the request is sent to LSP4j, but all of the code prior to the sending of the actual request would need to use `thenCompose` otherwise requests could still be overtaken by edits.

As one concrete example, looking at `initiateHoverRequest` I think it would be as simple as using `thenCompose`. But the problem with this is that this then means the response is processed on the single-threaded executor, I think? I imagine this could be fixed by used one of the *Async methods on CompletableFuture?

It seems that whichever approach is taken (this PR, or the other PR), some code changes are going to be required in the requests, and some documentation will be required on how to safely send requests whilst avoiding race conditions.

ahmedneilhussain:
Sorry, @mickaelistria for some reason I didn't get a notification that you had looped me into this PR, so was unaware you had opened it!!

I don't think this change will make any difference to outbound requests. As I discuss in the other PR, AFAICT the thread pool constructed at this point is only used to service *inbound* messages (responses, and any server-initiated requests/notifications).

mickaelistria:
Closing as not useful.

PR #256, opened 2022-10-10
Announce `workspace/configuration` client capability to LS

sebthom:
To receive `workspace/configuration` requests from an LS, the LS client needs to announce the respective capability as of LSP v3.6.0. See  https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#clientCapabilities

image

I encountered this issue while working on a Dart LS client and opened an issue against the Dart SDK https://github.com/dart-lang/sdk/issues/50175 where I got pointed to the spec.


rubenporras:
Hi @sebthom , looks good to me. If you do not mind I would wait till next Monday, just in case someone else has comments on the PR, and if no one commented on the PR, merge it.

PR #257, opened 2022-10-11
add outline icon for EnumMember symbol kind

sebthom:
|Without this PR:|With this PR:
|-|-
|image|image


akurtakov:
Do you have svg file from which you render those? It is nice to keep these around somewhere in case we need different size at later point. I know, rest around around but we have to start at some point.

sebthom:
@akurtakov I don't have the SVG. The images are from here https://github.com/eclipse-jdt/eclipse.jdt.ui/blob/master/org.eclipse.jdt.ui/icons/full/obj16/enum_alt_obj.png https://github.com/eclipse-jdt/eclipse.jdt.ui/blob/master/org.eclipse.jdt.ui/icons/full/obj16/enum_alt_obj%402x.png

akurtakov (approved):
Looks good to me.

Issue #258, opened 2022-10-16
LSP4E does not seem to support multiple editors for the same file, and multiple projects for the same file


treilhes:
The issue exhibited here only happens when the eclipse workspace contains a file accessible from more than one project which is a common case when working on maven multi modules projects (i think gradle multi modules projects are also impacted but i did not check)

Eclipse workspace with 2 projects : parent_project, nested_project
Where nested_project is contained in parent_project
```
parent_project
|
-- nested_project
    |
    -- file.ext

nested_project
|
-- file.ext   
```

With this kind of workspace, we have two cases:

- When the projects presentation is "Hierarchical", the file edited is the most nested and the method LSPEclipseUtils.findResourceFor(uri) will return the correct IResource (provided the function LSPEclipseUtils.findMostNested works properly)
- When the projects presentation is "Flat", the method   is totaly unreliable because more than one valid value can be returned and the most nested is not necessarily the one edited or the only one edited. (Uncomment the test LSPEclipseUtilsTest.testURIToResourceMappingWithNestedProject provided in pull request ?? to see the issue)

So when using the "Flat" project presentation the side effects can affect the following methods:

![Uploading side_effects.png…]()

Do you agree with my assumptions ?

mickaelistria:
> When the projects presentation is "Hierarchical" ... When the projects presentation is "Flat"

Actually the presentation here doesn't matter; as files may be opened from other workflows or actions; and those actions may sometimes open the "not-most-nested" file.
But in any case, fixing #259 would be a good thing and probably fix the issue described here.

treilhes:
In fact, the issue described here will remain after merging #259. 

A test to add in org.eclipse.lsp4e.test.edit.LSPEclipseUtilsTest to exhibit the issue:

```
	@Test
	public void testURIToResourceMappingWithNestedProject() throws CoreException { // like maven nested modules
		IProject project1 = null;
		IProject project2 = null;
		project1 = TestUtils.createProject(getClass().getSimpleName() + System.currentTimeMillis());
		project2 = TestUtils.createNestedProject(project1, project1.getName() + "suffix");
		
		IFile file = project2.getFile("res");
		file.create(new ByteArrayInputStream(new byte[0]), true, new NullProgressMonitor());
		
		IFile sameFile = project1.getFile(project1.getName() + "suffix/res");
		
		Assert.assertTrue(sameFile.exists());
		
[...]

Issue #32, opened 2021-12-17
UI Freeze


guw:
```
eclipse.buildId=4.21.0.I20210906-0500
java.version=11.0.9.1
java.vendor=Azul Systems, Inc.
BootLoader constants: OS=macosx, ARCH=x86_64, WS=cocoa, NL=en_US

org.eclipse.ui.monitoring
Error
Fri Dec 17 10:16:27 CET 2021
UI freeze of 8.3s at 10:16:19.705

...

org.eclipse.ui.monitoring
Info
Fri Dec 17 10:16:27 CET 2021
Sample at 10:16:26.556 (+1.335s)
Thread 'main' tid=1 (TIMED_WAITING)

Stack Trace
	at java.base@11.0.9.1/jdk.internal.misc.Unsafe.park(Native Method)
	at java.base@11.0.9.1/java.util.concurrent.locks.LockSupport.parkNanos(LockSupport.java:234)
	at java.base@11.0.9.1/java.util.concurrent.CompletableFuture$Signaller.block(CompletableFuture.java:1798)
	at java.base@11.0.9.1/java.util.concurrent.ForkJoinPool.managedBlock(ForkJoinPool.java:3128)
	at java.base@11.0.9.1/java.util.concurrent.CompletableFuture.timedGet(CompletableFuture.java:1868)
	at java.base@11.0.9.1/java.util.concurrent.CompletableFuture.get(CompletableFuture.java:2021)
	at org.eclipse.lsp4e.LanguageServerWrapper.supportsWorkspaceFolderCapability(LanguageServerWrapper.java:547)
	at org.eclipse.lsp4e.LanguageServerWrapper.canOperate(LanguageServerWrapper.java:536)
	at org.eclipse.lsp4e.LanguageServerWrapper.canOperate(LanguageServerWrapper.java:872)
	at org.eclipse.lsp4e.LanguageServiceAccessor.lambda$6(LanguageServiceAccessor.java:361)
	at org.eclipse.lsp4e.LanguageServiceAccessor$$Lambda$1631/0x0000000100178c40.test(Unknown Source)
	at java.base@11.0.9.1/java.util.stream.ReferencePipeline$2$1.accept(ReferencePipeline.java:176)
	at java.base@11.0.9.1/java.util.stream.ReferencePipeline$2$1.accept(ReferencePipeline.java:177)
	at java.base@11.0.9.1/java.util.HashMap$KeySpliterator.forEachRemaining(HashMap.java:1603)
[...]

PR #259, opened 2022-10-16
LSPEclipseUtils.findMostNested now returns the most nested project

treilhes:
Hello,

I will try to be the most precise possible.

The issue exhibited here only happens when the eclipse workspace contains a file accessible from more than one project which is a common case when working on maven multi modules projects (i think gradle multi modules projects are also impacted but i did not check)

Eclipse workspace with 2 projects : parent_project, nested_project
Where nested_project is contained in parent_project
```
parent_project
|
-- nested_project
    |
    -- file.ext

nested_project
|
-- file.ext   
```

With this kind of workspace, we have two cases:

- When the projects presentation is "Hierarchical", the file edited is the most nested and the method LSPEclipseUtils.findResourceFor(uri) will return the correct IResource (provided the function LSPEclipseUtils.findMostNested works properly)
- When the projects presentation is "Flat", the method   is totaly unreliable because more than one valid value can be returned and the most nested is not necessarily the one edited or the only one edited. (the test LSPEclipseUtilsTest.testURIToResourceMappingWithNestedProject provided in pull request to see the issue)

The bug in LSPEclipseUtils.findMostNested was realy simple like you can see in the code.
The function was returning the first processed IFile instead of the most nested which may leads to return an incorrect IResource from LSPEclipseUtils.findResourceFor(uri)

I think this pull request solves the issues #32, #112

Why?

Because while editing a file and temporary errors pop up then a non empty diagnostic is returned by the language server.
Provided we have a multi modules workspace like described earlier then:

- the call to LSPEclipseUtils.findResourceFor(uri) in LSPDiagnosticsToMarkers.accept(PublishDiagnosticsParams diagnostics) may return the wrong IResource
- then in LSPDiagnosticsToMarkers.updateMarkers(PublishDiagnosticsParams diagnostics, IResource resource) the disconnected boolean will be set to true because the IResource provided is not the one edited :

```
// A language server can scan the whole project and generate diagnostics for files that are not currently open in the IDE
// (the markers will show up in the problem view). If so, need to open the document temporarily but be sure to release it
// when we're done
final boolean disconnect = !diagnostics.getDiagnostics().isEmpty() && LSPEclipseUtils.getExistingDocument(resource) == null;
```
- if  only one file is connected to the language server, the server will be shut down and subsequent calls will return:
java.io.IOException: Stream closed

This pull request may solve issues related to multi modules workspace but only for users using the "Hierarchycal" project presentation. For those using the "Flat" project presentation, i created the issue #258 but i think it won't be a trivial one as LSP4E does not seem to currently support multiple editors for the same file opened from multiple projects.

### DIFF
--- a/org.eclipse.lsp4e.test/src/org/eclipse/lsp4e/test/TestUtils.java
+++ b/org.eclipse.lsp4e.test/src/org/eclipse/lsp4e/test/TestUtils.java
@@ -26,7 +26,9 @@ import java.util.concurrent.CountDownLatch;
 import java.util.function.BooleanSupplier;
 
 import org.eclipse.core.resources.IFile;
+import org.eclipse.core.resources.IFolder;
 import org.eclipse.core.resources.IProject;
+import org.eclipse.core.resources.IProjectDescription;
 import org.eclipse.core.resources.ResourcesPlugin;
 import org.eclipse.core.runtime.CoreException;
 import org.eclipse.core.runtime.NullProgressMonitor;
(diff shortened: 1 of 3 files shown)


mickaelistria:
1 test is failing, it looks like your proposed fix is incomplete.

mickaelistria:
Thanks!

PR #268, opened 2022-10-24
Minor code improvements

sebthom:


mickaelistria:
Thank you!

rubenporras:
out of curiosity, do we have a coding convention when to use var/final?

sebthom:
I am not aware of any. I am currently using the var keyword only when an object is instantiated using the new keyword so it is immediately visible what type the variable has.

mickaelistria:
No, there is no coding convention as long as code is readable.

PR #269, opened 2022-10-24
Ensure all docs are up-to-date before returning a pointer to the LS

mickaelistria:


org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServerWrapper.java
```diff
@@ -716,10 +716,16 @@ public LanguageServer getServer() {
...
 
+	private CompletableFuture<?> allDocumentsAreUpToDate() {
+		return CompletableFuture.allOf(connectedDocuments.values().stream().map(DocumentContentSynchronizer::lastChangeFuture).toArray(CompletableFuture<?>[]::new));
+	}
+
 	/**
-	 * Starts the language server and returns a CompletableFuture waiting for the
-	 * server to be initialized. If done in the UI stream, a job will be created
-	 * displaying that the server is being initialized
+	 * Starts the language serverm, ensure it's and returns a CompletableFuture waiting for the
```
sebthom, line 724: typo `serverm`

ahmedneilhussain:
> Is this an alternative proposal to #251? Maybe @ahmedneilhussain would be interested in it.

Hi @mickaelistria , is that indeed what you are intending with this PR - an alternative to #251 ?

org.eclipse.lsp4e/src/org/eclipse/lsp4e/DocumentContentSynchronizer.java
```diff
@@ -71,7 +72,7 @@ final class DocumentContentSynchronizer implements IDocumentListener {
 
 	private int version = 0;
 	private DidChangeTextDocumentParams changeParams;
-	private long modificationStamp;
```
rubenporras, line 74: not very important, but why do we modify the name of this field? It looks like it is still the modificationTimestamp, isn't it?

rubenporras (commented):
I understand that this would be an alternative solution to #251 for ordering requests/notification sent to the server. As in #251, this change also orders more than what is needed (we will wait for all document modifications, not only for modifications of the document we are going to send a new request/notification), but I do not think that there is any harm into it, as sending the modifications should be blazingly fast (for the record, I think that having also one thread should be also as fast for practical purposes).

It looks good to me. @ahmedneilhussain, do you think as well that it would solve the problem for ordering request/notification wrt. document modifications?

If so, if would be nice to merge it before the next LSP4E release.

tomgr:
I'm afraid this doesn't fix the concurrency issue as it still allows an edit to overtake a request. For example, suppose the editor does the following sequence of events: `documentEdit; hover; documentEdit`. Then this change prohibits the `hover` being moved before the first `documentEdit` (which is good), but it does not prevent the second `documentEdit` moving before the `hover`. This means the server can still execute the command on the wrong version of the document.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/DocumentContentSynchronizer.java
mickaelistria: modify can mean 2 things: if I modify a document with `document.set(...)`, then it doesn't change themodificationStamp in this class - the stamp is only read on open and save. Changing the name is just for disambiguation to clarify that it's not the modification of the document content, but really the "storage" stamp.

mickaelistria:
> . Then this change prohibits the hover being moved before the first documentEdit (which is good), but it does not prevent the second documentEdit moving before the hover. This means the server can still execute the command on the wrong version of the document

It is exactly what is intended: when an edit has happened (eg users has typed something), then all requests that are not sent should wait for synchronization with the latest edit because otherwise, they would run against an older document version and there result would be outdated.
So a `edit1;hover//edit2` (hover and edit happening in parallel, hover because of some background action such as mouse event and edit because of user typing int he doc) will lead to a good hover result if we send it in the order `edit1;edit2;hover` because hover will have happened against the current version of the document as it appears in user world; while `edit1;hover;edit2` will lead to a hover that is likely not relevant anymore after `edit2`.

mickaelistria:
Damn, I merged it by mistake as I wanted to first process feedback from @sebthom about the typo.
But now it's here, I would like to leave it as it and see how it behaves. More particularly, if @tomgr you can give it a try with next snapshot and tell whether you see it causing some regressions, it would be welcome and we'll revert it.

rubenporras:
Hi @mickaelistria , @ahmedneilhussain , I had no time to test it, and will be able to test this week either, but I have been thinking on you solution and the comments we had on the different PRs from different people.

I think it is true that the ordering enforce by this PR is good enough (taken strictly), but I do not think the following part of your last comment is correct:

>  if we send it in the order edit1;edit2;hover because hover will have happened against the current version of the document as it appears in user world

the current version of the document could have less content, so there is good chances the server will run into a `BadLocationException`, or the region might have shifted a lot, so it might point to a different element, so the result will be very confusing.

Still, I believe we could make it work if `LSPEclipseUtils.toTextDocumentPositionParamsCommon(T, int, IDocument)` would use [VersionedTextDocumentIdentifier](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#versionedTextDocumentIdentifier) instead of just https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocumentIdentifier. 

Then the server could check the version of the document in the request and the version that it has in his world, if the versions do not match, should error the request with `ContentModified`. That is my understanding of the LSP spec. I am quite sure the Xtext LS does not handle such cases yet (agree, @szarnekow?), so it migth rather run into a `BadLocationException` or show misleading results. Then LSP4E should do as in the spec

>  If clients receive a ContentModified error, it generally should not show it in the UI for the end-user. Clients can resend the request if they know how to do so. It should be noted that for all position based requests it might be especially hard for clients to re-craft a request.

In summary, I think we could say this PR partially solve the problem, but we need to deal with the fallout, that is, on LSP4E:

1. use versioned text document identifiers
2. handle ContentModified

On the server side (I think we should aim for LSP4E and Xtext to play nice):

1. Check the versioned document identifier

Does it makes sense what I wrote? If so I think we should start working on it. I would not wait until some users have exceptions or weird results. This looks to me like the kind of code that works fine for many people but could any time pop up and make the experience of some developers very bad.

Regards

tomgr:
> I think it is true that the ordering enforce by this PR is good enough (taken strictly), but I do not think the following part of your last comment is correct:
> 
>> if we send it in the order edit1;edit2;hover because hover will have happened against the current version of the document as it appears in user world
> 
> the current version of the document could have less content, so there is good chances the server will run into a BadLocationException, or the region might have shifted a lot, so it might point to a different element, so the result will be very confusing.

This is indeed the problem with the approach of this PR. To give an example of a more serious version of the problem, suppose the second hover request is actually `linkedEditingRange` or `prepareRename`. Then the server would compute the ranges on a different version of the document. This will result in LSP4E applying the wrong edits to the wrong locations in the user's document; e.g. a rename would not precisely edit the renamed identifier, but instead other characters.

> Still, I believe we could make it work if `LSPEclipseUtils.toTextDocumentPositionParamsCommon(T, int, IDocument)` would use [VersionedTextDocumentIdentifier](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#versionedTextDocumentIdentifier) instead of just https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocumentIdentifier.

Unless I've misunderstood I don't think this is allowed according to the LSP spec. If you look at the spec, relatively few places mention VersionedTextDocumentIdentifier. For example, `HoverParams` does not permit a `VersionedTextDocumentIdentifier`. Even if it was permitted, I'm not sure this is the right solution. Given a choice, it would be much better to only ever generate valid requests.

> This looks to me like the kind of code that works fine for many people but could any time pop up and make the experience of some developers very bad.

I completely agree with this.

@ahmedneilhussain is currently working on a stress test that should hopefully us to easily check if a potential fix really does fix the problem. Hopefully this will help resolve this issue throughly.

mickaelistria:
> the current version of the document could have less content, so there is good chances the server will run into a BadLocationException, or the region might have shifted a lot, so it might point to a different element, so the result will be very confusing.

Indeed, this doesn't prevent the server response for being outdated, as it doesn't prevent the user from editing while server is computing response. But actually, that part is not related to request ordering or even synchronization, it's more related to validating the response from the LS is still current according to the document version it's applied on (and reacting accordingly if not).
That topic is indeed not at all in the scope of this commit.

ahmedneilhussain:
I've added a test to show that this fix doesn't quite work, see https://github.com/eclipse/lsp4e/pull/295

rubenporras:
Hi @mickaelistria ,

> Indeed, this doesn't prevent the server response for being outdated, as it doesn't prevent the user from editing while server is computing response. But actually, that part is not related to request ordering or even synchronization, it's more related to validating the response from the LS is still current according to the document version it's applied on (and reacting accordingly if not)

Fine for me, I also do not think that preventing the server response for being outdated or preventing the user from editing while the server is computing a response is related to this commit. It is totally fine that no guarantees are done in this commit because we had none so far. We only need to keep in mind that we need to solve the problem later to avoid cornering ourselves, that is all. So yes, let us put it on the side the discussion regarding this commit (or https://github.com/eclipse/lsp4e/pull/251).

The topic that concerns me (I am not sure by now that all participants on the discussion are on the same page) is that the server must be able to process messages properly, and by extension to deliver correct responses, that is, the server should not: 1. throw exceptions, and 2. answer with wrong content. If responses are outdated, that is fine for now as I wrote in the previous paragraph. 

Do you agree that LSP4E must implement a pattern that allows the server to process messages properly?

My problem is that I believe that for the sever to process messages properly, the messages must be delivered to it in the same order as the events that trigger those happen on the client. This is the consequence of most messages just having a TextDocumentIdentifier (that has only an URI, no document version, as noted by @tomgr) as data, they have an implicit knowledge of the current state of the document in the server. I have tried to find an alternative solution, but I cannot find it. If messages have no document version, the server cannot know they are out of order, so it will throw BadLocationExceptions or answer with wrong content, and that is not going to work out.

Do you agree that if messages are out of order, the server cannot process them properly? If you disagree, how do you think the server should handle such messages out of order?

If we do not understand how the server should handle messages out of order , and as by https://github.com/eclipse/lsp4e/pull/295 we know that out of order delivering happens, I am in favor of reverting this commit and merge a variant of https://github.com/eclipse/lsp4e/pull/251, that one makes sure messages are sent in order. If the mayor problem with that PR is that it addresses several issues at once, I think we could trim it down to just do message ordering. Would that be a way forward for you? If we find something better, that is find as well, I am not attached to it :)

What do you think?

Regards

mickaelistria:
> Do you agree that LSP4E must implement a pattern that allows the server to process messages properly?
> Do you agree that if messages are out of order, the server cannot process them properly? If you disagree, how do you think the server should handle such messages out of order?

I do agree.
So do you think this patch breaks it more than previous state and thus should be reverted?

rubenporras:
Hi Mickael,

I think it will show a different set of errors because messages are reordered in a way which which still create issues. If the new issues are better/worse or less/more that before is hard to know. Thus I would revert because at least the old state was known for a longer time. That is just a feeling so no hard preference. For me it is more important that we look for the way forward after that, be it a variant of #251 or something else.

Regards

gerdleon:
Hi @mickaelistria ,
in our in-house testing we observe the out-of-order handling in LSP4E to be a major problem for our products.
Is the main requirement for a fix that the API does not change? If that is true, we think it is not really possible, do you have suggestion we could explore together?

Regards

mickaelistria:
> in our in-house testing we observe the out-of-order handling in LSP4E to be a major problem for our products.

Is this change causing or increasing the problems you see? Do you think it would be better to revert it?

ahmedneilhussain:
> Hi @mickaelistria , in our in-house testing we observe the out-of-order handling in LSP4E to be a major problem for our products. Is the main requirement for a fix that the API does not change? If that is true, we think it is not really possible, do you have suggestion we could explore together?
> 
> Regards

Hi @gerdleon maybe you could have a look at the proposal and discussion in https://github.com/eclipse/lsp4e/pull/251 ? We also came to the conclusion that a fix without API changes was not possible, so it then becomes a discussion about how to change the API so that we have a correct and reasonably elegant fix without being too disruptive.

gerdleon:
Hi,
@mickaelistria we are currently on an older state of LSP4E and did not test with the newest, yet (the products I own still need to use Java 11). What I can say for sure: the out of order sending of requests is causing problems (exceptions, random weird user behavior).  
But honestly, I cannot judge the impact of this change on the products I own because it was not tested.

Given that there are doubts about the correctness of the commit in this pull request, I agree with @ahmedneilhussain and would vote for looking again at #251. From what I understand it follows the implementation pattern used in VS Code which is the reference implementation and solves the problem fundamentally.
AFAIK the LSP specification is not explicit but I think following up on #251 will give us (and others building on top of LSP4E) a good foundation to built on.

PR #270, opened 2022-10-25
Use TestUtils.waitForAndAssertCondition instead of DisplayHelper

sebthom:
This PR increases the comprehensibility of test cases by using `TestUtils.waitForAndAssertCondition` instead of `new DisplayHelper() { ... }.waitForCondition` + `Assert.assertTrue`  which needs a lot of boilerplate.

While doing the changes I already identified a test case which is not really working `LanguageServiceAccessorTest#testReuseMultirootFolderLSAfterInitialProjectGotDeleted`. I added TODO/FIXME tags accordingly since I don't know how to fix it.


org.eclipse.lsp4e.test/src/org/eclipse/lsp4e/test/TestUtils.java
```diff
@@ -233,17 +237,66 @@ public static IEditorReference[] getEditors() {
...
+		assertTrue("Condition not met within expected time.",
+				waitForCondition(timeout_ms, display, condition));
+	}
+
+	public static void waitForAndAssertCondition(String errorMessage, int timeout_ms, Condition condition) {
+		assertTrue(errorMessage, waitForCondition(timeout_ms, condition));
+	}
+
+	public static void waitForAndAssertCondition(String errorMessage, int timeout_ms, Display display, Condition condition) {
+		var ex = new Exception[1];
```
rubenporras, line 255: why do we use an array with one position instead of just a simple variable?

org.eclipse.lsp4e.test/src/org/eclipse/lsp4e/test/TestUtils.java
```diff
@@ -233,17 +237,66 @@ public static IEditorReference[] getEditors() {
...
+		var isConditionMet = new DisplayHelper() {
```
rubenporras, line 256: this is just my taste, but I would use in general `val` for things which can be final.

org.eclipse.lsp4e.test/src/org/eclipse/lsp4e/test/TestUtils.java
```diff
@@ -233,17 +237,66 @@ public static IEditorReference[] getEditors() {
...
+	}
+
+	public static boolean waitForCondition(int timeout_ms, Condition condition) {
 		return waitForCondition(timeout_ms, PlatformUI.getWorkbench().getDisplay(), condition);
 	}
 
-	public static boolean waitForCondition(int timeout_ms, Display display, BooleanSupplier condition) {
-		return new DisplayHelper() {
+	public static boolean waitForCondition(int timeout_ms, Display display, Condition condition) {
+		var ex = new Exception[1];
```
rubenporras, line 281: same remark as above, why an array?

org.eclipse.lsp4e.test/src/org/eclipse/lsp4e/test/LanguageServiceAccessorTest.java
```diff
@@ -185,7 +187,7 @@ public void testGetOnlyRunningLanguageServers() throws Exception {
 		editor1 = openEditor(testFile1);
 		assertNotEmpty(getInitializedLanguageServers(testFile1, MATCH_ALL));
 
-		waitForCondition(5_000, () -> getActiveLanguageServers(MATCH_ALL).size() == 1);
+		waitForCondition(5_000, () -> getActiveLanguageServers(MATCH_ALL).size() > 0);
```
rubenporras, line 190: why this change?

org.eclipse.lsp4e.test/src/org/eclipse/lsp4e/test/TestUtils.java
sebthom: there is no val, this isn't scala/kotlin :-)

org.eclipse.lsp4e.tests.mock/src/org/eclipse/lsp4e/tests/mock/MockLanguageServerMultiRootFolders.java
```diff
@@ -64,7 +64,7 @@ public final class MockLanguageServerMultiRootFolders implements LanguageServer
 	private MockWorkspaceService workspaceService = new MockWorkspaceService(this::buildMaybeDelayedFuture);
 	private InitializeResult initializeResult = new InitializeResult();
 	private long delay = 0;
-	private boolean started;
+	private volatile boolean started;
```
rubenporras, line 67: why this change?

org.eclipse.lsp4e.test/src/org/eclipse/lsp4e/test/TestUtils.java
rubenporras: alright :-)

org.eclipse.lsp4e.test/src/org/eclipse/lsp4e/test/TestUtils.java
sebthom: this stems from the fact that inner anonymous/lambdas instances can only access final local fields but we need to write to a field.

org.eclipse.lsp4e.tests.mock/src/org/eclipse/lsp4e/tests/mock/MockLanguageServerMultiRootFolders.java
sebthom: because we have code that runs in other threads which queries the isStarted() method. without the volatile it is not guaranteed that these threads see the current value.

rubenporras (approved):

rubenporras:
@sebthom , shall I merge?

sebthom:
> @sebthom , shall I merge?

Sure, go for it. Thanks.

PR #272, opened 2022-10-28
LSP folding with ITextViewerLifecycle

angelozerr:
LSP folding with ITextViewerLifecycle


angelozerr:
This PR requires https://github.com/eclipse-platform/eclipse.platform.text/pull/107 to benefit with ITextViewerLifecycle

The benefit with this PR are:

 * the code is simplier
 * it doesn't create a new recolcier (a new Thread), it uses now the main Reconciler of the generic editor.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/folding/LSPFoldingReconcilingStrategy.java
```diff
@@ -320,4 +306,24 @@ public void setProgressMonitor(IProgressMonitor monitor) {
...
+
+	@Override
+	public void uninstall() {
+		setDocument(null);
+		if (viewer != null) {
+			viewer.removeProjectionListener(this);
+			viewer = null;
+		}
+		projectionDisabled();
+	}
```
rubenporras, line 328: could we move this block back to 178-194 so that the diff looks nicer?

rubenporras:
If the API gets merged upstream, this looks like a nice change.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/folding/LSPFoldingReconcilingStrategy.java
angelozerr: fixed

angelozerr:
> If the API gets merged upstream, this looks like a nice change.

It was merged.

rubenporras:
Thinking about it again, merging is not enough, right? We should wait for the API to be part a released `eclipse.platform.text` or at least an M1 version. Or would it be possible to release a newer `eclipse.platform.text` out of the schedule for [SimRel 2023-03](https://wiki.eclipse.org/Category:SimRel-2023-03)?, maybe @mickaelistria knows that? I am waiting for a release that contains https://github.com/eclipse-platform/eclipse.platform.text/pull/118, so I would be very happy if that would be possible.

mickaelistria:
> Thinking about it again, merging is not enough, right? We should wait for the API to be part a released eclipse.platform.text or at least an M1 version.

Yes.

> Or would it be possible to release a newer eclipse.platform.text out of the schedule for [SimRel 2023-03](https://wiki.eclipse.org/Category:SimRel-2023-03)?, maybe @mickaelistria knows that? I

Is it not possible at the moment and trying to change the existing Platform release processes to allow that would be a major challenge.

angelozerr:
@mickaelistria @rubenporras can we merge this PR?

It will reduce the number of Thread per file.

mickaelistria:
> can we merge this PR?

Not as long as it doesn't build.

rubenporras:
Hi @angelozerr , we could add a dependency in the Manifest to the new platform and merge this PR now, right? Do you plan to do that?

Note that we added a new Semantic Reconciler, so I think you would need to rebase and adapt that new one as well.

Regards

angelozerr:
> Do you plan to do that?

done.

> Note that we added a new Semantic Reconciler, so I think you would need to rebase and adapt that new one as well.

done.

mickaelistria:
I supported this change a while ago, before the API was in Platform. So I'm all in favor of merging now.

Issue #250, opened 2022-09-27
Support detail information in Outline View


gerdleon:
The LSP protocol allows the optional string "detail" in the "[DocumentSymbol](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#documentSymbol)".

When testing with the Outline View detail information was not shown. Only Symbol.kind and Symbol.name are shown. 

We could create a patchset to implement the support - unless someone else is already working on it. 
What do you think?

mickaelistria:
Such a contribution would be welcome!

gerdleon:
Hi Mickael,
Is it ok to remove the textual showing of the SymbolKind? It is already indicated by the icon and, I think, it would clash with the showing of the details. 
I cross-checked the outline in VS Code and it does not have an option to show the SymbolKind as text.

mickaelistria:
Yes, I believe it would be fine.

PR #275, opened 2022-11-01
Show the detail information from a DocumentSymbol

gerdleon:
In this change:
* the detail information provided in the DocumentSymbol response is shown in the outline view.
* the textual representation of the SymbolKind was removed

Addresses: #250 

Signed-off-by: gerdleon gerd.grossmann@avaloq.com


mickaelistria:
I think we should keep the "Show Kind" preference, as IIRC it was requested by some people in the past; but maybe just turn it off by default.

gerdleon:
Hi,
on the UI this will clash, right? 

Do you remember why it was requested? Is it the missing icons for some SymbolKinds? Then this could be solved more efficiently (add missing icons for example).

For the release notes the change in behavior could be explicitly mentioned.

To make sure I do not implement something silly, I cross-checked the behavior with VS Code and the LSP based Java support (provided by RedHat).
The response from the LS is:
`Result: [
    {
        "name": "myApp",
        "kind": 4,
        "range": {
            "start": {
                "line": 0,
                "character": 0
            },
            "end": {
                "line": 0,
                "character": 14
            }
        },
        "selectionRange": {
            "start": {
                "line": 0,
                "character": 0
            },
            "end": {
                "line": 0,
                "character": 14
            }
        },
        "detail": ""
    },
    {
        "name": "MyApp",
        "kind": 5,
        "range": {
            "start": {
                "line": 2,
                "character": 0
            },
            "end": {
                "line": 14,
                "character": 1
            }
        },
        "selectionRange": {
            "start": {
                "line": 2,
                "character": 13
            },
            "end": {
                "line": 2,
                "character": 18
            }
        },
        "detail": "",
        "children": [
            {
                "name": "main(String[])",
                "kind": 6,
                "range": {
                    "start": {
                        "line": 4,
                        "character": 4
                    },
                    "end": {
                        "line": 6,
                        "character": 5
                    }
                },
                "selectionRange": {
                    "start": {
                        "line": 4,
                        "character": 23
                    },
                    "end": {
                        "line": 4,
                        "character": 27
                    }
                },
                "detail": " : void",
                "children": []
            },
            {
                "name": "squareNumber(int)",
                "kind": 6,
                "range": {
                    "start": {
                        "line": 8,
                        "character": 4
                    },
                    "end": {
                        "line": 10,
                        "character": 5
                    }
                },
                "selectionRange": {
                    "start": {
                        "line": 8,
                        "character": 15
                    },
                    "end": {
                        "line": 8,
                        "character": 27
                    }
                },
                "detail": " : int",
                "children": []
            }
        ]
    }
]`

And the matching outline for this response is:
VSCodeOutline

mickaelistria:
> on the UI this will clash, right?

We'll see... and people who need it will provide feedback. Maybe we'll remove it fully later, but now we don't have enough feedback to decide whether it's safe to do it or not for most adopters. Disabling it by default will allow to get such feedback.

> Is it the missing icons for some SymbolKinds? Then this could be solved more efficiently (add missing icons for example).

I believ some people don't understand the icons much. In any case, I don't think we should fully remove it until it's solved.

The proposed implementation seems correct and consistent with VSCode.

gerdleon:
Hi,
could we be more "aggressive"? I think that using the behavior of VS Code as a template is reasonable - it is being used by millions of engineers.

Two questions:
1) How would you like to have the config option to show the SymbolKind? As a system property is simple to do.
2) If SymbolKind and Details are shown, how would you want to have it visualized in the Outline View?

mickaelistria:
> could we be more "aggressive"?

We should avoid it. Being aggressive means breaking some potential consumers of LSP4E.
Moreover, keeping the SHOW_KIND stuff as it is and "just" adding the description will result in a simpler, more focused change. Better for history tracking. 

> How would you like to have the config option to show the SymbolKind? As a system property is simple to do.

We already have this on current master; just keep it as it and focus only on adding the description.

> If SymbolKind and Details are shown, how would you want to have it visualized in the Outline View?

I don't really care, just concatenating the strings could be enough. People with an opinion will tell when they see it in action.

gerdleon:
I will prepare another commit with the suggestions if you want - I can follow the reasoning for a smaller more concise commit.

My reasoning for removing the textual representation of the SymbolKind at the same time were (and I think it is sound):
* _The main reason:_ showing the SymbolKind as text (as is) and additionally the detail information (for example concatenated with the SymbolKind) will result in a bad UI experience, unless there is investment into finding out how this should be done. TBH I can imagine some options that might make sense (e.g. replace the icons and show the text instead) but to get this right is also not quick.
* I checked the history and I saw that at least one bug was opened to hide this text (https://github.com/eclipse/lsp4e/commit/1d5509a7acaf01bf76193f73e3ef9f5500a051bb). Also to show the Symbolkind as text was part of the initial contribution 6 years ago, there I strongly suspect that it was because the icons for the SymbolKinds were missing.
* I observed unpredictable/buggy behavior with the ShowKind option. It is enabled by default but sometimes only shows after de-selecting and then selecting the option again. Since I believe the feature to be not very useful, I decided to remove it, instead of trying to fix this as well. 
* To me, it makes sense to follow an established example. E.g., neither VS Code nor Eclipse JDT Outline Views do have an option to display the text "Class" for the Symbol representing a "Class".
* IMO (but still very unsure), to display the SymbolKind as text seems to go against the spirit of LSP where the symbolkinds should transfer a common meaning/metapher - we should rather strive for meaningful icons

rubenporras:
Hi @mickaelistria , 

if it is about keeping the git history clean, we are for sure happy to have a PR to remove the symbol kind as text, then a second one to add the details (or vice versa).

Would that be fine for you?
 
If not, what about show the details if the server provides them (that is, non an empty string) instead of the symbol kind plus a plug in property to not show the symbol kind at all?

I know concatenation would not be good, our users care a lot about the UI and how it is presented, so just concatenating and showing both does not work for us, we need to look for a way to not show the symbol kind. I also think that showing one or the other is better than just concatenating and not harder to implement.

About how to split PRs, or if an option should be added we are quite flexible :) 

Regards

mickaelistria:
I can support your reasoning against SymbolKind; I'm personally not really attached to those and wouldn't mind not seeing them by default, but let's go for smaller and safer steps and commits. That will allow to facilitate progress, more easily and faster increase the value, while also facilitating further decisions based on new experience and facilitate the capability to revert some decisions if we need to. Imagine here we merge the patch and someone suddenly comes with the need to SymbolKind, then we're a bit stuck and reverting to SymbolKind will require much work in separating this commit. If we separate items now, we can tell those people to revert the patch the does remove SymbolKind and that's all.

rubenporras:
Great, then we will split this PR into two different ones and let us review one after the other.

PR #281, opened 2022-11-04
Allow DSP launch with custom environment variables

sebthom:


jonahgraham (changes requested):
Please use `DebugPlugin.getDefault().getLaunchManager().getEnvironment(ILaunchConfiguration)` to get the environment. It handles `ATTR_APPEND_ENVIRONMENT_VARIABLES`, expanding eclipse variables, and case insensitive environment variables on Windows.

Annoyingly that method returns a `String[]` and you have to split on the first `=` to convert to the `Map<String,String>` that `ProcessBuilder.environment()` operates on.

org.eclipse.lsp4e.debug/src/org/eclipse/lsp4e/debug/launcher/DSPLaunchDelegate.java
```diff
@@ -260,6 +278,9 @@ public void launch(DSPLaunchDelegateLaunchBuilder builder) throws CoreException
 				}
 
 				ProcessBuilder processBuilder = new ProcessBuilder(command);
+				if (builder.debugCmdEnvVars != null && !builder.debugCmdEnvVars.isEmpty()) {
```
jonahgraham, line 281: PS. The alternative to reformatting the environment array is to change the launch itself to use `DebugPlugin.exec` instead which expects that format. The DebugPlugin.exec just uses ProcessBuilder anyway.

sebthom: Good idea. thanks!

jonahgraham (approved):
LGTM once it builds fine it can be merged. I'll try to remember to come back, but ping me if I don't.

PR #282, opened 2022-11-07
Move findReferences in LSPEclipseUtils

angelozerr:
Move findReferences in LSPEclipseUtils

This PR provide the capability to move the search LSP references in the LSPEclipseUtils in order to use it in another plugin / component.

I need that to fix https://github.com/eclipse/wildwebdeveloper/issues/644 in order to execute the LSP search references from CodeLens with XML language server.

### DIFF
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LSPEclipseUtils.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LSPEclipseUtils.java
@@ -71,6 +71,7 @@ import org.eclipse.jface.text.ITextViewer;
 import org.eclipse.jface.text.RewriteSessionEditProcessor;
 import org.eclipse.jface.text.TextSelection;
 import org.eclipse.jface.viewers.ISelectionProvider;
+import org.eclipse.lsp4e.operations.references.LSSearchQuery;
 import org.eclipse.lsp4e.refactoring.CreateFileChange;
 import org.eclipse.lsp4e.refactoring.DeleteExternalFile;
 import org.eclipse.lsp4e.refactoring.LSPTextChange;
@@ -99,6 +100,7 @@ import org.eclipse.lsp4j.VersionedTextDocumentIdentifier;
 import org.eclipse.lsp4j.WorkspaceEdit;
 import org.eclipse.lsp4j.WorkspaceFolder;
 import org.eclipse.lsp4j.jsonrpc.messages.Either;
+import org.eclipse.lsp4j.services.LanguageServer;
 import org.eclipse.ltk.core.refactoring.CompositeChange;
 import org.eclipse.ltk.core.refactoring.DocumentChange;
 import org.eclipse.ltk.core.refactoring.PerformChangeOperation;
@@ -106,8 +108,10 @@ import org.eclipse.ltk.core.refactoring.RefactoringCore;
 import org.eclipse.ltk.core.refactoring.resource.DeleteResourceChange;
 import org.eclipse.mylyn.wikitext.markdown.MarkdownLanguage;
 import org.eclipse.mylyn.wikitext.parser.MarkupParser;
+import org.eclipse.search.ui.NewSearchUI;
 import org.eclipse.swt.graphics.RGB;
 import org.eclipse.swt.graphics.RGBA;
+import org.eclipse.swt.widgets.Display;
 import org.eclipse.text.edits.MalformedTreeException;
 import org.eclipse.text.edits.MultiTextEdit;
 import org.eclipse.text.edits.ReplaceEdit;
@@ -1132,4 +1136,32 @@ public class LSPEclipseUtils {
 		}
 	}
 
+	/**
+	 * Execute Eclipse Search UI to search LSP references from the given document and
+	 * offset.
+	 *
+	 * @param document
+	 *            the document.
+	 * @param offset
+	 *            the offset.
+	 * @param display
+	 *            the display to use to execute the search.
+	 */
+	public static void searchLSPReferences(@NonNull IDocument document, int offset, Display display) {
+		LanguageServiceAccessor
+				.getLanguageServers(document,
+						capabilities -> hasCapability(capabilities.getReferencesProvider())) //
+				.thenAcceptAsync(languageServers -> {
+					if (languageServers.isEmpty()) {
+						return;
+					}
+					LanguageServer ls = languageServers.get(0);
+					try {
+						LSSearchQuery query = new LSSearchQuery(document, offset, ls);
+						display.asyncExec(() -> NewSearchUI.runQueryInBackground(query));
+					} catch (BadLocationException e) {
+						LanguageServerPlugin.logError(e);
+					}
+				});
+	}
 }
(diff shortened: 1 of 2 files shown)


org.eclipse.lsp4e/src/org/eclipse/lsp4e/LSPEclipseUtils.java
```diff
@@ -1132,4 +1136,32 @@ public static boolean hasCapability(Either<Boolean, ? extends Object> eitherCapa
...
+	 *            the document.
+	 * @param offset
+	 *            the offset.
+	 * @param display
+	 *            the display to use to execute the search.
+	 */
+	public static void searchLSPReferences(@NonNull IDocument document, int offset, Display display) {
+		LanguageServiceAccessor
+				.getLanguageServers(document,
+						capabilities -> LSPEclipseUtils.hasCapability(capabilities.getReferencesProvider())) //
```
rubenporras, line 1153: Can we remove the selfreference to `LSPEclipseUtils`?

angelozerr:
Here a draft PR in WWD side https://github.com/eclipse/wildwebdeveloper/pull/942 which uses this utility class at https://github.com/eclipse/wildwebdeveloper/pull/942/files#diff-87df74520a640db8f0ba722477b478c6c3910e5ff9ceb22ad1f246c95b4a8655R34

angelozerr:
To see the result with XML references in WWD, see https://github.com/eclipse/wildwebdeveloper/pull/942#issuecomment-1305281070

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LSPEclipseUtils.java
angelozerr: fixed

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LSPEclipseUtils.java
```diff
@@ -1132,4 +1136,32 @@ public static boolean hasCapability(Either<Boolean, ? extends Object> eitherCapa
...
+	 */
+	public static void searchLSPReferences(@NonNull IDocument document, int offset, Display display) {
+		LanguageServiceAccessor
+				.getLanguageServers(document,
+						capabilities -> hasCapability(capabilities.getReferencesProvider())) //
+				.thenAcceptAsync(languageServers -> {
+					if (languageServers.isEmpty()) {
+						return;
+					}
+					LanguageServer ls = languageServers.get(0);
```
rubenporras, line 1158: I realize you are just moving the code, but is the general pattern not to loop over all LSs, why take here just the first one?

mickaelistria:
I'm not found of the idea of crowding LSPEclipseUtils with "functional" units; it's mostly a set of utils to translate LSP<->Eclipse elements and actions; not a util class that's meant at orchestrating the Language Servers. So I would rather see the `references` package being exported than this code being moved.
Also, I'm wondering if we could have the initial feature part of LSP by default? This particular pattern of showing references in CodeLens and having a click opening them seems relatively independent of the LS, so could we instead host the handlers in LSP4E and make that consumer should just bind their specific command ids (eg `xml.show.references`) to this handler?

angelozerr:
It is exactly that I try to explain here https://github.com/eclipse/wildwebdeveloper/issues/644#issuecomment-787952909 in otherwords standardize some client command because I use a custom structure uri as first argument and line/character at the second arguments.

If you wish to provide that in LSP4e it means other LSP server should use the same structure, it was the reason why I wanted to standardize those client command at https://github.com/microsoft/vscode-languageserver-node/issues/495#issuecomment-787952518

mickaelistria:
> It is exactly that I tried to explain here https://github.com/eclipse/wildwebdeveloper/issues/644#issuecomment-787952909 in otherwords standardize some client command because I use a custom structure uri as first argument and line/character at the second arguments.

OK, I didn't understand the grain of action. But now we're a bit forward and the goal and proposal is clearer, I think it's indeed worth adding the handlers in LSP4E for such standard structures and commands.

angelozerr:
> OK, I didn't understand the grain of action. But now we're a bit forward and the goal and proposal is clearer, I think it's indeed worth adding the handlers in LSP4E for such standard structures and commands.

IMHO I think LSP should specify standard command ID like `lsp.showReferences` and after that any language server could generate codelens with this standar command id, but I fear it will never done.

I tried to define `lsp4e.showReferences` but Ihave no idea how to bind / associate / alias ?this command ID with the custom ID of XML language server `xml.show.references`. I wonder if we could define this mapping with extension point like this:

```xml
<extension
         point="org.eclipse.lsp4e.languageServer">
      <server>
   ...
      </server>
      <commandMapping
            standardCommandId="lsp4e.showReferences"
            commandId="xml.show.references">
```

angelozerr:
@mickaelistria do you like the idea with commandMapping extension point? If you like it I could experiment it.

mickaelistria:
I'm fine with LSP4E defining an Eclipse command or a handler for such common operation.
About a command mapping/alias, I don't think it should be LSP4E but the Eclipse command framework hosting this functionality.

mickaelistria:
I'm curious: does VSCode has the concept of generic and reusable commands? Doesn't it define a common findreference command? If yes, we could just have LSP4E adopting the same ids as VSCode and try to have lemminx using those common command ids.

If not, I see 2 other good solutions:
1. Create a strategy to define alias commands (in Platform ideally), or
2. make LemMinX look at the LSP client id (provided by the client to the server as part of InitializationOptions) and return the LSP4E command ids when it's LSP4E running.

My favorite is 1.

angelozerr:
> Create a strategy to define alias commands (in Platform ideally), or

I fear it will take some time to do that, and I need this feature for WWD.

> make LemMinX look at the LSP client id (provided by the client to the server as part of InitializationOptions) and return the LSP4E command ids when it's LSP4E running.

It will work only with LemMinx and not with another LS language server which could not provide this feature.

> I'm curious: does VSCode has the concept of generic and reusable commands? 

Indeed vscode defines some commands https://vscode-docs.readthedocs.io/en/latest/extensionAPI/vscode-api-commands/ and for references it defines `editor.action.showReferences`, but it works with vscode structure and not with lsp structure. In other I defined the LemMinx references command and call the vscode editor.action.showReferences by translating LSP structure to vscode structure https://github.com/redhat-developer/vscode-xml/blob/698f913553d2063cd6d9d72f18b53ae766e71ff1/src/commands/registerCommands.ts#L114

I think we could follow the same idea (I don't know if it is possible):

 * LSP4e defines a command `lsp4e.showReferences`
 * WWD XML defines a command `lemminx.showReferences` with an handler class which call `lsp4e.showReferences`.

Once Platform will define a command mapping, WWD could use it and remove the `lemminx.showReferences` handler class.

mickaelistria:
> LSP4e defines a command lsp4e.showReferences

LSP4E already defines a handler for org.eclipse.ui.genericeditor.findReferences command.
This handler can be improved to read an (optional) parameter which would be used with a code lens, and if this parameter is not set, use current selection as it does now.

> WWD XML defines a command lemminx.showReferences with an handler class which call lsp4e.showReferences.

We can try that for the moment. Wild Web Developer can create a "delegating" handler for its specific command, using command and handler services to retrieve the org.eclipse.ui.gemericeditor.findReferences command/handler and implement enablement/execution by delegating to it.

mickaelistria:

PR #283, opened 2022-11-07
Provide new signature to open URI without UI.getActivePage()

angelozerr:
Provide new signature to open URI without UI.getActivePage()

This PR provides new methods in LSPEclipseUtils without having IWorkbenchPage as parameter. The benefit with that is:

 * it simplifies a little the code.
 * it allows for another plugin to use open methods without redefining an utility class UI.getActivePage() since UI class is not exported.


rubenporras (approved):
looks good to me

PR #286, opened 2022-11-09
Create file on hyperlink

angelozerr:
This PR provides the capability to create a file when hyperlink is processed. vscode provides this feature and it is very good for some usecass. Here a demo which shows the benefit of this PR:

CreateFileOnHyperlink

### DIFF
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LSPEclipseUtils.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LSPEclipseUtils.java
@@ -20,10 +20,12 @@
  *******************************************************************************/
 package org.eclipse.lsp4e;
 
+import java.io.ByteArrayInputStream;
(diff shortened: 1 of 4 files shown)


mickaelistria:
Good addition! Do you think you can add a test case that covers it?

angelozerr:
> Good addition! 

Glad it please you:)

> Do you think you can add a test case that covers it?

I fear that it can be hard to write test since there is a message dialog which is displayed. I never written this kind of test.

> Do you know if this PR will always open the default create file dialog or if it is configurable?

If you see the demo, you will notice that there are not some create file dialog, just a confirm dialog which ask to the user if he want to create the file.

mickaelistria:
> I fear that it can be hard to write test since there is a message dialog which is displayed.

OK. Let's merge as it and cross fingers it doesn't regress later. If it does, then it will be a more motivating opportunity to write a test.

PR #287, opened 2022-11-09
fix mylyn URI

rubenporras:
The 3.0.42 is no longer available

### DIFF
--- a/target-platforms/target-platform-latest/target-platform-latest.target
+++ b/target-platforms/target-platform-latest/target-platform-latest.target
@@ -6,7 +6,7 @@
 			<unit id="org.eclipse.mylyn.wikitext" version="0.0.0"/>
 			<unit id="org.eclipse.mylyn.wikitext.markdown" version="0.0.0"/>
 			<unit id="org.eclipse.mylyn.wikitext.markdown.ui" version="0.0.0"/>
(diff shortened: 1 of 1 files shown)


sebthom:
How about we point it to the latest 3.x release which is accessible via https://download.eclipse.org/mylyn/docs/releases/3.0/

mickaelistria:
+1 for using latest release as much as possible.

rubenporras:
I prefer personally fixed versions but pointing to the latest release is also fine for me.

mickaelistria:
> I prefer personally fixed versions but pointing to the latest release is also fine for me.

I think it depends on your goal: for the *project*, it's often -if not always- better to use latest version of dependencies, unless those are known to have bugs. This improves the consuming project that usually benefits from fixes and other improvements for free, but also allows to build against some maintained version of its dependency (usually only latest release for OSS projects) so getting support is easier, and this also improves collaboration between projects and a better distribution of quality/value/activity/... across projects and thus reducing the risk illustrated in https://xkcd.com/2347/ .
For a *product* where some quality gates, manual testing, commercial commitments... are at play, then fixed versions give more certainty and reduces risk for this particular product in this particular version; so it can make more sense.

PR #289, opened 2022-11-10
Support "Always Up" language servers

BoykoAlex:
Draft implementation of "alwaysUp" attribute for language server extension point.

### DIFF
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServiceAccessor.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServiceAccessor.java
@@ -465,6 +465,20 @@ public class LanguageServiceAccessor {
 		}
 	}
 
+	public static @NonNull LanguageServerWrapper startLanguageServer(@NonNull LanguageServerDefinition serverDefinition) throws IOException {
+		synchronized (startedServers) {
+			LanguageServerWrapper wrapper = startedServers.stream().filter(w -> w.serverDefinition == serverDefinition).findFirst().orElseGet(() -> {
+				LanguageServerWrapper w = new LanguageServerWrapper(serverDefinition, null);
+				startedServers.add(w);
+				return w;
+			});
+			if (!wrapper.isActive()) {
+				wrapper.start();
+			}
+			return wrapper;
+		}
+	}
+
 	private static LanguageServerWrapper getLSWrapperForConnection(@NonNull IDocument document,
 			@NonNull LanguageServerDefinition serverDefinition, @Nullable IPath initialPath) throws IOException {
 


mickaelistria:
@martinlippert added some time ago https://github.com/eclipse/lsp4e/blob/master/org.eclipse.lsp4e/schema/languageServer.exsd#L145 . Setting this lastDocumentDisconnectedTimeout to Integer.MAX_VALUE should achieve good enough results already, with an uptime of 68 years.

martinlippert:
I actually think that keeping the language server running for a longer time (like Integer.MAX_VALUE) is only one part of the question. I think the other part (when to startup the language server) also requires some discussion.

The PR from @BoykoAlex allows you to basically startup the language server at IDE startup time, independent of any project contents in your workspace. While I can see this could be useful, I think it would also be good to maybe have some kind of API on the extension that can decide when to start the language server, e.g.:

- when the first content-type connected file is opened
- when there are files from those content-types in the workspace
- when there are specific (but random) files in the workspace
- some custom logic

WDYT?

mickaelistria:
I think from LSP4E perspective, we could adopt the proposed idea as it's not too intrusive. Or maybe, just have LSP4E providing an `org.eclipse.lsp4e.LSStartup` extension factory that consumers could use in their IStartup with the syntax:
```xml
  <extension point="org.eclipse.ui.startup">
     <startup class="org.eclipse.lsp4e.LSStartup:org.eclipse.corrosion.rls"/>
  </extension>
```
?

About allowing multiple strategies, including not yet identified one, I believe it can become too custom to be worth being standardized in LSP4E. Consumers of LSP4E would IMO be more efficient at implementing whichever hook or listener they want by themselves and trigger the LS startup. Maybe we need a bit more API for it to be easy though.

rubenporras:
I also use https://github.com/eclipse/lsp4e/blob/master/org.eclipse.lsp4e/schema/languageServer.exsd#L145 as a poor man solution, if there is an option to start the LS on startup or something more configurable, that would be a good addition.

The proposed extension point

  <extension point="org.eclipse.ui.startup">
     <startup class="org.eclipse.lsp4e.LSStartup:org.eclipse.corrosion.rls"/>
  </extension>

would be just what we need if we combine with the Integer.MAX_VALUE to keep the server running, which we already use

BoykoAlex:
Reading the above, perhaps, all LSP4E needs is public `LanguageServiceAccessor.startLanguageServer(String id)` where `id` is the LS definition id from the extension point. The method would start the LS if not already started. The logic/conditions under which the LS to be started would go into a downstream LS plugin which could either start the LS at IDE start up (i.e. `IStartup` interface via the startup extension) or at some specials conditions met by the workspace projects (i.e. java project present with spring on the classpath).
How does this sound?

mickaelistria:
I imagine you can already achieve a startup in your code with a combination like
```java
var definition = LanguageServerRegistry.getInstance().getDefinition(languageServerId);
var dummyFile = File.createTempFile("initLS", "someExtension");
var fileBufferManager = FileBuffersPlugin.getFileBufferManager();
var dummyDocument = fileBufferManager.getFileBufferManager().createEmptyDocument(new Path(dummyFile.getAbsolutePath()), LocationKind.LOCATION);
getInitializedLanguageServer(dummyDocument, definition, cap -> true) // actually starts the LS
   .thenRun(() -> {
      fileBufferManager.disconnect(dummyFile.getAbsolutePath(), LocationKind.LOCATION, null);
      dummyFile.delete();
   });
```
But a shorter API such as `LanguageServiceAccessor.startLanguageServer(LanguageServerRegistry.getInstance().getDefinition(languageServerId))` would be welcome.

BoykoAlex:
@mickaelistria I'd stop at your proposal: `LanguageServiceAccessor.startLanguageServer(LanguageServerRegistry.getInstance().getDefinition(languageServerId))`

Let us work out things on our side and if this is all we need (which is very likely) then I'll change this PR accordingly and prepare it to be merged.

mickaelistria (approved):
Looks good. Please mark as "Ready for review" if this is ready to be merged on your end as well.

org.eclipse.lsp4e/plugin.xml
```diff
@@ -662,5 +662,5 @@
            </or>
          </enabledWhen>
       </reconciler>
-   </extension>   
+   </extension>
```
rubenporras, line 665: It would be good to remove this file form the commit.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServerWrapper.java
```diff
@@ -683,11 +683,13 @@ public void disconnect(URI uri) {
 			documentListener.documentClosed();
 		}
 		if (this.connectedDocuments.isEmpty()) {
-			if (this.serverDefinition.lastDocumentDisconnectedTimeout != 0) {
-				removeStopTimer();
-				startStopTimer();
-			} else {
-				stop();
+			if (!this.serverDefinition.alwaysUp) {
```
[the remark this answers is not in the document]
BoykoAlex: Switched to a solution where we just have the method for starting the language server

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServiceAccessor.java
```diff
@@ -465,6 +465,20 @@ private static LanguageServerWrapper getLSWrapper(@Nullable IProject project,
 		}
 	}
 
+	public static LanguageServerWrapper startLanguageServer(@NonNull LanguageServerDefinition serverDefinition) throws IOException {
+		synchronized (startedServers) {
+			LanguageServerWrapper wrapper = startedServers.stream().filter(w -> w.serverDefinition == serverDefinition).findFirst().orElse(null);
+			if (wrapper == null) {
+				wrapper = new LanguageServerWrapper(serverDefinition, null);
```
rubenporras, line 472: This looks good to me and I would be happy to merge, but I was wondering if we could do better than `orElse(null)`. Could we not just use `new LanguageServerWrapper(serverDefinition, null)` in the `orElse()` above and then just call `startedServers.add(wrapper)`, the startedServers is a set so even if the wrapper is already in it, nothing bad should happen. If we want to avoid the add when the value was already present, we could use `orElseGet()`.

What do you think about that?

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServiceAccessor.java
```diff
@@ -465,6 +465,20 @@ private static LanguageServerWrapper getLSWrapper(@Nullable IProject project,
 		}
 	}
 
+	public static LanguageServerWrapper startLanguageServer(@NonNull LanguageServerDefinition serverDefinition) throws IOException {
```
rubenporras, line 468: add @NonNull to the return value?

mickaelistria:
@BoykoAlex when it's ready to be merged, please click "Ready for review".

BoykoAlex:
@mickaelistria @rubenporras do you know what i can do with `enabledWhen` element under `languageServer` extension `contentTypeMapping`? I'd like to check if the resource being opened in the editor belongs to a Java project with spring boot on the classpath... Is there a way to get to the resource with propertyTesters under `enabledWhen` element? I looked at the context object which is being passed to the evaluation and seems like i might be able to get active `IEditorInput` from it... However, I'm curious how `enabledWhen` is being used now...

rubenporras:
Hi @BoykoAlex,

I do not know this API very well, but we have testers that override this method 

```
  @Override
  public boolean test(final Object receiver, final String property, final Object[] args, final Object expectedValue) {
    Assert.isInstanceOf(receiver, IContainer.class, "The receiver must be an IContainer."); //$NON-NLS-1$
    IContainer container = (IContainer) receiver;
    IProject project = container.getProject();
    if (project.isOpen()) {
      ...
    } else {
      return false;
    }
  }
```

and then something like

```
            <enabledWhen>
               <iterate
                     operator="and">
                     <adapt type="org.eclipse.core.resources.IContainer">
                      <test
                            property="com.........myTester">
                      </test>
                     </adapt>
               </iterate>
            </enabledWhen>

```
does it bring you further?

BoykoAlex:
@mickaelistria thank you - I'll give it a try... Was roaming yesterday through lsp4e code and after looking at https://github.com/eclipse/lsp4e/blob/dd39ca0db5e6abd3cb5f59a3e63cf6b2e1a20d46/org.eclipse.lsp4e/src/org/eclipse/lsp4e/enablement/EnablementTester.java#L61 I've decided that evaluation context is just instance of `Object` that is not useful at all :-\

BoykoAlex:
Ready to merge this one. I'll move `enabledWhen` proposals into a separate PR if applicable

rubenporras (approved):

rubenporras:
@mickaelistria , ok to merge?

mickaelistria:
Yes, let's just ensure we have a green CI build before merging (the last build had some rare but random test failure, let's just verify it's not persistent)

mickaelistria:
Thanks!

PR #295, opened 2022-11-11
Add a torture test for dispatch ordering

ahmedneilhussain:
Illustrates the problem with document updates and other requests sometimes interleaving in the wrong order.

### DIFF
--- a/org.eclipse.lsp4e.test/src/org/eclipse/lsp4e/test/edit/DocumentDidChangeTest.java
+++ b/org.eclipse.lsp4e.test/src/org/eclipse/lsp4e/test/edit/DocumentDidChangeTest.java
@@ -18,24 +18,36 @@ import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNotNull;
 
 import java.io.File;
+import java.net.URI;
+import java.util.Collections;
 import java.util.List;
+import java.util.Vector;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.atomic.AtomicInteger;
 import java.util.function.Predicate;
+import java.util.stream.Collectors;
 
 import org.eclipse.core.filesystem.EFS;
 import org.eclipse.core.resources.IFile;
 import org.eclipse.core.resources.IProject;
 import org.eclipse.core.runtime.CoreException;
+import org.eclipse.jface.text.IDocument;
 import org.eclipse.jface.text.ITextViewer;
 import org.eclipse.lsp4e.LSPEclipseUtils;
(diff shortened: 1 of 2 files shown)


ahmedneilhussain:
Not intended for merge - just for contrast with the single-threaded-dispatch PR https://github.com/eclipse/lsp4e/pull/251

@mickaelistria @rubenporras hopefully this shows the problem a little more clearly (and that the accidentally-merged solution doesn't fix it...)

Sadly with this sort of thing Your Mileage May Vary, as they say: it will depend on your build environment and number of cores and so on. It fails reliably for me on my local machine, and if I create some noise on the machine by running a build or other CPU-intensive task in the background then it will fail much more extensively, e.g. 70 iterations out of 500 will arrive out of order.

rubenporras:
I also see a lot of value on this test, but we know it will be flaky with the current code, so we cannot merge, can we?

Or is the reasoning that this tests shows that https://github.com/eclipse/lsp4e/pull/251 is an improvement and that we should clean up https://github.com/eclipse/lsp4e/pull/251 to include just message ordering and this test (remove e.g. the check for outdated responses) and send that for review? Is this so, @mickaelistria?

I know I am coming again and again back to the same point, but I really do not know if we have by now the same understanding on the way forward.

mickaelistria:
Thanks, I'm looking at it right now.

mickaelistria:
Thanks a lot for this test; I now (finally!) understand the issue. My previous understanding was limited to the view of the LS client, and I didn't realize that the issue was more the impact on server side. This tests makes it clear and I can reproduce the issue in 100% of time (usually between 495 and 498 success per test out of 500) locally.
I'll revisit the other patches and proposals with my now hopefully complete understanding of the problem.

ahmedneilhussain:
Thanks @mickaelistria that's great. If you want to talk through any of the points (either the problem, or the various proposed solutions) in more detail, please let us know.

mickaelistria:
I've spent some time on it and added some more logs (mainly to see when the requests are sent by the client; then received by the server and also logged explicitly whether hover was received too late or too early relatively to the change event). This gave various results but one of them seemed to be that the change events seemed to be sometimes happening later; which is strange according to the code. I found out 1 possible cause (I've submitted PR #318 ), and it seems to make the test pass. I don't hope it's a complete solution to all problems; but IMO it would be good to merge it and evaluate it against other issues you've noticed. Maybe it happens to be enough for most issues.

mickaelistria:
This one was merged with #318.
It was a great contribution to this topic, allowing to clearly define the issue and more easily test potential solutions. Thanks a lot!

PR #302, opened 2022-11-15
Don't add initially disabled breakpoints

mickaelistria:
Fixes https://github.com/eclipse/lsp4e/issues/301


org.eclipse.lsp4e.debug/src/org/eclipse/lsp4e/debug/debugmodel/DSPBreakpointManager.java
```diff
@@ -17,6 +17,8 @@
 import java.util.Objects;
 import java.util.concurrent.CompletableFuture;
 
+import javax.xml.transform.Source;
```
jonahgraham, line 20: Bad import?

jonahgraham (changes requested):

jonahgraham:
Other than the import change this change LGTM

org.eclipse.lsp4e.debug/src/org/eclipse/lsp4e/debug/debugmodel/DSPBreakpointManager.java
mickaelistria: Indeed.

PR #303, opened 2022-11-16
Increase bundle version rane for mylyn plugins, as the 3.0.40 is no l…

rubenporras:
…onger available for download.


rubenporras:
@mickaelistria , okay to merge this? 

We are getting the following

Unresolved requirement: Require-Bundle: org.eclipse.lsp4e
  -> Bundle-SymbolicName: org.eclipse.lsp4e; bundle-version="0.14.0.202208241334"; singleton:="true"
     org.eclipse.lsp4e [343]
       Unresolved requirement: Require-Bundle: org.eclipse.mylyn.wikitext; bundle-version="3.0.40"

after upgrading mylyn to 3.0.45 in our packaging with lsp4e 0.14.0. I guess we would get the same error even with the latest snapshot based on master. That is the reason behind the commit.

rubenporras:
@mickaelistria , of course it would need to pass the build successfully, which it does not.

I do not understand the current error

> [ERROR]   Missing requirement: org.eclipse.lsp4e 0.15.0.qualifier requires 'osgi.bundle; org.eclipse.mylyn.wikitext [3.0.40,3.0.45]' but it could not be found

we have 3.0.42 in the target file. I also do not understand why it was building fine before if we where pinning the version to 3.0.40 and had 3.0.42 in the target file. 

Tycho is a bit of a mystery to me.

org.eclipse.lsp4e/META-INF/MANIFEST.MF
```diff
@@ -29,8 +29,8 @@ Require-Bundle: org.eclipse.core.runtime;bundle-version="3.12.0",
  org.eclipse.ui.navigator;bundle-version="3.6.100",
  org.eclipse.lsp4j;bundle-version="[0.18.0,0.19.0)",
  org.eclipse.lsp4j.jsonrpc;bundle-version="[0.18.0,0.19.0)",
- org.eclipse.mylyn.wikitext;bundle-version="3.0.40",
- org.eclipse.mylyn.wikitext.markdown;bundle-version="3.0.40",
+ org.eclipse.mylyn.wikitext;bundle-version="[3.0.40,3.0.45]",
```
mickaelistria, line 32: Why are you setting and upper bound. "3.0.40" actually means ">= 3.0.40" according to https://docs.osgi.org/specification/osgi.core/8.0.0/framework.module.html#i3189032 , so this is correct.

rubenporras:
I understand, I was reading the same spec now :) I close the PR and will investigate further what is the problem with our setup.

Issue #219, opened 2022-08-23
LSPFormatHandler.execute() does not wait for the format request, might trash the editor contents


rubenporras:
Since LSPFormatHandler.execute() does not wait for the format request blocking the editor and there is no visual feedback that the system is executing a format action, if the user types while format is being processed, once the format request comes back, the async method

```
				formatter.requestFormatting(document, (ITextSelection) selection).thenAcceptAsync(
						edits -> shell.getDisplay().asyncExec(() -> formatter.applyEdits(document, edits)));

```

will apply the edits which came back from the formatter, and either throw BadLocationException or trash the content or the editor buffer (by replacing existing regions which are modified with the contents returned by the formatter).

rubenporras:
@mickaelistria , I suggest, that to solve the issue, the editor should wait before applying the edits by using `thenAccept` instead of `thenAcceptAsync` or get the future using a timeout and discarding the answer if the timeout is exceeded.

Since here the user manually requests formatting, I think the timeout can be high, so to me using `thenAccept` is just fine. What do you think?

mickaelistria:
I believe that wouldn't be enough: it's very possible that `requestFormatting` takes a lot of time, and editor isn't frozen during that time and user can change content in the meantime. Using thenAccept wouldn't block the editor either.
I believe we really need a way to explicitly lock the document or the viewer here; and to unlock it upon completaion or failure.

PR #308, opened 2022-11-18
Prevent LSPFormatHandler.execute from trashing the editor contents

rubenporras:
If between a format request and the moment the text edits are applied the document has been edited, the text edit is discarded instead of applied. Thus we prevent user edits which happened concurrent to the request to be trashed without needing to lock the document or the UI.

When this happens the user gets informed with a notification.


rubenporras:
@ahmedneilhussain , this is an alternative to https://github.com/eclipse/lsp4e/pull/174 for formatting. I like it because it is a very small change but still does the job.
@mickaelistria , would be fine for you to merge this version?

ahmedneilhussain:
@rubenporras I don't think this can work, unfortunately.

This change does:

1. Read version
2. Send formatting request
3. Read version on receipt of edits
4. Apply changes if 1 == 3

I don't think this can work because of some of the other issues we discus elsewheres: it's not safe to read the version 'now' from the document and assume that that's the one that the server will be using to compute the response. I think the following is still possible:

1. Document at version a
2. User applies edits -> version b
3. Read version as b
4. Send format request
5. Server starts computing response based on document at version a
6. Document update a->b dispatched to server
7. Format response (based on a) received
8. Read document version as b, and incorrectly conclude it is safe to apply edits received in 7

The reason https://github.com/eclipse/lsp4e/pull/174 is more complicated than your PR is that in the absence of dispatch ordering, you have to use the DocumentContentSynchronizer to keep track of what version the server has been told about, and ensure that sending a versioned request is done atomically.

A big part of the motivation for #251 is that guaranteeing the dispatch ordering makes this kind of subtle issue go away: it makes reading the document version in the foreground (i.e. on the event thread) a lot safer.

rubenporras:
Hi @ahmedneilhussain ,

I think that both problems should can be treated independently. In this PR does not make any assumption on the topic about sending out of order messages from the client to the server will because, even if client messages are ordered, one still need to check that the result can be applied. The client must always check that the answer to the server still applies (and that is clearly in the LSP spec), that is the only thing I make sure here. So I am only solving https://github.com/eclipse/lsp4e/issues/219.

This solution would work also after  #251 or anything else that achieves something similar is merged, and it is quite simple :)

Regards

ahmedneilhussain:
You are right in that the 'optimistic locking' problem is *mostly* orthogonal to the dispatch ordering one. However, the former needs to take account of the latter if you want to guarantee that document corruption does not occur.

That was the reason #174 is more complex than this one. You need to make sure that the version of the document you apply the changes to is the one that the server used to compute those changes. Your solution checks that the [client] version of the document you apply the changes to is the same as the [client] version you read when you sent the format request: that may or may not be the one the server uses to compute the edits, depending on whether we are unlucky with our format request overtaking an earlier edit en route to the server. If you apply edits computed on a different version of the document, then corruption will occur, even if from the client's point of view nothing changed...

Your change mitigates one source of corruption, but if you are worried about corruption then you are worried about corruption: it doesn't matter whether it arises from the user typing after the format request is sent or because the format request is sent before earlier edits have caught up - corruption may still occur.

I suppose this change removes one source of it, so you could take a view on it being nevertheless a worthwhile improvement, but to be definitely secure #174 or something like it is needed unless one first implements something like #251 :-(

rubenporras:
Exactly, this change removes one source of it (one which cannot be removed by #251, wy the way). I agree that to be definitely secure formatting something like https://github.com/eclipse/lsp4e/pull/251 is needed, but there is no harm in fixing one source of trouble while we find a suitable pattern for ordering the messages sent by the client, specially if the fix is that easy as shown in here :)

mickaelistria (approved):

mickaelistria:
@rubenporras The general idea is OK, but there are some compile issue that need to be fixed before we can merge.

rubenporras:
Thanks for the review. I have missed those errors in my IDE. I will fix them before merging.

rubenporras:
I am traveling, I will fix the remaining problem later

PR #310, opened 2022-11-18
[DO NOT MERGE] Enablement for specific input

BoykoAlex:
`enabledWhen` doesn't have editor input, fie, document or any kind of input context for evaluation. It is used in `ContentTypeToLanguageServerDefinition#isEnabled()` calls. These calls are made also from content type -> LS preference page which is independent of any input context.

I'm curious if it is worth making a similar `enabledInputWhen` tester able to take input context such as `IDocument`, `IEditorInput`, `IFile` etc. Therefore whenever there is input available we'd pass it over to ContentTypeToLanguageServerDefinition#isEnabledFor(Object input)` rather than `isEnabled()`. The PR adds implementation for this.

The goal i have in mind is start Boot LS for a workspace that has at least one java project with spring boot on the classpath. Shut down Boot LS if there are no more Spring Boot projects left in the workspace. Now if Boot LS is shut down and user opens a Java file (or any other boot related content-type files) in an editor I wouldn't want to start Boot LS for a file that does not belong to a Spring Boot project. Hence, the `enabledInputWhen` would help in this case.
What do you think? Do you see any issues/obstacles achieving the goal above with the proposed PR?

Thanks in advance for feedback on this topic!!!

### DIFF
--- a/org.eclipse.lsp4e/schema/languageServer.exsd
+++ b/org.eclipse.lsp4e/schema/languageServer.exsd
@@ -158,6 +158,7 @@ If set to a number bigger than zero, the server will run until the timeout is re
       <complexType>
          <sequence>
             <element ref="enabledWhen" minOccurs="0" maxOccurs="1"/>
+            <element ref="enabledInputWhen" minOccurs="0" maxOccurs="1"/>
          </sequence>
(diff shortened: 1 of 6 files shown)


mickaelistria:

PR #318, opened 2022-11-23
No async did change

mickaelistria:


mickaelistria:
So the test clearly shows it's not enough. Locally I can reproduce errors, with 10000 iterations I get less than 10 "too late" Hover processed. I believe the change has prevented some "too early" from happening. I hope that if we start removign some Async, we can get it all working without functionality decrease.

rubenporras:
Hi @mickaelistria,

I am happy to see that you are doing improvements, but I am worried because I do not understand what is the plan to get it to work fully. I do not think that removing some Async will eliminate the out of order messages, will it?

Would you mind explaining what is the idea to fully prevent out of order messages? Is there a rough timeline? The reason I would like to know is that this is the only blocking issue we have on LSP4E before taking it into production.

For a production ready editor we need to it to be fully reliable. Otherwise I know in advance users will complain and be unhappy, and out of order messages will make analysis of any kind of bug very challenging. Thus for us, 10 "too late" out of 10000, or just one, or even less is not good enough. I understand that software has bugs, but this one is known and can be effectively removed.

 Regards

mickaelistria:
>  I do not think that removing some Async will eliminate the out of order messages, will it?

I think it can actually elimate the vast majority of them. Using non-Async then* operations makes that the following operation runs in the same thread, and as LSP4J uses a thread-local and LSP4E configures it to use a singleExecutor, then it means that forcing usage of the same thread cause that this same thread is "blocking" other operations. So chains of operations that are using non-Async then will actually run sequentially without leaving a room for other change to jump in between.

> The reason I would like to know is that this is the only blocking issue we have on LSP4E before taking it into production.

I believe we need to review usage of then*Async everywhere, making sure that the Async is only called when we're not orchestrating the LS any longer, but when we're moving to consumption on client-side.
As for the issues you see, please try to create a test case for those. It will help to review whether the current state of thing is already better (I think it is), sufficiently better (I think we're not too far) and what particular parts can be improved, typically by removing some Async in LS side..

> hus for us, 10 "too late" out of 10000, or just one, or even less is not good enough. I understand that software has bugs, but this one is known and can be effectively removed.

Please submit the cases where you see such issues and we can try to fix them. Again, I do believe than removing then*Async here and there can be the right fix. But having scenarios to harness the development would be what would lead to best result.

rubenporras:
So you are proposing something like https://github.com/eclipse/lsp4e/pull/322, right?

So in essence instead of using a new single thread dispatcher as in https://github.com/eclipse/lsp4e/pull/251 we just use the application main thread. Is this correct? From my understanding that would work and I am happy with it. Do you agree @ahmedneilhussain  and @tomgr?

mickaelistria:
> So in essence instead of using a new single thread dispatcher as in https://github.com/eclipse/lsp4e/pull/251 we just use the application main thread.

It's not the application main thread (which is the one that powers UI and should never be blocker), it's the language server connection thread that is used.

ahmedneilhussain:
> > So in essence instead of using a new single thread dispatcher as in #251 we just use the application main thread.
> 
> It's not the application main thread (which is the one that powers UI and should never be blocker), it's the language server connection thread that is used.

Hi @mickaelistria unfortunately that's not the case. @tomgr and I were looking through your revised approach, and noted that it would have the effect of making the UI thread do the bulk of the dispatch work. I was literally spending this morning throwing together a test showing that this is the case and that this can cause delays in the UI if the remote server does not consume messages 'instantly'.

Please see https://github.com/eclipse/lsp4e/pull/323
(NB we don't intend this PR to be merged as is, it is just to illustrate the point).

I have double checked this by using master following your recent changes to run the corrosion plugin locally. You can try yourself: launch the plugin in debug mode, open a rust file, then place a breakpoint at https://github.com/eclipse/lsp4e/blob/ce79cb56970bf6b661639009bf2a433d4d186d69/org.eclipse.lsp4e/src/org/eclipse/lsp4e/DocumentContentSynchronizer.java#L163
If you start typing into the rust file then you should find that breakpoint trips on the event thread.

ahmedneilhussain:
The reason the UI thread is doing the bulk of the work is that the majority of our calls to the language server are now essentially of the form
`getInitialisedServer().thenCompose(...)`

However, when you write `t.andThen(() -> f())` and t is a completable future that *has already completed* then this is the same as just writing `f()` directly - i.e. the lambda supplied to any of the `.andThen`, `.thenCompose`, `thenApply` methods runs in the caller's thread.

The pattern that happens the majority of the time for lsp4e is that there will be an initial delay when establishing the connection to the server/setting up the initial documentOpen message. While those calls are ongoing, the way we chain further calls mean that they will indeed be dispatched asynchronously once the initialisation completes, on the server *response* thread.

However, once the initialisation futures have succeeded, any further calls chained onto them will just dispatch immediately on the callers thread. Since most things in lsp4e/eclipse are initiated on the SWT event thread, it will be this thread that will be writing the messages to the server.

ahmedneilhussain:
> > So in essence instead of using a new single thread dispatcher as in #251 we just use the application main thread.
> 
> It's not the application main thread (which is the one that powers UI and should never be blocker), it's the language server connection thread that is used.

As we mentioned in the discussion on #251, the thread pool started at https://github.com/eclipse/lsp4e/blob/ce79cb56970bf6b661639009bf2a433d4d186d69/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServerWrapper.java#L276 is only used by RemoteEndpoint for *inbound* messages from the server. There's no 'language server connection thread' for outbound messages at the moment - that's why #251 creates one!!

The reason (prior to your recent changes) that most calls to the language server did not take place on the event thread was because we were mostly using constructs of the form `t.thenApplyAsync( {code} )`, in which case `{code}` will execute on one of the default thread worker pool's threads - but unfortunately without any control over the dispatch order.

tomgr:
> Do you think it would be best to revert it or better to keep it as it? We do need a release of LSP4E tonight or tomorrow...

One reason to consider reverting the change would be that with this change, it's more likely code will be executed on the UI thread. If someone is running LSP4E with a synchronous blocking server at the moment then when they upgrade LSP4E, they might find their Eclipse instance locks up quite a bit. I'm not sure how likely a scenario that is, but if you think it is likely I would revert the change to avoid the impact.

rubenporras:
> I was literally spending this morning throwing together a test showing that this is the case and that this can cause delays in the UI if the remote server does not consume messages 'instantly'

@ahmedneilhussain, @tomgr: thanks a lot for this insight, it is very valuable. I have noticed that the deserialization work would be happening on the UI (or as I called main) thread when testing myself, but I thought the work was as quick as serialization would go. I was not aware that it would block until the server actually consumes the message.

ahmedneilhussain:
> > I was literally spending this morning throwing together a test showing that this is the case and that this can cause delays in the UI if the remote server does not consume messages 'instantly'
> 
> @ahmedneilhussain, @tomgr: thanks a lot for this insight, it is very valuable. I have noticed that the deserialization work would be happening on the UI (or as I called main) thread when testing myself, but I thought the work was as quick as serialization would go. I was not aware that it would block until the server actually consumes the message.

Thanks @rubenporras - JSON serialization/deserialization is *probably* not going to be a problem most of the time, but it could be an issue for large requests, i.e. when working with large open files. A badly-behaved server could really lock up eclipse, so is potentially much more of an issue.

tomgr:
> As for a final solution, I have the impression that ideally, LSP4J could provide to the launcher or someplace else an Executor to use for the future that are created when invoking the API. Do you think it would make sense? Do you think it would be possible?

I don't think that's enough, if I understand correctly. Let's suppose that the server is already initialised; in this case the call to `getInitializedServer()` will return an already-completed future. I believe this is still the case under your new proposal? If so we are a bit stuck here:

1. If we use `thenApply(f)` then `f` will be executed on the UI thread.
2. If we use `thenApplyAsync(f)` then `f` will be executed on the Java-supplied default executor, which will break the ordering guarantees.

Given this, I don't see how overriding the executor will help. I suppose the basic problem is that the `thenApplyAsync` methods (which we need to use to avoid UI threading) are defined as using the Java supplied executor.

I think given the importance of getting the order of requests sent to the server correct, I would really like an API that does not require someone to understand the details of the underlying futures API too much, and doesn't depend on them thinking too much about which thread executes their code. The nice thing about #251 is that it provides a harder-to-misuse API which should help improve the robustness of future commands without much effort. I think any fix that works by overriding executors is fragile: it will still rely on the user carefully using `thenApply` and `thenApplyAsync` in precisely the right places to ensure a) the UI thread is not blocked and b) the document ordering is respected.

Maybe there is a better API than #251, but I do think the concept of providing a hard-to-misuse base API is worth pursuing.

PR #320, opened 2022-11-27
Do not fail on non-accessible resources

mickaelistria:
Some language servers may send diagnostics for resources that are not accessible (eg closed).
Do not fail in such case.

### DIFF
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/diagnostics/LSPDiagnosticsToMarkers.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/diagnostics/LSPDiagnosticsToMarkers.java
@@ -86,7 +86,7 @@ public class LSPDiagnosticsToMarkers implements Consumer<PublishDiagnosticsParam
 		try {
 			String uri = diagnostics.getUri();
 			IResource resource = LSPEclipseUtils.findResourceFor(uri);
-			if (resource != null && resource.exists()) {
+			if (resource != null && resource.isAccessible()) {
 				updateMarkers(diagnostics, resource);
 			} else {
 				LSPEclipseUtils.findOpenEditorsFor(LSPEclipseUtils.toUri(uri)).stream()
@@ -169,7 +169,7 @@ public class LSPDiagnosticsToMarkers implements Consumer<PublishDiagnosticsParam
 					}
 				});
 			} catch (CoreException e) {
-				if (resource.exists()) {
+				if (resource.isAccessible()) {
 					LanguageServerPlugin.logError(e);
 				}
 			} finally {


org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/diagnostics/LSPDiagnosticsToMarkers.java
```diff
@@ -130,6 +130,9 @@ private void updateEditorAnnotations(@NonNull ISourceViewer sourceViewer, Publis
 	}
 
 	private void updateMarkers(PublishDiagnosticsParams diagnostics, IResource resource) throws CoreException {
+		if (!resource.isAccessible()) {
+			return;
```
rubenporras, line 134: if we want to just skip silently this resource, can we not just remove the `LanguageServerPlugin.logError(e)` on line 176? That would be better for the performance, as testing for existence is not for free, and we expect to usually have the resource available. If need be, we can also check accessibility in addition to existence when catching the `CoreException`.

mickaelistria: Did you see some issue with current logging? I'm not against changing it, but if there is no actual reason to do it, I'd rather keep it separated from this change.

rubenporras: No, I do not have a problem with the current logging. I have a problem only with calling  `!resource.isAccessible()` as a preemptive measure, since it has a cost and the condition will happen very rarely. I would prefer to be optimistic, and catch an exception if the resource is not accessible. To do that we would just adding `!resource.isAccessible()` to the condition on line 175 (or 172 before this change). It would be equivalent to this PR but creating a log entry, that is why I proposed to remove the log entry we created in that branch. If we are happy with the log entry (which I am), I would just move line 133 to line 175.

Note that line 175 already checks `resource.exists()`, which in some implementation is the same as `resource.isAccessible()`,  so I wonder if you have some stack trace available, maybe the existing catch is not on the right place?

mickaelistria: You're entirely right. I did change the patch to replace existing  `resource.exists()` by `resource.isAccessible()` and it provides a good result.

rubenporras (approved):
This looks good to me. While at it, should we also not replace the call to `exists` on line 172 to `isAccessible` as well?

mickaelistria:
> While at it, should we also not replace the call to exists on line 172 to isAccessible as well?

Yes, done. Thanks!

PR #321, opened 2022-11-29
Currently the log is using currentTimeMillis, see below

rubenporras:
[t=1667223932171] LSP4E to ...:

For debugging and readability this is not useful, as usually one would like to correlate logs from the client and the server, instead use the standard:

[2022-11-29T13:20:11.220291200+01:00] LSP4E to ...:

Also use a `StringBuilder` directly to create the message instead of creating a new `String` with the payload and then adding that to other strings with `+`. Thus we avoid one `String` creation.

### DIFF
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LoggingStreamConnectionProviderProxy.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LoggingStreamConnectionProviderProxy.java
@@ -20,6 +20,7 @@ import java.io.OutputStream;
 import java.net.URI;
 import java.nio.file.Files;
 import java.nio.file.StandardOpenOption;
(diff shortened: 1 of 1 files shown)


rubenporras:
Something such as 
2022-11-29T13:20:11.220291200+01:00
has also the milisecond information (it has actually more precision than that), so if you have a bunch of logs, it is also possible to know how much time passed (sure it is not as straightforward, but very easy). 
It has the huge advantage of allowing you not only to know the elapsed time but also when the event happened, that is very useful when analyzing logs from users that reported. It also enables matching logs from the language server to the ones from the language client.

I think it this is more sensible default, but if you think it is not, we would need a way to customize it. Should I rather work on that?

mickaelistria:
OK, let's try.

PR #322, opened 2022-11-29
Use thenApply for hover requests

rubenporras:
following commit 534eb03436ba86d587899f7c38e1863717f7371f to avoid sending out of order messages from the client.


mickaelistria:
As explained in https://github.com/eclipse/lsp4e/pull/318#issuecomment-1330738521 and as opposed to my initial understanding, using non-async `then` operation can cause other issues. So let's abandon that path.

rubenporras:
@tomgr, @ahmedneilhussain , thanks for looking into such level of detail on this and the reverted Async implementation.

PR #323, opened 2022-11-29
[DO NOT MERGE] Add a 'test' showing that a blocking server can block the UI thread

ahmedneilhussain:
The effect of some of the concurrency changes recently means that many of the server calls will dispatch on the UI thread. If the server is blocking then this can in turn cause the UI to block.

This is not *necessarily* a problem for all language servers, but it might be in certain circumstances. A language server that is completely single threaded in nature might need to wait for a long computation (e.g. a refactoring) to finish before it can respond to further input. Even a multi-threaded language server that is capable of queueing requests and handling multiple ongoing computations simultaneously can obviously become overloaded.

The mechanism for back-pressure isn't immediately obvious, and will depend on connection between client and server, but for the default stdin/stdio stream-backed connection, then once the server becomes tied up it will stop reading from its input stream. If the streams are unbuffered, then the language client will block attempting to write. I am reliably informed that 8k is a normal-ish default buffer size for a unix pipe, so have created a payload in this test that will fill up the buffer in a couple of requests.

The results are obviously environment dependent, but I get dispatch timings of 14s and every hover request dispatched on the UI thread (i.e. `Display.getCurrent() != null`)

### DIFF
--- a/org.eclipse.lsp4e.test/src/org/eclipse/lsp4e/test/edit/DocumentDidChangeTest.java
+++ b/org.eclipse.lsp4e.test/src/org/eclipse/lsp4e/test/edit/DocumentDidChangeTest.java
@@ -24,6 +24,7 @@ import java.util.Collections;
 import java.util.List;
 import java.util.Vector;
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.atomic.AtomicInteger;
 import java.util.function.Predicate;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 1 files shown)


ahmedneilhussain:
FYI @gerdleon

mickaelistria:
Note that we could already merge it as `@Ignore` and include some comments about the issue it does cover and pointers to discussions. This may facilitate further progress.

ahmedneilhussain:
The new api merged in #344 explicitly has tests for making sure remote request dispatch doesn't run on the UI, so this is now part of the core tests.

Issue #326, opened 2022-12-01
README: What are the install requirements?


benjaoming:
Am very excited to be trying out LSP4E as a user. I'm not a Java developer, but I'm a big Eclipse user anyways (we do exist!).

I'm trying to install LSP4E to the LiClipse 9.0.1, but I have the following error:

```
Cannot complete the install because some dependencies are not satisfiable
  Software being installed: a.jre.javase 14.0.0
  Software being installed: Language Server Protocol client for Eclipse IDE (Incubation) 0.15.0.202211292024 (org.eclipse.lsp4e 0.15.0.202211292024)
  Cannot satisfy dependency:
    From: Language Server Protocol client for Eclipse IDE (Incubation) 0.15.0.202211292024 (org.eclipse.lsp4e 0.15.0.202211292024)
    To: osgi.ee; (&(osgi.ee=JavaSE)(version=17))
```

I originally used the default `openjdk 11.0.17 2022-10-18` on Ubuntu. I've since tried to add openjdk-17-jdk, but that didn't fix the issue.

Do I need Java SE from Oracle to run the extension?

Is it possible that this requirement is explained and made more explicit in the README?

mickaelistria:
The requirement is Java 17.
But the stange part is "Software being installed: a.jre.javase 14.0.0"  which seems to reference something else.
These days, it's easier to embed the Java runtime in the installation by installing Eclipse JustJ in your IDE: use https://download.eclipse.org/justj/jres/latest/ in your Install Software dialog and install it first, it may resolve all other issues.

benjaoming:
Thanks @mickaelistria! It turns out that LiClipse runs with its own bundled JRE version 14. So that's probably why installing a new system JRE didn't change anything.

I tried to install "JustJ Adoptium OpenJDK Hotspot JRE Base" version 17 from https://download.eclipse.org/justj/jres/17/updates/release but that doesn't work well with LiClipse, since it makes it unable to start.

```
java.lang.RuntimeException: Application "com.brainwy.liclipse.rcp.application" could not be found in the registry.
```

Since LiClipse is a bundled version of Eclipse, I've asked them to consider bundling LSP4E.

sebthom:
The issue seems resolved and not a problem in LSP4E. Closing the issue for now. Feel free to reopen.

PR #327, opened 2022-12-01
Update README.md

benjaoming:
This is a very drafty proposal to address #326 

I have no idea whatsoever about what the normal procedures are. As a non-Java dev, I found it hard to understand what exactly is required to run LSP4E in Eclipse.


mickaelistria:
It looks like most of the problems you face are because of the Liclipse bundle adding some weirdness. LSP4E is not meant to be installed by end-users but more as a framework for Plugin developers to bind their language server into Eclipse IDE. So I don't think the requirements need to be documented here; that target audience (plugin developers) do know very well about how to detect the minimal Java version.

Issue #337, opened 2022-12-09
LSP4E client says that code resolver for codeAction is supported but it is not true


angelozerr:
You can see teh problem with WWD

Create a file XML like this:

```xml
<foo></foo>
```

Hover the foo root element, teh following code actions appear:

![image](https://user-images.githubusercontent.com/1932211/206766737-27bf91e0-5eb9-44c8-9d3f-401b7811415e.png)

If you click on it, it does nothing. It is because lemminx version embedded by WWD supports now resolve code action.

LSP4E says that it support resolve code action, but it is not implemented.

PR #340, opened 2022-12-13
CodeAction MarkerResolution support resolveCodeAction

mickaelistria:
Fixes https://github.com/eclipse/lsp4e/issues/337


mickaelistria:
@angelozerr Would you mind testing it?

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/codeactions/CodeActionMarkerResolution.java
```diff
@@ -55,6 +64,18 @@ public String getLabel() {
 
 	@Override
 	public void run(IMarker marker) {
+		if (codeAction.getEdit() == null) {
+			String languageServerId = marker.getAttribute(LSPDiagnosticsToMarkers.LANGUAGE_SERVER_ID, (String)null);
+			LanguageServerDefinition definition = languageServerId != null ? LanguageServersRegistry.getInstance().getDefinition(languageServerId) : null;
```
angelozerr, line 69: I think the resolve code action must be done with the same languageserver instance which have created the codeaction.
I will set this instance in the CodeActionMarkerResolution constructor.

The another thing which is important is to check if server has the capability to resolve codeaction.

rubenporras: Yes, I agree with both open points, they should be done.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/codeactions/CodeActionMarkerResolution.java
```diff
@@ -55,6 +64,18 @@ public String getLabel() {
 
 	@Override
 	public void run(IMarker marker) {
+		if (codeAction.getEdit() == null) {
+			String languageServerId = marker.getAttribute(LSPDiagnosticsToMarkers.LANGUAGE_SERVER_ID, (String)null);
```
rubenporras, line 68:
```suggestion
			String languageServerId = marker.getAttribute(LSPDiagnosticsToMarkers.LANGUAGE_SERVER_ID);
```

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/codeactions/CodeActionMarkerResolution.java
```diff
@@ -55,6 +64,18 @@ public String getLabel() {
...
+			LanguageServerDefinition definition = languageServerId != null ? LanguageServersRegistry.getInstance().getDefinition(languageServerId) : null;
+			try {
+				CompletableFuture<LanguageServer> ls = definition != null ? LanguageServiceAccessor.getInitializedLanguageServer(marker.getResource(), definition, cap -> true) : null;
```
rubenporras, line 71: I wonder if we should add a method
	public static CompletableFuture<LanguageServer> getInitializedLanguageServer(@NonNull LanguageServerDefinition lsDefinition, Predicate<ServerCapabilities> capabilitiesPredicate)
			throws IOException
that does not need a IFile or IResource, in the end we already know which LS we need, so the resource is not really going to change the server we are looking for.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/codeactions/LSPCodeActionsMenu.java
```diff
@@ -113,31 +113,15 @@ public IStatus runInUIThread(IProgressMonitor monitor) {
...
-											public void widgetSelected(SelectionEvent e) {
-												if (codeAction.getEdit() != null) {
-													LSPEclipseUtils.applyWorkspaceEdit(codeAction.getEdit(), codeAction.getTitle());
-												}
-												if (codeAction.getCommand() != null) {
-													executeCommand(info, codeAction.getCommand());
-												}
-											}
-										});
-									}
```
rubenporras, line 140: I do not understand why we can change this code as part of the PR. What if the server does not support resolveCodeAction?

mickaelistria: > I do not understand why we can change this code as part of the PR

It's just some better factorization.

>  What if the server does not support resolveCodeAction?

I believe this case is already handled in CodeActionCompletionProposal, so the factorization fixes it.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/codeactions/CodeActionMarkerResolution.java
mickaelistria: > public static CompletableFuture getInitializedLanguageServer(@nonnull LanguageServerDefinition lsDefinition, Predicate capabilitiesPredicate)

There can be multiple instance of a language server for a given LanguageServerDefinition; not all language servers set "singleton" nor support workapceFolders. So the resource matters.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/codeactions/CodeActionMarkerResolution.java
mickaelistria: > I think the resolve code action must be done with the same languageserver instance which have created the codeaction. I will set this instance in the CodeActionMarkerResolution constructor.

OK

> The another thing which is important is to check if server has the capability to resolve codeaction.

OK

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/codeactions/CodeActionMarkerResolution.java
mickaelistria: This forces to cast back to String.

angelozerr:
@mickaelistria I tested with LemMinx and it doesn't work. I created a little PR at https://github.com/eclipse/lsp4e/pull/343 with the fix

Issue #341, opened 2022-12-13
logging only logs byte array toString() at this time (using WWD)


mickaelistria:
### Discussed in https://github.com/eclipse/lsp4e/discussions/339

<div type='discussions-op-text'>

<sup>Originally posted by **jcompagner** December 13, 2022</sup>
I think LSP4E is doing this kind of logging right? or is it something that WWD provides?

But if i have enable all logging and now with the latest release i only have:

2022-12-12T11:35:19.875849+01:00] LSP4E to org.eclipse.wildwebdeveloper.angular:
[B@2c7d716a
[2022-12-12T11:35:19.876360300+01:00] LSP4E to org.eclipse.wildwebdeveloper.angular:
[B@6d407b51

and that is the case in all the log files i have for example eslint:

[2022-12-09T12:37:41.681577900+01:00] LSP4E to org.eclipse.wildwebdeveloper.eslint:
[B@6904bc4c
[2022-12-09T12:37:41.682609200+01:00] LSP4E to org.eclipse.wildwebdeveloper.eslint:
[B@58a736e8
[2022-12-09T12:37:41.764017700+01:00] LSP4E to org.eclipse.wildwebdeveloper.eslint:

so this is a bit useless

And i wanted to see why i still have in the latest release of WWD now and then a hanging startup (Angular language server startup processing the tsconfig.json file that never ends)
</div>

mickaelistria:
@jcompagner Thanks for the report. Would you be interested in submitting a PR that fixes it?

rubenporras:
I think I broke it, I am looking into creating a PR with the fix.

PR #342, opened 2022-12-14
Fix regression introduced in dab2eb290eb860f6086e7cb090cdcfdfeb7dfb7d

rubenporras:


rubenporras:
for https://github.com/eclipse/lsp4e/issues/341

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LoggingStreamConnectionProviderProxy.java
```diff
@@ -120,7 +120,7 @@ private String message(String direction, byte[] payload) {
 		builder.append(' ');
 		builder.append(id);
 		builder.append(":\n"); //$NON-NLS-1$
-		builder.append(payload);
+		builder.append(new String(payload));
```
jonahgraham, line 123: I think you should add explicit encoding (probably UTF-8\*) so that users who are using regional charset by default don't end up with corrupted messages, or worse unhandled exceptions thrown here.

Note that the call to getBytes() below has a similar issue.

\* Note that LSP support the encoding header and LSP4J support that too so if you are intercepting the raw stream that can cause issues if you don't also read that header. In practice I have no idea if anyone uses anything other than UTF-8, I hope they don't.

PR #343, opened 2022-12-14
Support left of CodeActionProvider

angelozerr:
Support left of CodeActionProvider


rubenporras (approved):
thanks for the fix

PR #344, opened 2022-12-14
Concurrency guarantees - new LS access api

ahmedneilhussain:
This is a new API for LSP4e that is intended to expose access to the asynchronous LSP api in a way that makes it easier to give strong guarantees that the server will receive requests and see document updates in the same order that things happen on the client.

Unfortunately it has proved impractical to provide these strong guarantees using an approach that is nearly identical to the old one, but the new model tries to mitigate the changes by making common use-cases convenient with a higher-level API that takes care of the complexities of handling responses from multiple servers for a single document.

This PR takes the core work that was proposed in draft form in #251 and #333


ahmedneilhussain:
@mickaelistria , @rubenporras this PR is the essence of the proposed API from the earlier PRs. I have based off of recent master and just cherry picked and fixed/tidied up the relevant bits. I've tried to apply the comments you kindly supplied for the earlier PRs.

This has the core API but has not ported any of the examples to actually use it. Nor does it include any of the support for optimistic locking that we have discussed as a related issue. If we can get the API merged in then existing code can be ported across in small, independent PRs.

Tasks I will look to complete tomorrow as part of this PR:
* Add some unit tests, including porting across the two torture tests
* Add deprecation notices to the legacy API

Apologies that some unit tests got caught up when I used the eclipse refactoring to extract `ILSWrapper` from `LanguageServerWrapper`. I'll try and sort this out but I've had enough for today!

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServerWrapper.java
```diff
@@ -774,12 +787,87 @@ protected IStatus run(IProgressMonitor monitor) {
...
+				}
+			}
+			return CompletableFuture.completedFuture(null);
+		}).thenApply(server -> server == null ? null : this);
 	}
 
+	@Override
 	/**
 	 * Warning: this is a long running operation
 	 *
```
rubenporras, line 873: we should move the comment to the interface, shouldn't we? The same applies other methods in this file with now an @Override tag

org.eclipse.lsp4e/src/org/eclipse/lsp4e/ILSWrapper.java
```diff
@@ -0,0 +1,66 @@
...
+public interface ILSWrapper {
+
+	/**
+	 * @return The language ID that this wrapper is dealing with if defined in the
+	 *         content type mapping for the language server
+	 */
+	String getLanguageId(IContentType[] contentTypes);
+
+	/**
+	 * Warning: this is a long running operation
```
rubenporras, line 28: Is it really "a long running operation"? Or it is only so until the server has been initialized? If it is as I think, the comment is wrong (I notice you have just copied it)

org.eclipse.lsp4e/src/org/eclipse/lsp4e/ILSWrapper.java
```diff
@@ -0,0 +1,66 @@
...
+import org.eclipse.jdt.annotation.NonNull;
+import org.eclipse.lsp4j.ServerCapabilities;
+import org.eclipse.lsp4j.services.LanguageServer;
+
+/**
+ * Represents a running language server, with access to the server's capabilities and
+ * methods to dispatch requests/notifications.
+ *
+ */
+public interface ILSWrapper {
```
rubenporras, line 19: For consistency, I think we should name this `ILanguageServerWrapper`, without abbreviation.

rubenporras (commented):
Thanks for the PR. 
I will try to do the real review tomorrow. Today I only had time to have a first glance.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/ILSWrapper.java
ahmedneilhussain: Sure, if that's the consensus. I didn't go for the obvious name because this isn't *necessarily* tied to just exposing bits of `LanguageServerWrapper` - this could be a different class altogether that wraps/delegates to the existing code. I toyed with the idea of a `SingleServerExecutor` object, but that's maybe overkill. So long as there's a clear distinction between running requests on a LS and doing harmless things like querying for capabilities, and doing more low-level lifecycle-related stuff like explicitly starting and stopping them, which I want to keep as separate specialist code.

rubenporras: maybe `IServerExecutor`? I would be fine with that or other name, but it looks to me that we do not use abbreviations for java classes, which is an usual java convention, so I would just avoid `ILSWrapper`.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/ILSWrapper.java
ahmedneilhussain: Interesting point: it's only a (potentially) long-running operation on a wrapper for a server that's never been started. Once started, the server capabilities are cached and so are henceforth immediately available even if it shuts down due to an idle time-out.

So the comment is correct on the raw `LanguageServerWrapper` method, but in any circumstances in which you are handed an instance of this interface by e.g. `LSPExecutors.computeFirst` the initialisation will have been guaranteed to have completed.

Thanks, I hadn't thought that through properly!

org.eclipse.lsp4e/src/org/eclipse/lsp4e/DocumentContentSynchronizer.java
```diff
@@ -270,7 +237,7 @@ public void documentAboutToBeSaved() {
 
 
 		try {
-			List<TextEdit> edits = executeOnCurrentVersionAsync(ls -> ls.getTextDocumentService().willSaveWaitUntil(params))
+			List<TextEdit> edits = languageServerWrapper.executeImpl(ls -> ls.getTextDocumentService().willSaveWaitUntil(params))
```
rubenporras, line 240: why not calling `languageServerWrapper.execute`? The javadoc of `executeImpl` says
 "Internal hook for the LSPexecutor implementations", which does not hold for this case.

mickaelistria (commented):
I'm doing a review commit per commit, and have left a few comments for consideration before we merge the 1st one.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/DocumentContentSynchronizer.java
```diff
@@ -116,36 +113,37 @@ public DocumentContentSynchronizer(@NonNull LanguageServerWrapper languageServer
...
+//			Function<LanguageServer, ? extends CompletionStage<U>> fn) {
+//		AtomicReference<CompletableFuture<U>> resValueFuture = new AtomicReference<>();
+//		lastChangeFuture.updateAndGet(f -> {
+//			CompletableFuture<U> valueFuture = f.thenComposeAsync(fn);
+//			resValueFuture.set(valueFuture);
+//			// We ignore any exceptions that happen when executing the given future
+//			return valueFuture.handle((value, error) -> this.languageServer);
+//		});
+//		return resValueFuture.get();
+//	}
```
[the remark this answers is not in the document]
ahmedneilhussain: D'Oh. Yes. Sure. **sorry**

So much for my porting the changes over from the draft branch so that things would be cleaner...

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServerWrapper.java
```diff
@@ -189,6 +193,8 @@ public void dirtyStateChanged(IFileBuffer buffer, boolean isDirty) {
 	private Timer timer;
 	private AtomicBoolean stopping = new AtomicBoolean(false);
 
+	private final ExecutorService dispatcher;
```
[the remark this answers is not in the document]
ahmedneilhussain: Unfortunately I think that'd probably be of little use or even counterproductive: if people want the benefits/guarantees they need to use the new API so that the dispatcher is used *correctly*. If we expose the dispatcher then we risk it being abused in dependent plugins over which we have little control. Direct access would allow people to tie the thread up with long-running unrelated work or even cause a deadlock.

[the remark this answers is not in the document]
mickaelistria: ok

ahmedneilhussain:
> I'm doing a review commit per commit, and have left a few comments for consideration before we merge the 1st one.

Thanks @mickaelistria can you also take a look at @rubenporras version of this approach in #347 ? Most of what @rubenporras has done is just some nice tidying up, but his approach to replace the `BiFunction` methods on the executor and get rid of `ILanguageServerWrapper` is a critical difference. It's a difference of taste and style rather than of correctness so I cannot argue convincingly for my way over his - it's one for the community to take a collective view, I think!

mickaelistria:
Yes, I'll look at #347 as well. I'd just like to get there step by step and merge increments one after the other to better focus further discussions where necessary.

ahmedneilhussain:
> Yes, I'll look at #347 as well. I'd just like to get there step by step and merge increments one after the other to better focus further discussions where necessary.

Thanks @mickaelistria - #347 supersedes this one, so basically we need to pick whichever approach we prefer and then fix that up ready for merge.

Thanks for reviewing.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LSPExecutor.java
```diff
@@ -0,0 +1,265 @@
...
+import org.eclipse.jface.text.IDocument;
+import org.eclipse.lsp4j.ServerCapabilities;
+import org.eclipse.lsp4j.services.LanguageServer;
+
+/**
+ *
+ */
+public abstract class LSPExecutor<E extends LSPExecutor<E>> {
+
+	/**
```
mickaelistria, line 26: Could all those methods be part of LanguageServers ? At the moment, LanguageServers is only containing static entry point; I imagine it could also contain those methods.

So `LanguageServers.forDocument()` would return an instance of LanguageServers, on which one could call eg `collectAll`.

Eg
```java
LanguageServers.forDocument(doc).collectAll(ls ls.completion(param))
  .thenApplyAsync(lsResponse -> toIDECompletionItems(lsResponse))
  .thenAcceptAsync(uiModel -> showInUI(uiModel), display);
```
(note that `Display` is now an Executor so it gets easier to pass it to futures)

I'm under the impression things get slightly easier if we don't have distinct LanguageServers and LSPExecutor.

rubenporras:
Hi @mickaelistria ,

in case it helps, I have rebased now on master https://github.com/eclipse/lsp4e/pull/347, which would be my proposal to improve on this PR. Of course the "improve" part is subjective but I think it addresses many of the points you brought up on this PR. Would you like to take a look and see if we could merge it?

Regards

org.eclipse.lsp4e/src/org/eclipse/lsp4e/DocumentContentSynchronizer.java
ahmedneilhussain: Well, `DocumentContentSynchronizer` is 'trusted code' for the purposes of using the internal API. The difference between the internal and external is just the extra `.andThenAsync(Function.identity())` designed to make sure the general public doesn't tie up listener threads by composing lots of extra work to the returned future. Since we're just doing a get (and indeed a timeout!) here there is no need for the extra async, and indeed we want the result as quickly as possible without any potential scheduling delay in delivering it via a pool thread.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/ILSWrapper.java
ahmedneilhussain: Renamed to ILanguageServerWrapper

org.eclipse.lsp4e/src/org/eclipse/lsp4e/ILSWrapper.java
ahmedneilhussain: I have clarified the comment

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LSPExecutor.java
ahmedneilhussain: I've consolidated `LSPExecutor` onto `LanguageServers` as you suggested.

ahmedneilhussain:
Hi @mickaelistria I have now also rebased and hopefully addressed all the major review points. I see @rubenporras has also tidied up his version. Hopefully we can agree on one or the other and get whichever one merged in that we prefer. 

I think either one can be merged wholesale rather than piecemeal - since we haven't ported any code across to use the new API there is minimal interaction with the existing codebase and so no real advantage to an incremental approach. That wasn't true for the commit you previously cherry-picked, as that introduced the dispatch thread and switched `DocumentContentSynchronizer` across to use it.

Can potentially squash this down into one or two commits if that would help.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LSPExecutor.java
ahmedneilhussain: Have followed your suggestion and moved all code into a single `LanguageServers` class.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServiceAccessor.java
```diff
@@ -584,6 +583,32 @@ private static Collection<LanguageServerWrapper> getMatchingStartedWrappers(@Non
...
+	 * @param request
+	 * @param onlyActiveLS
+	 * @return
+	 */
+	@NonNull
+	static List<@NonNull CompletableFuture<LanguageServerWrapper>> getWrappers(@Nullable IProject project, Predicate<ServerCapabilities> request, boolean onlyActiveLS) {
+		List<@NonNull CompletableFuture<LanguageServerWrapper>> result = new ArrayList<>();
+		for (LanguageServerWrapper wrapper : startedServers) {
+			if ((!onlyActiveLS || wrapper.isActive()) && (project == null || wrapper.canOperate(project))) {
+				if (capabilitiesComply(wrapper, request)) {
```
rubenporras, line 604: This part which is common between this method and getLanguageServers can be factored out. Then this method can go to `LanguageServers`. I will create a PR that we can merge standalone to make this PR smaller.

ahmedneilhussain: Thanks @rubenporras

org.eclipse.lsp4e/src/org/eclipse/lsp4e/ILanguageServerWrapper.java
```diff
@@ -0,0 +1,79 @@
...
+import org.eclipse.jdt.annotation.NonNull;
+import org.eclipse.lsp4j.ServerCapabilities;
+import org.eclipse.lsp4j.services.LanguageServer;
+
+/**
+ * Represents a running language server, with access to the server's capabilities and
+ * methods to dispatch requests/notifications.
+ *
+ */
+public interface ILanguageServerWrapper {
```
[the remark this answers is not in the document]
ahmedneilhussain: OK, done! I'm going to go with the consensus on this one ;-)

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServiceAccessor.java
rubenporras: You could remove this method now by rebasing and moving the specific part to LanguageServers. See my method in the LanguageServerExecutor in my PR:
 
```
		@Override
		protected @NonNull List<@NonNull CompletableFuture<@Nullable LanguageServerWrapper>> getServers() {
			// Compute list of servers from project & filter
			List<@NonNull CompletableFuture<LanguageServerWrapper>> wrappers = new ArrayList<>();
			for (LanguageServerWrapper wrapper :  LanguageServiceAccessor.getStartedWrappers(project, getFilter(), !restartStopped)) {
				wrappers.add(wrapper.getInitializedServer().thenApply(ls -> wrapper));
			}
			return wrappers;
		}
```

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServers.java
```diff
@@ -0,0 +1,344 @@
...
+import org.eclipse.jdt.annotation.Nullable;
+import org.eclipse.jface.text.IDocument;
+import org.eclipse.lsp4j.ServerCapabilities;
+import org.eclipse.lsp4j.services.LanguageServer;
+
+/**
+ * Main entry point for accessors to run requests on the language servers, and some utilities
+ * for manipulating the asynchronous response objects in streams
+ */
+public abstract class LanguageServers<E extends LanguageServers<E>> {
```
rubenporras, line 39: what about naming this LanguageServerExecutor?

ahmedneilhussain: If you remember, I originally had `LSPExecutor` as the abstract class with the logic, and `LanguageServers` as the API entry point with the static factory methods, but @mickaelistria had a preference for just having one class called `LanguageServers`...

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServiceAccessor.java
ahmedneilhussain: Thanks - have done as you suggested.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServers.java
rubenporras: Okay, I have forgotten than. I will rename the class in my PR then.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServerWrapper.java
```diff
@@ -827,8 +854,8 @@ void notifyOnLatestVersion(Consumer<LanguageServer> fn) {
 	 * @return Async result that guarantees the wrapped server will be active and connected to the document. Wraps
 	 * null if the server does not support the requested capabilities or could not be started.
 	 */
-	CompletableFuture<LanguageServerWrapper> connectIf(IDocument document, Predicate<ServerCapabilities> filter) {
-		return getInitializedServer().thenComposeAsync(server -> {
+	@NonNull CompletableFuture<@Nullable LanguageServerWrapper> connectIf(@NonNull IDocument document, @NonNull Predicate<ServerCapabilities> filter) {
```
rubenporras, line 857: this method is not used any longer after this PR, is it?

ahmedneilhussain: Yes it is! It's critical to the `LSPDocumentExecutor.getServers()` implementation!

rubenporras: Sorry, I have overlooked that. Then, in order to make this PR smaller, can the changes to `LSW#connectIf` be merged as an standonle PR? I am not sure why we do the changes, so for this one I will not split it to another PR myself

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServerWrapper.java
```diff
@@ -797,23 +797,50 @@ protected IStatus run(IProgressMonitor monitor) {
...
 	 * @return Async result
 	 */
-	<T> CompletableFuture<T> executeOnLatestVersion(Function<LanguageServer, ? extends CompletionStage<T>> fn) {
-		return getInitializedServer().thenComposeAsync(fn, this.dispatcher);
+	public <T> CompletableFuture<T> execute(@NonNull Function<LanguageServer, ? extends CompletionStage<T>> fn) {
+		// Send the request on the dispatch thread, then additionally make sure the response is delivered
+		// on a thread from the default ForkJoinPool. This makes sure the user can't chain on an arbitrary
+		// long-running block of code that would tie up the server response listener and prevent any more
+		// inbound messages being read
+		return executeImpl(fn).thenApplyAsync(Function.identity());
```
mickaelistria, line 824: Good!

mickaelistria (approved):
I like this current proposal is good to be merged. It's relatively small and easy to read, not really adding complexity over current state and gives good guarantees.
It's a good showcase of the saying "Easy solutions are often the result of a hard work".

@rubenporras What do you think?

rubenporras:
HI @mickaelistria, I would be happy as well if we merge this one. I also think it it is good work. There some things that I would do different (as in  https://github.com/eclipse/lsp4e/pull/347) but I think it is best if we merge this one, and I rebase my PR on top, then we can take a look and think if it is an improvement or not.

ahmedneilhussain:
Thanks @mickaelistria @rubenporras .
I'll take the example code that I ported to the new API and submit those as individual PRs.

mickaelistria:
> Thanks @mickaelistria @rubenporras .

Well, thank YOU first ;)

> I'll take the example code that I ported to the new API and submit those as individual PRs.

Great, thanks in advance!

PR #347, opened 2022-12-20
Improved proposal based on Pr 344

rubenporras:
I have worked on #344 and have prepared this proposal (it is an improvement for me) and would be happy if we would merge this or something similar.
The main differences to #344 is that I have extended the LSW so that it can be used to get services from the `LS`, that eliminates the need of having APIs with `(LanguageServerWrapper)` and `(LanguageServerWrapper, LanguageServer)`. I have added copyright headers and improved small things. 
Finally I have added one `LanguageServerExecutor` when the entry point is an existing `(LanguageServerWrapper)`.
I have also removed the TODOs.

### DIFF
--- a/org.eclipse.lsp4e.test/src/org/eclipse/lsp4e/test/LanguageServersTest.java
+++ b/org.eclipse.lsp4e.test/src/org/eclipse/lsp4e/test/LanguageServersTest.java
@@ -702,17 +702,17 @@ public class LanguageServersTest {
 
 		CompletableFuture<List<LSWPair>> async = LanguageServers.forDocument(document)
 				.withFilter(capabilities -> LSPEclipseUtils.hasCapability(capabilities.getHoverProvider()))
-				.collectAll((w, ls) -> ls.getTextDocumentService().hover(params).thenApply(h -> new LSWPair(w, ls)));
+				.collectAll(w -> w.getTextDocumentService().hover(params).thenApply(h -> new LSWPair(w, w.getServer())));
 
 		final List<LSWPair> result = async.join();
 
 		final AtomicInteger matching = new AtomicInteger();
 
 		assertEquals("Should have had two responses", 2, result.size());
-		assertNotEquals("LS should have been different proxies", result.get(0).server, result.get(1).server);
+		assertNotEquals("LS should have been different proxies", result.get(0).wrapper, result.get(1).wrapper);
 		result.forEach(p -> {
 			p.wrapper.execute(ls -> {
-				if (ls == p.server) {
+				if (ls == p.wrapper.getServer()) {
 					matching.incrementAndGet();
 				}
 				return CompletableFuture.completedFuture(null);
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/DocumentContentSynchronizer.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/DocumentContentSynchronizer.java
@@ -234,7 +234,7 @@ final class DocumentContentSynchronizer implements IDocumentListener {
 
 
 		try {
-			List<TextEdit> edits = languageServerWrapper.executeImpl(ls -> ls.getTextDocumentService().willSaveWaitUntil(params))
+			List<TextEdit> edits = languageServerWrapper.execute(ls -> ls.getTextDocumentService().willSaveWaitUntil(params))
 				.get(lsToWillSaveWaitUntilTimeout(), TimeUnit.SECONDS);
 			try {
 				LSPEclipseUtils.applyEdits(document, edits);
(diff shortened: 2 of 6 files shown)


rubenporras:
@ahmedneilhussain , @mickaelistria , I hope this is an improvement on #344  and that it helps getting a new API into master so that we can rework the consumers. Feel free to use it for further improvements.
I will be on holidays from the 24th of December to the 7th of January. I will catch up then whit whatever version is then on review or merged :)

ahmedneilhussain:
I understand why `ILanguageServerWrapper` and the bifunction overloads were unappealing. However part of the motivation for the change is to discourage users from accessing the language server directly, because then it's very hard to get the concurrency behaviour correct - part of the problem with the existing API was that it was too low-level.

If the user does something like 

```
executor.computeAll(wrapper -> wrapper.getDocumentService().hover(params).thenCompose(hoverResponse - new HoverObject(hoverResponse, wrapper));
```
and `HoverObject` at some subsequent point does something like 
```
this.wrapper.execute({ some more work });
```
then that's fine.

If they then do
```
this.wrapper.getTextDocumentService().hover(params);
```
then they will completely bypass the single-threaded dispatch queue.

ahmedneilhussain:
I guess it's a matter of taste how far an API should encourage and make it easy for users to Do The Right Thing, and how far it should enforce it. Everything has to get PR-approved so I guess we could just put dire warnings about not using the methods directly outside the executor framework on `LanguageServerWrapper.getTextDocumentServer` and its sibling methods, and not approve any contributions that violate that rule.

ahmedneilhussain:
Note that we would need some method of getting access to a client-specific subclass of `LanguageServer` so that they could make use of the custom extensions. I suppose this could be added to `LanguageServerWrapper` along the lines

```
<T extends LanguageServer> Optional<T> as(Class<T> class) {
  if (this.languageServer instanceOf T) {
    return Optional.of((T)this.languageServer);
  }
  return Optional.empty();
}
```

ahmedneilhussain:
I haven't gone through everything exhaustively but most of the tidy-up rearrangement you've done looks nice and sensible to me, @rubenporras .

rubenporras:
> I guess it's a matter of taste how far an API should encourage and make it easy for users to Do The Right Thing, and how far it should enforce it. Everything has to get PR-approved so I guess we could just put dire warnings about not using the methods directly outside the executor framework on `LanguageServerWrapper.getTextDocumentServer` and its sibling methods, and not approve any contributions that violate that rule.

Yes, it is indeed not black and white what is better and what is worse. I think that we already have ways to bypass the executor framework with the single dispatcher, that is why we are doing this API :) So I prefer this API to the one you proposed originally, but it is in the end mostly a matter of taste. I like more my proposal here but would be happy with the other one as well.

rubenporras:
> Note that we would need some method of getting access to a client-specific subclass of `LanguageServer` so that they could make use of the custom extensions. I suppose this could be added to `LanguageServerWrapper` along the lines
> 
> ```
> <T extends LanguageServer> Optional<T> as(Class<T> class) {
>   if (this.languageServer instanceOf T) {
>     return Optional.of((T)this.languageServer);
>   }
>   return Optional.empty();
> }
> ```

Indeed, that is a good point. Indeed, I still prefer what I propose here plus a method such as `as`, because it makes coding for the normal case seamless, but we could not implement `LanguageServer` in `LanguageServerWrapper` as I did and instead  just offer a method `unwrap`, that would cover all cases because the caller could do the instanceof/cast.

ahmedneilhussain:
> > I guess it's a matter of taste how far an API should encourage and make it easy for users to Do The Right Thing, and how far it should enforce it. Everything has to get PR-approved so I guess we could just put dire warnings about not using the methods directly outside the executor framework on `LanguageServerWrapper.getTextDocumentServer` and its sibling methods, and not approve any contributions that violate that rule.
> 
> Yes, it is indeed not black and white what is better and what is worse. I think that we already have ways to bypass the executor framework with the single dispatcher, that is why we are doing this API :) So I prefer this API to the one you proposed originally, but it is in the end mostly a matter of taste. I like more my proposal here but would be happy with the other one as well.

@rubenporras one thing we could do is make LanguageServerWrapper.languageServer a thread local which is only set on the dispatch thread. Thus any attempt to call LanguageServerWrapper.getDocumentService() outside of a the executor framework would fail.

Might be going a bit too far down the bondage-and-discipline route. And still wouldn't prevent a user from getting the document service inside a function passed into an executor method, storing it as a field inside some object, and then (ab)using it further down the line.

mickaelistria:
I'm quite skeptical about `LanguageServerWrapper extends LanguageServer`; I don't see it as a simplificaion and I'm afraid it actually adds some possible confusion for users about what object to use when and how. Generally, Design Patterns rule state that OO works better if we "prefer composition over inheritance" as it leaves more freedom to both ends and allow to properly split the duties and roles.
Here, it's been made clear that we need some explicit pattern to ensure good orchestration, and that direct access to LanguageServer breaks the required pattern; I see this change as a step in the opposite direction by making some patterns seem optional, or allowing more easily some constructs that wouldn't obey the required orchestration requirements (eg one can now use LanguageServerWrapper, and then use multiple operations on it, bypassing the useful execute method).

PR #349, opened 2022-12-22
Fix encoding

rubenporras:
Use UTF-8 encoding explicitly so that users who are using regional
charset by default don't end up with corrupted messages, or worse
unhandled exceptions thrown here.

According to the LSP specification:
"The content part is encoded using the charset provided in the
Content-Type field. It defaults to UTF-8, which is the only encoding
supported right now. If a server or client receives a header with a
different encoding than UTF-8 it should respond with an error."

So even though the code in LSP4J reads the Content-Type header and
evaluates it, and it could thus supports different encoding, LSP4E does
not need to support it, and in practice, no server should use that
either.


mickaelistria:
Thanks!

PR #357, opened 2023-01-12
Fix possible freeze in SemanticHighlightReconcilerStrategy

vrubezhny:
Getting rid of infinite waiting in SemanticHighlightReconcilerStrategy

### DIFF
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/semanticTokens/SemanticHighlightReconcilerStrategy.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/semanticTokens/SemanticHighlightReconcilerStrategy.java
@@ -14,6 +14,8 @@ import java.util.List;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionException;
 import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
 import java.util.function.Function;
 
 import org.eclipse.core.runtime.IProgressMonitor;
@@ -81,6 +83,8 @@ import org.eclipse.swt.custom.StyledText;
  */
 public class SemanticHighlightReconcilerStrategy
 		implements IReconcilingStrategy, IReconcilingStrategyExtension, ITextPresentationListener {
+
+    private static final int FULL_RECONCILE_TIMEOUT_S = 10;
 
     private final boolean disabled;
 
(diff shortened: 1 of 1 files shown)


org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/semanticTokens/SemanticHighlightReconcilerStrategy.java
```diff
@@ -82,6 +84,8 @@
 public class SemanticHighlightReconcilerStrategy
 		implements IReconcilingStrategy, IReconcilingStrategyExtension, ITextPresentationListener {
 
+    private static final int FULL_RECONCIL_TIMEOUT_S = 10;
```
rubenporras, line 87: should this not be `FULL_RECONCILE_TIMEOUT_S`?

rubenporras:
I guess it is. I expect that in practice by the time 10 seconds have passed the request should have been answered or the user has typed again (and the the requests has been cancelled).

rubenporras:
> @rubenporras Yet another question: should we interrupt current thread in case of `InterruptedException` - https://github.com/eclipse/lsp4e/blob/master/org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/semanticTokens/SemanticHighlightReconcilerStrategy.java#L298 by changing the catch block to:
> 
> ```
> 			} catch (ExecutionException e) {
> 				LanguageServerPlugin.logError(e);
> 			} catch (InterruptedException) {
> 				LanguageServerPlugin.logError(e);
> 				Thread.currentThread().interrupt();
> 			}
> ```
> 
> * as it's usually done in hovers, content assist, etc. - or we need this thread to stay uninterrupted?

It was an oversight not to add `Thread.currentThread().interrupt()`, so let us add it (even thought it is not clear what is the advantage of doing so, I think we should follow the pattern we already have)

vrubezhny:
> It was an oversight not to add `Thread.currentThread().interrupt()`, so let us add it (even thought it is not clear what is the advantage of doing so, I think we should follow the pattern we already have)

OK.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/semanticTokens/SemanticHighlightReconcilerStrategy.java
vrubezhny: ACK

rubenporras (approved):

mickaelistria:
Thanks!

Issue #358, opened 2023-01-12
JSON messages going to LS have URIs with non-ASCII chars


BoykoAlex:
A URI to a file named: `ClassWithSpécialCharacter.java` (note non-ASCII `é` in the name) in its string form in the JSON still looks the same without `é` escaped. VSCode LSP client sends this url with file name that looks like: `ClassWithSp%C3%A9cialCharacter.java` which is correct. URIs with unicode chars are illegal... The issue I have faced is `Paths(URI.create(uriStr))` throws `java.lang.IllegalArgumentException: Bad escape`

Test project with such file: https://github.com/spring-projects/sts4/tree/main/headless-services/spring-boot-language-server/src/test/resources/test-projects/test-annotation-indexing-parent/test-annotation-indexing%20with%20space

BoykoAlex:
Should be fixed now. Will see if lsp4j will have anything in addition via https://github.com/eclipse/lsp4j/issues/692

PR #368, opened 2023-01-15
Use fully quoted/encoded URIs via URI.toASCIIString() #358

sebthom:
This PR replaces `URI#toString()` calls with `URI#toASCIIString()` calls, either directly or indirectly by using the new `LSPEClipseUtils#toTextDocumentIdentifier` methods to construct `TextDocumentIdentifier` instances.

This addresses https://github.com/eclipse/lsp4e/issues/358

### DIFF
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/DocumentContentSynchronizer.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/DocumentContentSynchronizer.java
@@ -48,7 +48,6 @@ import org.eclipse.lsp4j.MessageType;
 import org.eclipse.lsp4j.Range;
 import org.eclipse.lsp4j.ServerCapabilities;
 import org.eclipse.lsp4j.TextDocumentContentChangeEvent;
-import org.eclipse.lsp4j.TextDocumentIdentifier;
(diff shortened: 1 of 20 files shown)


mickaelistria:
While this fixes the issue; I'm a bit worried that we kind of require a particular type of URIs for things to work; and that we need to think about this everywhere. I'm certain that since URI is a very popular type, we'll again in the future use URI.toString() somewhere and cause another occurrence of this issue.
Using more TextDocumentIdentifier everywhere in place of URIs seems like a good idea as we can enforce the in TextDocumentIdentifier. Do you think that this should also be fixed in LSP4J if it's defined in the protocol? I imagine LSP4J could take care of of using the ASCIIString of URIs when generating a TextDocumentIdentifier.

sebthom:
I agree, it would be better if LSP4J would normalize the URIs.

BoykoAlex:
@sebthom @mickaelistria I'd go with this PR and in addition raise an issue against lsp4j to check what they can do about it: https://github.com/eclipse/lsp4j/issues/692
What do you think?

sebthom:
I am fine with this.

mickaelistria:
That would be OK with me as well.

Issue #273, opened 2022-10-28
Two CodeMiningReconciler are created for one file


angelozerr:
When you open a file like JSON, I noticed that there are 2 CodeMiningReconciler which are installed which is very annoying since codemining can be expensive.

The reason about those 2 CodeMiningReconciler is that it is declared in generic editor plugin and lsp4e plugin. I think LSP4e should remove the declaration of code mining provider.

sebthom:
@mickaelistria I can confirm that two instances are created. Any reason to not remove the CodeMiningReconciler from LSP4E?

mickaelistria:
> Any reason to not remove the CodeMiningReconciler from LSP4E?

Not that I remember. Removing it sounds like a good idea.

PR #370, opened 2023-01-16
Don't register CodeMiningReconciler #273

sebthom:


mickaelistria:
@sebthom I didn't try it; but if it's working for you, feel free to merge.

sebthom:
In my test the CodeMiningReconciler is still installed into the textviewer but the instantiation of a second object does not happen.

Issue #371, opened 2023-01-16
Dispatcher Threads are leaked


sebthom:
When closing the last file of a project, the associated language server is shutdown.
However two dispatcher threads are not stopped. Repeatably closing/opening the same file then results in more and more dispatcher threads.

Here is an example after opening/closing the same JS file several times.

![image](https://user-images.githubusercontent.com/426959/212719113-b1074cfe-b910-422f-8825-489534562dfa.png)

The LanguageServerWrapper#stopDispatcher dispatcher method is only invoked via LanguageServiceAccessor#shutdownAllDispatchers once the LSP4E bundle gets stopped.

ahmedneilhussain:
Hi - @mickaelistria  I was unaware that this code had been committed to master. I thought we were still waiting to decide between #344 and #347 and then polish whichever one up we decided was the better approach.

Did you cherry pick some bits for master? I wasn't aware...!

ahmedneilhussain:
On closer inspection, it's not the dispatcher thread that is being leaked. It's the *launcher* threads. This is a longstanding issue (which I had spotted, and hadn't got round to mentioning to anyone). This are the threads created at https://github.com/eclipse/lsp4e/blob/4c79b78e52bad00acca50ac3e7d004bf0b43b59a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServerWrapper.java#L291 which are used by the LSP4j to listen to LS responses. At the moment they get started from scratch every time the LSW starts the server and not cleaned up on LSW.shutdown, hence the leak. (Note that although we supply a new 'pool' executor, only one listener thread can ever get started from it by LSP4j).

It's a longstanding issue - if you run the test suite and monitor the threads from eclipse, you will see rather a lot of them by the time the suite finishes...!

ahmedneilhussain:
The dispatcher is intentionally kept alive for the lifetime of the plugin, and reused if we restart the LS. That's by my design (debatably so, but by design nonetheless!). So although if you shutdown the LS and never reuse it from within your eclipse session, you will get an idle dispatcher thread hanging around, it won't be leaked.

ahmedneilhussain:
Cleaning up the launcher threads is probably a good idea and should be fixed, but it wasn't introduced by https://github.com/eclipse/lsp4e/commit/5ea29bf033f6b07c2d9a9ea769ec0d83bde4f219 ...

[...]

PR #373, opened 2023-01-16
Stop dispatcher executor service when LanguageServer stops #371

sebthom:
This PR stops the dispatcher executor when the language server stops. If the language server is started again, a new dispatcher executor will be started.

### DIFF
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServerWrapper.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServerWrapper.java
@@ -194,7 +194,7 @@ public class LanguageServerWrapper {
 	private Timer timer;
 	private AtomicBoolean stopping = new AtomicBoolean(false);
 
-	private final ExecutorService dispatcher;
+	private ExecutorService dispatcher;
 
 	/**
 	 * Map containing unregistration handlers for dynamic capability registrations.
@@ -219,14 +219,23 @@ public class LanguageServerWrapper {
 		this.initialPath = initialPath;
 		this.serverDefinition = serverDefinition;
 		this.connectedDocuments = new HashMap<>();
-		String threadNameFormat = "LS-" + serverDefinition.id + "-dispatcher-%d"; //$NON-NLS-1$ //$NON-NLS-2$
-		this.dispatcher = Executors
-				.newSingleThreadExecutor(new ThreadFactoryBuilder().setNameFormat(threadNameFormat).build());
-
+		startDispatcher();
 	}
 
-	void stopDispatcher() {
-		this.dispatcher.shutdownNow();
+	private synchronized void startDispatcher() {
+		if(dispatcher != null && !dispatcher.isShutdown()) {
+			return;
+		}
+		final String threadNameFormat = "LS-" + serverDefinition.id + "-dispatcher-%d"; //$NON-NLS-1$ //$NON-NLS-2$
+		this.dispatcher = Executors
+				.newSingleThreadExecutor(new ThreadFactoryBuilder().setNameFormat(threadNameFormat).build());
+	}
+
+	synchronized void stopDispatcher() {
+		if(dispatcher == null) {
+			return;
+		}
+		dispatcher.shutdownNow();
 	}
 
 	/**
@@ -269,6 +278,7 @@ public class LanguageServerWrapper {
 				stop();
 			}
 		}
+		startDispatcher();
 		if (this.initializeFuture == null) {
 			final URI rootURI = getRootURI();
 			this.initializeFuture = CompletableFuture.supplyAsync(() -> {
@@ -564,6 +574,8 @@ public class LanguageServerWrapper {
 		this.languageServer = null;
 
 		FileBuffers.getTextFileBufferManager().removeFileBufferListener(fileBufferListener);
+
+		stopDispatcher();
 	}
 
 	/**


sebthom:
I am also wondering if with this change the ´LanguageServiceAccessor.shutdownAllDispatchers();` is still required. If something needs to be explicitly stopped on bundle shutdown wouldn't that be the running language servers instead of only the dispatcher executors anyway?

mickaelistria:
This looks good to me.
@ahmedneilhussain any concern with this patch?

rubenporras:
It also looks good to me.

ahmedneilhussain:
> This looks good to me.
> @ahmedneilhussain any concern with this patch?

I think it's probably OK as far as it goes, but it won't address the leak. As I pointed out on https://github.com/eclipse/lsp4e/issues/371 it's not the dispatcher threads that are leaked, it's the launcher threads.

The dispatcher threads currently do stick around after the LS shuts down, and get reused when it restarts, but there should never be more than one per LS.

I think it's *probably* OK (and even desirable) to scope the dispatcher thread lifetime to that of the LS it manages. I think one reason I didn't do this in the first place was because I toyed initially with having a single dispatcher for *all* LS (which would not have been a good idea, on balance), but also because I was also keeping the startup/shutdown comms (which involves sending messages to the server, obviously) separate from the dispatcher thread, for fear of introducing a subtle deadlock. I think it's probably safe to make this change so long as the startup/shutdown code doesn't actually *use* the dispatcher thread. However I would prefer not to make this change until more of the rest of #344  or #347 is merged in, as it's hard to assess the impact in the absence of most of the code that would use it!

ahmedneilhussain:
> I am also wondering if with this change the `LanguageServiceAccessor.shutdownAllDispatchers();` is still required. If something needs to be explicitly stopped on bundle shutdown wouldn't that be the running language servers instead of only the dispatcher executors anyway?

You are correct that if LS shutdown closes the dispatcher threads then there shouldn't be any need for bundle shutdown logic. However I'd be a bit careful: the single threaded executor threads are not daemon threads and will prevent JVM termination if not explicitly shut down: I found this out the hard way by discovering my changes caused Eclipse to hang on exit until I added this logic...

The pooled executors behave differently and shut down their threads if they have been idle for some time: this is why the launcher threads don't cause an eclipse hang despite the fact that we are leaking them.

I'd be a bit careful before taking this out - the start/stop logic on LanguageServerWrapper is very complicated and probably not as robust to exceptions as it might be (I think we probably ought to be using `finally` more than we are...). I think it's harmless to leave in - it won't do anything if there's nothing to do.

sebthom:
> > I am also wondering if with this change the `LanguageServiceAccessor.shutdownAllDispatchers();` is still required. If something needs to be explicitly stopped on bundle shutdown wouldn't that be the running language servers instead of only the dispatcher executors anyway?
> 
> You are correct that if LS shutdown closes the dispatcher threads then there shouldn't be any need for bundle shutdown logic. However I'd be a bit careful: the single threaded executor threads are not daemon threads and will prevent JVM termination if not explicitly shut down: I found this out the hard way by discovering my changes caused Eclipse to hang on exit until I added this logic...
> 
> The pooled executors behave differently and shut down their threads if they have been idle for some time: this is why the launcher threads don't cause an eclipse hang despite the fact that we are leaking them.
> 
> I'd be a bit careful before taking this out - the start/stop logic on LanguageServerWrapper is very complicated and probably not as robust to exceptions as it might be (I think we probably ought to be using `finally` more than we are...). I think it's harmless to leave in - it won't do anything if there's nothing to do.

I was more wondering about why only stop the dispatchers and not explicitly stop all LanguageServer/LanguageServerWrappers (including the dispatchers)

sebthom:
> I think it's probably OK as far as it goes, but it won't address the leak. 
I don't understand why this does not address the leak. I can see that the threads are now terminated when I close the last file and don't pile up if I repeatedly open/close files as it was before.

> The dispatcher threads currently do stick around after the LS shuts down, and get reused when it restarts, but there should never be more than one per LS.
Then maybe the problem is that LS are not reused?

ahmedneilhussain:
> > I think it's probably OK as far as it goes, but it won't address the leak.
> > I don't understand why this does not address the leak. I can see that the threads are now terminated when I close the last file and don't pile up if I repeatedly open/close files as it was before.
> 
> > The dispatcher threads currently do stick around after the LS shuts down, and get reused when it restarts, but there should never be more than one per LS.
> > Then maybe the problem is that LS are not reused?

Hi - I think we are maybe misunderstanding one another. Can you be clear about *which* threads you are worried about, and which ones are affected by the fix?

If you look at the image in https://github.com/eclipse/lsp4e/issues/371#issue-1535150205 do we agree that there are *two* kinds of threads.

Some end in `-dispatcher` - these are the ones added by me in the change that @mickaelistria committed. They are created by the line at https://github.com/eclipse/lsp4e/blob/4c79b78e52bad00acca50ac3e7d004bf0b43b59a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServerWrapper.java#L223 In my original implementation, there should only be one per LS/Wrapper. After the LS starts, the threads will hang around if the LS times out and shuts down, and be reused if the LS starts up, but won't be 'leaked'. Your fix in this PR changes it so that they get closed after the LS exits.


Some other threads end in `-launcher`. These get created at https://github.com/eclipse/lsp4e/blob/4c79b78e52bad00acca50ac3e7d004bf0b43b59a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServerWrapper.java#L291 and you get a new one every time the LS starts up. They 'leak' as there is nothing explicitly tracking them and shutting them down (I think the thread pools will eventually time out after idle activity). This particularly affects the unit tests as those involve starting and stopping a lot of LS in a very short space of time.

Your PR doesn't address the launcher threads at all, but you can see from the screenshot in the bug report that it is those ones that are accumulating when the LS stops and starts multiple times.

Can you just confirm which threads we are talking about here? They are two sets of unrelated executors!

ahmedneilhussain:
I'm just looking at submitting a PR to fix the `-launcher` thread leak, as I have noticed it before and it is probably a good thing to fix.

ahmedneilhussain:
No problem @sebthom - this isn't an unreasonable change, but maybe one for when the rest of my work is integrated and we have made sure that there aren't any more fundamental issues.

Good that people are looking at the resource management/lifecycle stuff. Can never be too paranoid about that kind of thing...

ahmedneilhussain:
I've created #376 just to try and tidy this sort of stuff up in tests (I don't think it's a serious issue for production)

PR #376, opened 2023-01-17
Improve thread cleanup/recycling in tests

ahmedneilhussain:
This PR attempts to mitigate some of the behaviour noted in https://github.com/eclipse/lsp4e/issues/371

I have picked through the code and checked the documentation on `Executors.newCachedThreadPool()` and established that the behaviour is probably benign as far as production goes: prior to this PR we created a new cached thread pool for each LS every time `LanguageServerWrapper.start()` was called, with no corresponding cleanup code.
However:
1. Only one thread is ever instantiated from each thread pool by the launcher (used for listening for LS responses and incoming server->client requests)
2. As documented [here](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/concurrent/Executors.html) `Executors.newThreadPoolExecutor()` creates threads on demand and shuts them down after 60 seconds of idleness, so unless LS have been configured with very very short idle timeouts, then these threads will close down shortly after `LanguageServerWrapper.close()` is called in any case.

The behaviour was more of an issue in testing, as each test creates a new `LanguageServerWrapper` and up to three threads (the launcher/listener thread, the recently-added [by me] dispatcher thread, and the thread for the mock language server - implicitly created by `MockConnectionProvider`).

I've tried to tidy this up - if you run the full test suites in the debugger in Eclipse you can see hundreds of threads starting. It's possible that in some environments the scheduler overhead of this might be causing some of our test stability issues.

### DIFF
--- a/org.eclipse.lsp4e.test/src/org/eclipse/lsp4e/test/MockConnectionProvider.java
+++ b/org.eclipse.lsp4e.test/src/org/eclipse/lsp4e/test/MockConnectionProvider.java
@@ -21,7 +21,10 @@ import java.nio.channels.Pipe;
 import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
+import java.util.function.Function;
 
 import org.eclipse.lsp4e.server.StreamConnectionProvider;
 import org.eclipse.lsp4e.tests.mock.MockLanguageServer;
@@ -37,6 +40,8 @@ public class MockConnectionProvider implements StreamConnectionProvider {
 	private Future<Void> listener;
 	private Collection<Closeable> streams = new ArrayList<>(4);
 	
+	private static ExecutorService testRunner = Executors.newCachedThreadPool();
+	
 	@Override
 	public void start() throws IOException {
 		Pipe serverOutputToClientInput = Pipe.open();
@@ -46,7 +51,7 @@ public class MockConnectionProvider implements StreamConnectionProvider {
 		InputStream serverInputStream = Channels.newInputStream(clientOutputToServerInput.source());
 		OutputStream serverOutputStream = Channels.newOutputStream(serverOutputToClientInput.sink());
 		Launcher<LanguageClient> launcher = LSPLauncher.createServerLauncher(MockLanguageServer.INSTANCE, serverInputStream,
-				serverOutputStream);
+				serverOutputStream, testRunner, Function.identity());
 		clientInputStream = Channels.newInputStream(serverOutputToClientInput.source());
 		clientOutputStream = Channels.newOutputStream(clientOutputToServerInput.sink());
 		listener = launcher.startListening();
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServerWrapper.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServerWrapper.java
@@ -222,14 +222,14 @@ public class LanguageServerWrapper {
 		this.serverDefinition = serverDefinition;
 		this.connectedDocuments = new HashMap<>();
 		String projectName = (project != null && project.getName() != null) ? ("@" + project.getName()) : "";  //$NON-NLS-1$//$NON-NLS-2$
-		String dispatcherThreadNameFormat = "LS-" + serverDefinition.id + projectName + "-dispatcher-%d"; //$NON-NLS-1$ //$NON-NLS-2$
+		String dispatcherThreadNameFormat = "LS-" + serverDefinition.id + projectName + "#dispatcher"; //$NON-NLS-1$ //$NON-NLS-2$
 		this.dispatcher = Executors
 				.newSingleThreadExecutor(new ThreadFactoryBuilder().setNameFormat(dispatcherThreadNameFormat).build());
 
 		// Executor service passed through to the LSP4j layer when we attempt to start the LS. It will be used
 		// to create a listener that sits on the input stream and processes inbound messages (responses, or server-initiated
 		// requests).
-		String listenerThreadNameFormat = "LS-" + serverDefinition.id + projectName + "-listener-%d"; //$NON-NLS-1$ //$NON-NLS-2$
+		String listenerThreadNameFormat = "LS-" + serverDefinition.id + projectName + "#listener-%d"; //$NON-NLS-1$ //$NON-NLS-2$
 		this.listener = Executors
 				.newCachedThreadPool(new ThreadFactoryBuilder().setNameFormat(listenerThreadNameFormat).build());
 	}
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServerWrapper.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServerWrapper.java
@@ -196,6 +196,8 @@ public class LanguageServerWrapper {
 
 	private final ExecutorService dispatcher;
 
+	private final ExecutorService listener;
+
 	/**
 	 * Map containing unregistration handlers for dynamic capability registrations.
 	 */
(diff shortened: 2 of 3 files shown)


ahmedneilhussain:
I should say in the interests of honesty that having a single thread pool for all instances of the mock server might be a possible source of contention/cross-talk between tests. I don't think so, because:

- it's still a *pool*
-  from which only one thread is used per mock-LS
- and only one mock LS should be active in most tests.
- So there should be no danger of thread exhaustion even if occasionally a mock LS has not finished shutting down by the time the next one starts up.

ahmedneilhussain:
Changes that affect production:
* Rename launcher threads as listener threads to better reflect what they do
* Keep listener thread pool for life of LangauageServerWrapper rather than creating a new one on start()
* Add comments clarifying the above

Changes for test only:
* Add shutdown of the dispatcher and listener threads to the test shutdown process so that they don't accrue
* Have a single thread pool for the test language "server" for the lifetime of the test suite

ahmedneilhussain:
@mickaelistria @sebthom for you to ponder...

sebthom:
I gave it a try. Before I could see the launcher threads piling up by repeatedly opening/closing the last file of a project. This is not happening anymore. This listener threads are terminated after a while when the last file of a project was closed. So this looks good.

Some other observations:
- The only threads staying alive are the dispatcher threads which are created one per Language Server per project. So if you work on a large workspace and maybe even have projects which use multiple Language Servers you can still end up with some dozen if not hundreds of lingering dispatcher threads. for example for a wildwebdeveloper project I see three dedicated dispatcher threads running per project (jsts-dispatcher, angular-dispatcher, eslint-dispatcher). these threads keep running even if I close or delete a project from a workspace.

- The thread names are only unique per project. E.g. if I open a typescript file in three different projects, I see the following threads:
image
Maybe there is a better way to enumerate the dispatcher threads, e.g. using a static incrementing AtomicInteger, so that we endup with e.g.
```
org.eclipse.wildwebdeveloper.jsts-dispatcher-1
org.eclipse.wildwebdeveloper.jsts-listener-1-1
org.eclipse.wildwebdeveloper.jsts-listener-1-2
org.eclipse.wildwebdeveloper.jsts-dispatcher-2
org.eclipse.wildwebdeveloper.jsts-listener-2-1
org.eclipse.wildwebdeveloper.jsts-listener-2-2
```

rubenporras:
> * Maybe there is a better way to enumerate the dispatcher threads, e.g. using a static incrementing AtomicInteger, so that we endup with e.g.

Yes, that would be better than having non-unique names, but what about including IProject#getName if there is a project as part of the dispatcher thread name? Would that not be even more meaningful?

sebthom:
> > * Maybe there is a better way to enumerate the dispatcher threads, e.g. using a static incrementing AtomicInteger, so that we endup with e.g.
> 
> Yes, that would be better than having non-unique names, but what about including IProject#getName if there is a project as part of the dispatcher thread name? Would that not be even more meaningful?

yes, that would be even more meaningful.

rubenporras:
> > > * Maybe there is a better way to enumerate the dispatcher threads, e.g. using a static incrementing AtomicInteger, so that we endup with e.g.
> > 
> > 
> > Yes, that would be better than having non-unique names, but what about including IProject#getName if there is a project as part of the dispatcher thread name? Would that not be even more meaningful?
> 
> yes, that would be even more meaningful.

@ahmedneilhussain , would you mind doing that change in your PR? 

Thanks

ahmedneilhussain:
> > > > * Maybe there is a better way to enumerate the dispatcher threads, e.g. using a static incrementing AtomicInteger, so that we endup with e.g.
> > > 
> > > 
> > > Yes, that would be better than having non-unique names, but what about including IProject#getName if there is a project as part of the dispatcher thread name? Would that not be even more meaningful?
> > 
> > 
> > yes, that would be even more meaningful.
> 
> @ahmedneilhussain , would you mind doing that change in your PR?
> 
> Thanks

I can take a look later today, probably.

ahmedneilhussain:
> * The only threads staying alive are the dispatcher threads which are created one per Language Server per project. So if you work on a large workspace and maybe even have projects which use multiple Language Servers you can still end up with some dozen if not hundreds of lingering dispatcher threads. for example for a wildwebdeveloper project I see three dedicated dispatcher threads running per project (jsts-dispatcher, angular-dispatcher, eslint-dispatcher). these threads keep running even if I close or delete a project from a workspace.

I had forgotten that some LS can only handle one project, so multiple projects for a language => multiple LS and multiple dispatchers. That is an argument in favour of shutting down the dispatchers when the LS closes, as per your PR @sebthom . Although if you have a lot of projects each with its own LS then maybe the large number of OS *processes* is going to be more of an issue than the large number of threads...!

ahmedneilhussain:
Hi @sebthom @rubenporras are you OK with the above implementation of your suggestion?

sebthom:
Ok, here follows a rather lengthy response, despite it is just about cosmetics :smiley:

a) I think we can ditch the index from the dispatcher thread name since it is a SingleThreadExecutor. So instead of:
```
Thread [LS-org.eclipse.wildwebdeveloper.jsts@myproject-dispatcher-0] (Running)	
Thread [LS-org.eclipse.wildwebdeveloper.eslint@myproject-dispatcher-0] (Running)	
```
it would just look like this:
```
Thread [LS-org.eclipse.wildwebdeveloper.jsts@myproject-dispatcher] (Running)	
Thread [LS-org.eclipse.wildwebdeveloper.eslint@myproject-dispatcher] (Running)	
```

b) Since there are often long project names with dots and dashes in it I would suggest to use a different separator for the dispatcher/listener suffix or even reordering the thread name components:
Instead of:
```
Thread [LS-org.eclipse.wildwebdeveloper.jsts@com.acme.groupId.my-great-project-dispatcher] (Running)
Thread [LS-org.eclipse.wildwebdeveloper.jsts@com.acme.groupId.my-great-project-listener-0] (Running)
Thread [LS-org.eclipse.wildwebdeveloper.eslint@com.acme.groupId.my-great-project-dispatcher] (Running)
Thread [LS-org.eclipse.wildwebdeveloper.eslint@com.acme.groupId.my-great-project-listener-0] (Running)
```
Replace the last dash separator with e.g. an # or another @
```
Thread [LS-org.eclipse.wildwebdeveloper.jsts@com.acme.groupId.my-great-project#dispatcher] (Running)
Thread [LS-org.eclipse.wildwebdeveloper.jsts@com.acme.groupId.my-great-project#listener-0] (Running)
Thread [LS-org.eclipse.wildwebdeveloper.eslint@com.acme.groupId.my-great-project#dispatcher] (Running)
Thread [LS-org.eclipse.wildwebdeveloper.eslint@com.acme.groupId.my-great-project#listener-0] (Running)
```
Or place the project name at the end
```
Thread [LS-org.eclipse.wildwebdeveloper.jsts#dispatcher@com.acme.groupId.my-great-project] (Running)
Thread [LS-org.eclipse.wildwebdeveloper.jsts#listener-0@com.acme.groupId.my-great-project] (Running)
Thread [LS-org.eclipse.wildwebdeveloper.eslint#dispatcher@com.acme.groupId.my-great-project] (Running)
Thread [LS-org.eclipse.wildwebdeveloper.eslint#listener-0@com.acme.groupId.my-great-project] (Running)
```
Or at the front
```
Thread [LS-com.acme.groupId.my-great-project#org.eclipse.wildwebdeveloper.jsts-dispatcher] (Running)
Thread [LS-com.acme.groupId.my-great-project#org.eclipse.wildwebdeveloper.jsts-listener-0] (Running)
Thread [LS-com.acme.groupId.my-great-project#org.eclipse.wildwebdeveloper.eslint-dispatcher] (Running)
Thread [LS-com.acme.groupId.my-great-project#org.eclipse.wildwebdeveloper.eslint-listener-0] (Running)
```

sebthom:
> I had forgotten that some LS can only handle one project, so multiple projects for a language => multiple LS and multiple dispatchers.

So far I only have seen that LSP4E launches language servers per project. How does LSP4E decide to launch per workspace or project?

ahmedneilhussain:
> Ok, here follows a rather lengthy response, despite it is just about cosmetics 😃
> 
> a) I think we can ditch the index from the dispatcher thread name since it is a SingleThreadExecutor. So instead of:

I've made this change and also adopted hash as a field separator. I wasn't sure if `String.format(fs, int)` would throw if `fs` lacked a `%d` or something to consume the argument, but I checked and unconsumed arguments are fine (missing arguments are not!)

Does that look better @sebthom ?

ahmedneilhussain:
> > I had forgotten that some LS can only handle one project, so multiple projects for a language => multiple LS and multiple dispatchers.
> 
> So far I only have seen that LSP4E launches language servers per project. How does LSP4E decide to launch per workspace or project?

Ha! I looked at things the other way round - I am only familiar with my company's plugin, which is multi-project, so I assumed that ours was the majority case...

IIRC it's part of the LSP server capabilities - it declares on startup whether it can handle multiple projects or not, and if not is up to the client to start more servers.

In LSP4e, `LanguageServerWrapper` exposes methods `canOperate(IDocument)` and `canOperate(IProject)`. In the various methods to get the appropriate wrapper for an operation, there's logic to loop through the existing wrappers to see whether any is appropriate for a particular operation. If a wrapper is able to handle the content type, but is already open for a different project and the LS can't handle multiple projects, then a new wrapper/server must be started; if it's a multi-project LS or the wrapper is idle and not attached to a project then it can be used without starting a new one.

Something like that!

See
https://github.com/eclipse/lsp4e/blob/9ffc1703b5b44caebbb3e76126d57047252079c5/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServiceAccessor.java#L445
https://github.com/eclipse/lsp4e/blob/9ffc1703b5b44caebbb3e76126d57047252079c5/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServiceAccessor.java#L346

for example.

sebthom:
> I've made this change and also adopted hash as a field separator. I wasn't sure if `String.format(fs, int)` would throw if `fs` lacked a `%d` or something to consume the argument, but I checked and unconsumed arguments are fine (missing arguments are not!)
> 
> Does that look better @sebthom ?

Looks good!

sebthom:
> > Ok, here follows a rather lengthy response, despite it is just about cosmetics 😃
> > a) I think we can ditch the index from the dispatcher thread name since it is a SingleThreadExecutor. So instead of:
> 
> I've made this change and also adopted hash as a field separator. I wasn't sure if `String.format(fs, int)` would throw if `fs` lacked a `%d` or something to consume the argument, but I checked and unconsumed arguments are fine (missing arguments are not!)
> 
> Does that look better @sebthom ?

@ahmedneilhussain  I just learned about the singleton flag in language server registration after checking the "canOperate" methods as you suggested. I think we should also check I language server is a singleton and then not add the project name to the threads. I.e. evaluate `serverDefinition.isSingleton`

ahmedneilhussain:
> @ahmedneilhussain I just learned about the singleton flag in language server registration after checking the "canOperate" methods as you suggested. I think we should also check I language server is a singleton and then not add the project name to the threads. I.e. evaluate `serverDefinition.isSingleton`

I was unaware of the singleton flag!! I thought that multi-project support was 'live' information that you could only get from the ServerCapabilities returned on startup by a running LS. Does the singleton flag express this statically?

@mickaelistria can you clarify?

sebthom:
Found this in languageServer.exsd:
"Whether a single instance of this language server should be instantiated and reused for all supported files. Sort of workaround for https://github.com/Microsoft/language-server-protocol/issues/674 https://github.com/Microsoft/language-server-protocol/issues/674"

ahmedneilhussain:
Ah, thanks! So much for my cursory attempt to RTFM...

ahmedneilhussain:
Done @sebthom

sebthom:
LGTM

rubenporras (approved):
looks good. Thanks

PR #379, opened 2023-01-19
Extract API to get list of started language server wrappers

rubenporras:
Extract API to get list of started language server wrappers and reuse it inside LanguageServiceAccessor

### DIFF
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServiceAccessor.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServiceAccessor.java
@@ -65,7 +65,7 @@ public class LanguageServiceAccessor {
 		// this class shouldn't be instantiated
 	}
 
-	private static final Set<LanguageServerWrapper> startedServers = new CopyOnWriteArraySet<>();
+	private static final Set<@NonNull LanguageServerWrapper> startedServers = new CopyOnWriteArraySet<>();
 	private static final Map<StreamConnectionProvider, LanguageServerDefinition> providersToLSDefinitions = new HashMap<>();
 
 	/**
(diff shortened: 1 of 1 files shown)


rubenporras:
@ahmedneilhussain: this would be the PR with the extracted method, but I obviously made a mistake, so I need to take another look at it

rubenporras:
@sebthom , this is an API that we would use for https://github.com/eclipse/lsp4e/pull/347 or for https://github.com/eclipse/lsp4e/pull/344. I extracted it here so that we can review the change independently. The good thing is that having the API factored out already makes the existing code smaller :)

sebthom:
Looks good to me. 

I generally somehow find the LanguageServerAccessor class a bit messy. I am wondering if it really is necessary to have so many methods to retrieve LanguageServers and LanguageServerWrappers and if we not e.g. could get rid of the methods returning LanguageServers and consistently work with the wrappers.

rubenporras:
@sebthom , yes, I agree that we should do away with the LanguageServers and always use the wrapper. But for doing that, the wrapper needs to wrap around the functions that the server gives access to. That is why in https://github.com/eclipse/lsp4e/pull/347 the class `LanguageServerWrapper` implements `LanguageServer`, because it reduces a lot the APIs needed, one does not need any longer to deal with wrappers and with servers, wrappers are enough.

I can create standalone PR if you like to simplify further the APIs if we would use such a pattern

ahmedneilhussain:
> Looks good to me.
> 
> I generally somehow find the LanguageServerAccessor class a bit messy. I am wondering if it really is necessary to have so many methods to retrieve LanguageServers and LanguageServerWrappers and if we not e.g. could get rid of the methods returning LanguageServers and consistently work with the wrappers.

Have you looked in detail at #344 (my proposal) or #347 (Ruben's tweaks of it) @sebthom ? The changes were made to improve concurrency guarantees, but in the course of doing so I tried to offer a higher level API. In a draft version of the changes in #333 I ported some code across to use the API (see the later commits in that PR). I have not ported those across yet as we have tried to keep a more focused PR and will submit the examples (and port more client code) when the API change is merged. The unit tests also give a flavour of what consuming code should look like.

I think one reason for the messiness is that `LanguageServiceAccessor` is a little bit too fine-grained, and among the many methods it's not always obvious what the guarantees are and when they should be used.

I hope my API makes it much easier to implement the common use-cases. `LanguageServiceAccessor` is still used in my proposal, since there is a lot of battle-hardened logic that would be hard to rewrite from scratch, but the intention is that it becomes low-level code that most consumer should not invoke directly. Over time it can be tidied up and dead code removed.

Issue #377, opened 2023-01-18
memory leak in ConnectDocumentToLanguageServerSetupParticipant


martinlippert:
While analyzing a heap dump of an IDE instance running for some time, I came across the `ConnectDocumentToLanguageServerSetupParticipant` and it looks like the Job objects are not removed from `SUBMITTED_JOBS` all the time.

It looks like the `run` method should remove the job objects from the map in all the cases, not only after getting the language servers.

ahmedneilhussain:
You're right, that's my bad. I'll submit a patch.

martinlippert:
Fixed via https://github.com/eclipse/lsp4e/pull/380

PR #380, opened 2023-01-19
Test teardown - make sure global object is cleaned out on all paths

ahmedneilhussain:
The SUBMITTED_JOBS was introduced to help test teardown so that we can cancel any of these jobs if the test finishes before they have a chance to run. Wasn't clearing out the map on all paths.

Addresses https://github.com/eclipse/lsp4e/issues/377

### DIFF
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/ConnectDocumentToLanguageServerSetupParticipant.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/ConnectDocumentToLanguageServerSetupParticipant.java
@@ -70,18 +70,25 @@ public class ConnectDocumentToLanguageServerSetupParticipant implements IDocumen
 			protected IStatus run(IProgressMonitor monitor) {
 				ITextFileBuffer buffer = ITextFileBufferManager.DEFAULT.getTextFileBuffer(document);
 				if (buffer == null || buffer.getLocation() == null) { // document no more relevant
+					SUBMITTED_JOBS.remove(this);
 					return Status.OK_STATUS;
 				}
 				if (monitor.isCanceled()) {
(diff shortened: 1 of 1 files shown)


ahmedneilhussain:
@martinlippert thanks for the bug report. My bad. Hopefully this will address the issue.

Could use weak references or a weak hash map [wrapped as a set] but probably overkill. Let me know what you think.

martinlippert (approved):
Looks good to me. Thanks for taking a look!!!

org.eclipse.lsp4e/src/org/eclipse/lsp4e/ConnectDocumentToLanguageServerSetupParticipant.java
```diff
@@ -83,6 +85,11 @@ protected IStatus run(IProgressMonitor monitor) {
 					});
```
sebthom, line 85: If `LanguageServiceAccessor.getLanguageServers()` fails and thenRun() is not executed, would the job still be removed from SUBMITTED_JOBS?

ahmedneilhussain: Fair point - I've correct this to do it in the main body of the job run method in all cases - once we get to that point it's too late to cancel so there's no point in keeping that entry for the test teardown.

I don't think it's easy to make the tests bulletproof - my original change was just an attempt to mitigate a major source of noise/flakiness. But test hooks shouldn't cause resource leaks in production...

sebthom (approved):

martinlippert:
Thanks @ahmedneilhussain for the fix and thanks @sebthom for the additional review, much appreciated.

PR #382, opened 2023-01-19
Rename new api for core document changes

rubenporras:
Tidy API the new API for executing requests and sending notifications by
removing the "onLatestVersion". It is a bit verbose to add this to the
the method names, since we will change all calls to use this two
methods, it is enough to have this feature of the methods as part of the
javadoc.

Also remove some comment which has been left over and removed some
unnecessary 'this'.


rubenporras:
@mickaelistria , this renaming is subjective, but @ahmedneilhussain and me both renamed the methods in our versions of the API to be something shorter. I think that since we will use this methods quite a bit, it is good if they have short names, still being meaningful, of course.
If you agree, I would merge this PR.

rubenporras:
not needed any longer

PR #386, opened 2023-01-23
Reduce public API introduced by new API.

rubenporras:
Also rename the executors to remove abbreviations.


rubenporras:
Is this fine with you @ahmedneilhussain ?

PR #387, opened 2023-01-23
Reduce public API and rename methods to have a name closer to its implementation

rubenporras:


rubenporras:
Is this fine with you @ahmedneilhussain?

ahmedneilhussain:
> Is this fine with you @ahmedneilhussain?

I'm sure it's fine @rubenporras , but I'll just run through everything to make sure later today when I've finished cannibalising the bits of #333 as separate submissions.

PR #388, opened 2023-01-23
API: Enhancement to support async completion on the UI thread

ahmedneilhussain:
Should make for cleaner code where the follow up to a LS call needs to run on the UI thread anyway

### DIFF
--- a/org.eclipse.lsp4e.test/src/org/eclipse/lsp4e/test/LanguageServersTest.java
+++ b/org.eclipse.lsp4e.test/src/org/eclipse/lsp4e/test/LanguageServersTest.java
@@ -613,6 +613,65 @@ public class LanguageServersTest {
 	}
 	
 	@Test
+	public void testDeliverOnUI() throws Exception {
+		final AtomicInteger hoverCount = new AtomicInteger();
+		MockLanguageServer.INSTANCE.setTextDocumentService(new MockTextDocumentService(MockLanguageServer.INSTANCE::buildMaybeDelayedFuture) {
+			@Override
+			public synchronized void didChange(DidChangeTextDocumentParams params) {
+				super.didChange(params);
+			}
+			
+			@Override
+			public synchronized CompletableFuture<Hover> hover(HoverParams position) {
+				Hover hoverResponse = new Hover(Collections.singletonList(Either.forLeft("HoverContent" + hoverCount.incrementAndGet())), new Range(new Position(0,  0), new Position(0, 10)));
+				return CompletableFuture.completedFuture(hoverResponse);
+			}
+		});
+		
+		IFile testFile = TestUtils.createUniqueTestFileMultiLS(project, "Here is some content");
+		IEditorPart editor = TestUtils.openEditor(testFile);
+		ITextViewer viewer = LSPEclipseUtils.getTextViewer(editor);
+		final IDocument document = viewer.getDocument();
+		
+		final HoverParams params = new HoverParams();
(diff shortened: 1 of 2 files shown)


ahmedneilhussain:
@mickaelistria @rubenporras (and others) what do you think?

As you know, the new API already makes sure any publicly returned CompletableFuture completes on a background thread to make sure the users can't schedule arbitrary follow-up work and tie up the LS inbound listener thread - that's what all those `.thenApplyAsync(Function.identity())` calls are doing.

I realised it was a simple enhancement to optionally deliver the results on the UI thread rather than an arbitrary pool thread. In many cases this is how the results are going to get used anyway, so this gets rid of a layer of nested lambda functions.

mickaelistria:
I don't think it's necessary to have new API for that. Since Display implements Executor, one can now do `future.thenAcceptAsync(result -> widget.setText(result), widget.getDisplay())` to run code in UI Thread. One new construct is IMO not better than finding way to encourage consumers to use a more standard pattern.

The typical pattern would be
```java
LanguageServers.collectAll(ls -> ls.getCompletion()) // get LS info
  .thenApplyAsync(lsCompletions -> lsCompletions.stream().map(lsCompletion -> toEclipseCompletion(lsCompletion))) // map it to Eclipse domain in another thread
  .thenAcceptAsync(eclipseCompletions -> showInUI(eclipseCompletions), display);
```
So one should rarely go straight from LS to UI Thread because data transformation should not require/block one nor the other; thus I don't think the proposed API is so helpful.

ahmedneilhussain:
Ha! I hadn't realised `Display` implements `Executor` - you're quite right, that works out of the box, so this change is unnecessary.

Cheers.

mickaelistria:
> Ha! I hadn't realised Display implements Executor

It's pretty recent; an great idea and implementation from @laeubi

ahmedneilhussain:
> I don't think it's necessary to have new API for that. Since Display implements Executor, one can now do `future.thenAcceptAsync(result -> widget.setText(result), widget.getDisplay())` to run code in UI Thread. One new construct is IMO not better than finding way to encourage consumers to use a more standard pattern.
> 
> The typical pattern would be
> 
> ```java
> LanguageServers.collectAll(ls -> ls.getCompletion()) // get LS info
>   .thenApplyAsync(lsCompletions -> lsCompletions.stream().map(lsCompletion -> toEclipseCompletion(lsCompletion))) // map it to Eclipse domain in another thread
>   .thenAcceptAsync(eclipseCompletions -> showInUI(eclipseCompletions), display);
> ```
> 
> So one should rarely go straight from LS to UI Thread because data transformation should not require/block one nor the other; thus I don't think the proposed API is so helpful.

By the way, the first `.thenApplyAsync(lsCompletions -> etc. etc.` doesn't need the `async` - it's already in another thread by virtue of that `.thenApplyAsync(Function.identity())` in `LanguageServers`

mickaelistria:
> By the way, the first .thenApplyAsync(lsCompletions -> etc. etc. doesn't need the async - it's already in another thread by virtue of that .thenApplyAsync(Function.identity()) in LanguageServers

Right, and that's great because it's extremely easy to miss the Async suffix. Just by writing this example, I had already missed it ;)

laeubi:
> So one should rarely go straight from LS to UI Thread because data transformation should not require/block one nor the other; thus I don't think the proposed API is so helpful.

Just in case you want visual feedback there is now also `BusyIndicator#compute` and `BusyIndicator#execute`

PR #389, opened 2023-01-23
New api - port hover

ahmedneilhussain:
Based on code originally submitted in #333  - this ports `LSPHover` to use the new API. Note that there is no timeout code in the revised logic, because there is no intermediate future exposed [the call to getLanguageServer is now rolled up into internal logic], but this should be fine as the consuming code in `getHoverRegion()` incorporates the timeout.

The API already excludes any nulls from the returned list, so that post-processing step is no longer needed here either.

Have tested it manually using the Corrosion plugin as well.

### DIFF
--- a/org.eclipse.lsp4e.test/src/org/eclipse/lsp4e/test/LanguageServersTest.java
+++ b/org.eclipse.lsp4e.test/src/org/eclipse/lsp4e/test/LanguageServersTest.java
@@ -148,7 +148,7 @@ public class LanguageServersTest {
 		List<String> hovers = result.join();
 		
 		assertTrue(hovers.contains("HoverContent1"));
-		assertFalse(hovers.contains("HoverContent2"));
+		assertFalse(hovers.contains(null));
 	}
 	
 	@Test
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/hover/LSPTextHover.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/hover/LSPTextHover.java
@@ -40,8 +40,9 @@ import org.eclipse.jface.text.ITextViewer;
 import org.eclipse.jface.text.Region;
 import org.eclipse.lsp4e.LSPEclipseUtils;
 import org.eclipse.lsp4e.LanguageServerPlugin;
-import org.eclipse.lsp4e.LanguageServiceAccessor;
+import org.eclipse.lsp4e.LanguageServers;
 import org.eclipse.lsp4j.Hover;
+import org.eclipse.lsp4j.HoverParams;
 import org.eclipse.lsp4j.MarkedString;
 import org.eclipse.lsp4j.MarkupContent;
 import org.eclipse.lsp4j.jsonrpc.messages.Either;
@@ -193,28 +194,15 @@ public class LSPTextHover implements ITextHover, ITextHoverExtension {
 	private void initiateHoverRequest(@NonNull ITextViewer viewer, int offset) {
 		final IDocument document = viewer.getDocument();
 		this.lastViewer = viewer;
-		this.request = LanguageServiceAccessor
-			.getLanguageServers(document, capabilities -> LSPEclipseUtils.hasCapability(capabilities.getHoverProvider()))
-				.thenApplyAsync(languageServers -> // Async is very important here, otherwise the LS Client thread is in
-													// deadlock and doesn't read bytes from LS
-				languageServers.stream()
-					.map(languageServer -> {
-						try {
-								return languageServer.getTextDocumentService()
-										.hover(LSPEclipseUtils.toHoverParams(offset, document))
-										.get(GET_TIMEOUT_MS, TimeUnit.MILLISECONDS);
-						} catch (ExecutionException | BadLocationException e) {
-							LanguageServerPlugin.logError(e);
-							return null;
-						} catch (InterruptedException e) {
-							LanguageServerPlugin.logError(e);
-							Thread.currentThread().interrupt();
-							return null;
-						} catch (TimeoutException e) {
-							LanguageServerPlugin.logWarning("Could not get hover due to timeout after " + GET_TIMEOUT_MS + " miliseconds", e); //$NON-NLS-1$ //$NON-NLS-2$
-							return null;
-						}
-					}).filter(Objects::nonNull).collect(Collectors.toList()));
+		try {
+			HoverParams params = LSPEclipseUtils.toHoverParams(offset, document);
+
+			this.request = LanguageServers.forDocument(document)
+				.withFilter(capabilities -> LSPEclipseUtils.hasCapability(capabilities.getHoverProvider()))
+				.collectAll((wapper, server) -> server.getTextDocumentService().hover(params));
+		} catch (BadLocationException e) {
+			LanguageServerPlugin.logError(e);
+		}
 	}
 
 	@Override
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/hover/LSPTextHover.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/hover/LSPTextHover.java
@@ -45,6 +45,7 @@ import org.eclipse.lsp4j.Hover;
 import org.eclipse.lsp4j.HoverParams;
 import org.eclipse.lsp4j.MarkedString;
 import org.eclipse.lsp4j.MarkupContent;
+import org.eclipse.lsp4j.ServerCapabilities;
 import org.eclipse.lsp4j.jsonrpc.messages.Either;
 import org.eclipse.mylyn.wikitext.markdown.MarkdownLanguage;
 import org.eclipse.mylyn.wikitext.parser.MarkupParser;
@@ -62,7 +63,7 @@ public class LSPTextHover implements ITextHover, ITextHoverExtension {
 
 	private IRegion lastRegion;
 	private ITextViewer lastViewer;
-	private CompletableFuture<List<Hover>> request;
+	private CompletableFuture<@NonNull List<@NonNull Hover>> request;
 
 	@Override
 	public String getHoverInfo(ITextViewer textViewer, IRegion hoverRegion) {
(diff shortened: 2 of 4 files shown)


mickaelistria:
What about the timeout?

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/hover/LSPTextHover.java
```diff
@@ -193,28 +194,15 @@ public IRegion getHoverRegion(ITextViewer textViewer, int offset) {
...
-							LanguageServerPlugin.logWarning("Could not get hover due to timeout after " + GET_TIMEOUT_MS + " miliseconds", e); //$NON-NLS-1$ //$NON-NLS-2$
-							return null;
-						}
-					}).filter(Objects::nonNull).collect(Collectors.toList()));
+		try {
+			HoverParams params = LSPEclipseUtils.toHoverParams(offset, document);
+
+			this.request = LanguageServers.forDocument(document)
+				.withFilter(capabilities -> LSPEclipseUtils.hasCapability(capabilities.getHoverProvider()))
+				.collectAll((wapper, server) -> server.getTextDocumentService().hover(params));
```
rubenporras, line 202:
```suggestion
				.collectAll(server -> server.getTextDocumentService().hover(params));
```

rubenporras: Since we have overloaded methods we can use the version with only one argument :)

ahmedneilhussain:
> What about the timeout?

I mentioned this above - I think this is covered by the calling code that has a timeout for when it ultimately tries to dereference the hover.

https://github.com/eclipse/lsp4e/blob/5a75b544c083ab6b89ef11851dc07d2fcbe556ed/org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/hover/LSPTextHover.java#L177

Is there a need to have an intermediate stage timeout?

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/hover/LSPTextHover.java
ahmedneilhussain: D'Oh. Hoist by my own petard ;-)

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/hover/LSPTextHover.java
```diff
@@ -193,28 +194,15 @@ public IRegion getHoverRegion(ITextViewer textViewer, int offset) {
...
-						} catch (TimeoutException e) {
-							LanguageServerPlugin.logWarning("Could not get hover due to timeout after " + GET_TIMEOUT_MS + " miliseconds", e); //$NON-NLS-1$ //$NON-NLS-2$
-							return null;
-						}
-					}).filter(Objects::nonNull).collect(Collectors.toList()));
+		try {
+			HoverParams params = LSPEclipseUtils.toHoverParams(offset, document);
+
+			this.request = LanguageServers.forDocument(document)
+				.withFilter(capabilities -> LSPEclipseUtils.hasCapability(capabilities.getHoverProvider()))
```
rubenporras, line 201:
```suggestion
				.withCapabilities(ServerCapabilities::getHoverProvider)
```

rubenporras: Now that I look only at this small consumer, It looks like it makes to have a method `withCapabilities` that would allow to write code as simple as in my suggested replacement. Would that be possible?

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/hover/LSPTextHover.java
```diff
@@ -193,28 +194,15 @@ public IRegion getHoverRegion(ITextViewer textViewer, int offset) {
...
+				.collectAll(server -> server.getTextDocumentService().hover(params));
```
angelozerr, line 202: Please keep in mind that it should really nice to support cancel checker. For instance with LemMinx Maven, hover can take some time because compute is complex. When user stop the hover, it should cancel all ComplecteableFuture `server.getTextDocumentService().hover(...)`.

This cancel should be done before `this.request = LanguageServers.forDocument(document)`

I spend some times to try to support this cancel by trying to use CompletableFuture API (without success). The only solution that I found is to collect in a list all call of server.getTextDocumentService().hover(..) and loop for the list to cancel it.

See my draft with documentLink but it is the same thing for hover https://github.com/eclipse/lsp4e/pull/238

If you find a better idea with my LSPRequest wrapper, you are welcome!

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/hover/LSPTextHover.java
rubenporras: Yes, something like 

```
	public E withCapability(final @NonNull Function<ServerCapabilities, Either<Boolean, ? extends Object>> serverCapabilities) {
		this.filter = f -> LSPEclipseUtils.hasCapability(serverCapabilities.apply(f));
		return (E)this;
	}
```

in `LanguageServers` should work.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/hover/LSPTextHover.java
```diff
@@ -193,28 +194,15 @@ public IRegion getHoverRegion(ITextViewer textViewer, int offset) {
...
-						} catch (InterruptedException e) {
-							LanguageServerPlugin.logError(e);
-							Thread.currentThread().interrupt();
-							return null;
-						} catch (TimeoutException e) {
-							LanguageServerPlugin.logWarning("Could not get hover due to timeout after " + GET_TIMEOUT_MS + " miliseconds", e); //$NON-NLS-1$ //$NON-NLS-2$
-							return null;
-						}
-					}).filter(Objects::nonNull).collect(Collectors.toList()));
+		try {
```
rubenporras, line 197: To avoid JDT null warnings we need to add
```
if (document == null) {
	return;
}
```
at this point

rubenporras: I cannot comment on that line, but to avoid JDT warnings, we also need to annotate the declaration of request to be
`	private CompletableFuture<@NonNull List<@NonNull Hover>> request;`
could you do that?

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/hover/LSPTextHover.java
ahmedneilhussain: No objection from me. Anything that pushes the boilerplate into common code and makes the calling code more focused on the problem domain is nice. Shall I go ahead and commit your suggestion?

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/hover/LSPTextHover.java
rubenporras: Yes, I agree, supporting a cancel checker is a must, at least for a mature product. Since we are introducing now a new API, I think it makes sense to see it can accommodate the case.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/hover/LSPTextHover.java
rubenporras: Sure, I am happy if you just add it as part of this small PR, if it gets bigger, we could merge it standalone

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/hover/LSPTextHover.java
rubenporras: Hi @ahmedneilhussain , I think this PR would not prevent handling cancelation as done in #238. One can still wrap the request into something that can be cancelled once a newer request is initiated, right? If so we can merge this one without preventing cancellation handling, and then I would merge this PR independent of the cancellation support which does not exist yet.

In can also refactor it later to add some `withCapability` method to `LanguageServers`

org.eclipse.lsp4e.test/src/org/eclipse/lsp4e/test/LanguageServersTest.java
```diff
@@ -148,7 +148,7 @@ public synchronized CompletableFuture<Hover> hover(HoverParams position) {
 		List<String> hovers = result.join();
 		
 		assertTrue(hovers.contains("HoverContent1"));
-		assertFalse(hovers.contains("HoverContent2"));
+		assertFalse(hovers.contains(null));
```
rubenporras, line 151: why do we need to change "HoverContents2" with `null`? I do not understand this part.

rubenporras:
I am done with the review, I would approve it if I would understand why we need to change the test, but I do not. Besides, something with the PR is off, because the Legal Agreement check has failed. Could you fix that and other comments (if appropriate)?

org.eclipse.lsp4e.test/src/org/eclipse/lsp4e/test/LanguageServersTest.java
ahmedneilhussain: I put it as part of the first commit (which got rid of the superfluous call to shutdown the dispatch threads) as it's maintenance not really part of this main change. I realised the test is wrong, or rather incomplete - `testCollectAllExcludesNulls()` should be making sure that any null response from a language server gets filtered out of the returned list. The test (owing to some cut-and-paste errors and a brain fade on my part) wasn't ensuring that properly. If you look at the mock it can never generate 'HoverContent2' it is supposed to be generating null on the second call, and the test should explicitly be testing for the null getting filtered out.

I just realised this in the course of the main change - I had made some errors in earlier drafts of this work ensuring that the null semantics were preserved when code was rewritten to the new API, so I checked this carefully and spotted the flawed test.

Can split this fix out if you think it is still unclear - I kept it as a separate commit to emphasise it was an unrelated change.

ahmedneilhussain:
Thanks @rubenporras it seems that my personal email (which is not associated with my account and license declaration) gets used when I commit your suggestion from within the PR. I will rebase to correct the error and also implement your other suggestion.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/hover/LSPTextHover.java
ahmedneilhussain: Sure

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/hover/LSPTextHover.java
ahmedneilhussain: Implemented

rubenporras (approved):

PR #396, opened 2023-01-24
Remove internal method in LanguageServerWrapper for LanguageServers

rubenporras:


rubenporras:
@ahmedneilhussain , okay with you?

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServers.java
```diff
@@ -106,7 +107,7 @@
 	public <T> List<@NonNull CompletableFuture<@Nullable T>> computeAll(BiFunction<? super LanguageServerWrapper, LanguageServer, ? extends CompletionStage<T>> fn) {
 		return getServers().stream()
 				.map(cf -> cf
-						.thenCompose(w -> w == null ? CompletableFuture.completedFuture(null) : w.executeImpl(ls -> fn.apply(w, ls)).thenApplyAsync(Function.identity())))
```
ahmedneilhussain, line 109: I really think `executeImpl` is needed as an internal hook here, otherwise you get two layers of `.thenApplyAsync(Function.identity())`.

It's probably mostly benign to do so. but it will decrease throughput - that maybe won't affect LS things much, but don't forget that the ForkJoinPool is a shared resource, so we should avoid flooding it with very small tasks that just pass the work onto another async task.

For the sake of saving a 1-line method, I'd really rather keep it as it is.

ahmedneilhussain:
I think having an executor for a specific wrapper is fine though - I had thoughts along similar lines. An executor that is specific to a server definition might be useful at some point in the future, too.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServers.java
rubenporras: The `execute` is the same as `executeImpl` before, so there is no added layer, right?

Otherwise, yes, in general I see your point about avoiding layes when possible to keep good performance.

The main idea is not to just save a one line method, but rather to make the API more consistent, so if we merge this there is a `sendNotification` and an `execute` method, and both behave in the same way, by just running the code, without composing. Without merging this PR, the equilavent to `sendNotification` is `executeImpl`, not `execute`. I find it hard to reason about such an API. It feels there is one method too many.

ahmedneilhussain: Ah, sorry, I misread the diff. You are right, but that has the reverse problem - someone using `execute()` will receive a completable future that completes on the LS listener thread. So the user can add a long running block `.andThen(sleepForAges())` that will potentially tie up the listener thread.

I'm trying to stop people doing that, hence all those `.andThenAsync(Function.identity())` which are added on the end of any public methods.

Unfortunately it means that can't have public methods that themselves are built out of public methods, or you get the problem of unnecessary churn in the ForkJoinPool. Hence the original need for `execute()` which is the same as `executeImpl()` but with the async-identity pattern stuck on the end.

ahmedneilhussain: Would `executeInternal()` or some other name help make things more consistent? It's not API - it's internal code *used* by the public API!

There's no need for two variants of `sendNotification()` because it's a void method that is 'done' as soon as the message is scheduled into the dispatcher queue. There's no `CompletableFuture` result where we have to worry about which thread the completion takes place on.

rubenporras:
also closing this one as there is no agreement that the change is beneficial

PR #398, opened 2023-01-24
prevent NPE when formatting multiple files

sebthom:
This PR prevents:

```java
Jan 24, 2023 8:17:54 PM org.eclipse.lsp4j.jsonrpc.RemoteEndpoint fallbackResponseError
SEVERE: Internal error: java.lang.NullPointerException: Cannot invoke "org.eclipse.lsp4j.ServerCapabilities.getDocumentFormattingProvider()" because "this.serverCapabilities" is null
java.util.concurrent.CompletionException: java.lang.NullPointerException: Cannot invoke "org.eclipse.lsp4j.ServerCapabilities.getDocumentFormattingProvider()" because "this.serverCapabilities" is null
	at java.base/java.util.concurrent.CompletableFuture.encodeThrowable(CompletableFuture.java:315)
	at java.base/java.util.concurrent.CompletableFuture.completeThrowable(CompletableFuture.java:320)
	at java.base/java.util.concurrent.CompletableFuture$AsyncRun.run(CompletableFuture.java:1807)
	at java.base/java.util.concurrent.CompletableFuture$AsyncRun.exec(CompletableFuture.java:1796)
	at java.base/java.util.concurrent.ForkJoinTask.doExec(ForkJoinTask.java:373)
	at java.base/java.util.concurrent.ForkJoinPool$WorkQueue.topLevelExec(ForkJoinPool.java:1182)
	at java.base/java.util.concurrent.ForkJoinPool.scan(ForkJoinPool.java:1655)
	at java.base/java.util.concurrent.ForkJoinPool.runWorker(ForkJoinPool.java:1622)
	at java.base/java.util.concurrent.ForkJoinWorkerThread.run(ForkJoinWorkerThread.java:165)
Caused by: java.lang.NullPointerException: Cannot invoke "org.eclipse.lsp4j.ServerCapabilities.getDocumentFormattingProvider()" because "this.serverCapabilities" is null
	at org.eclipse.lsp4e.LanguageServerWrapper.lambda$18(LanguageServerWrapper.java:932)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1511)
	at org.eclipse.lsp4e.LanguageServerWrapper.registerCapability(LanguageServerWrapper.java:906)
	at org.eclipse.lsp4e.LanguageClientImpl.lambda$2(LanguageClientImpl.java:136)
	... 6 more
```

@ahmedneilhussain I am using the new LanguageServers class, can you please have a look if I use it as intended?

any feedback is welcome!

### DIFF
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/format/LSPFormatFilesHandler.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/format/LSPFormatFilesHandler.java
@@ -16,6 +16,8 @@ import java.util.Collections;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.Future;
 
 import org.eclipse.core.commands.AbstractHandler;
 import org.eclipse.core.commands.ExecutionEvent;
@@ -34,11 +36,11 @@ import org.eclipse.jdt.annotation.NonNull;
 import org.eclipse.jface.text.IDocument;
 import org.eclipse.jface.text.TextSelection;
 import org.eclipse.lsp4e.LanguageServerPlugin;
+import org.eclipse.lsp4e.LanguageServers;
 import org.eclipse.lsp4e.LanguageServersRegistry;
 import org.eclipse.lsp4e.operations.format.LSPFormatter.VersionedFormatRequest;
 import org.eclipse.lsp4e.ui.Messages;
 import org.eclipse.lsp4e.ui.UI;
-import org.eclipse.lsp4j.TextEdit;
 import org.eclipse.osgi.util.NLS;
 import org.eclipse.ui.part.FileEditorInput;
 import org.eclipse.ui.texteditor.DocumentProviderRegistry;
@@ -58,7 +60,14 @@ public class LSPFormatFilesHandler extends AbstractHandler {
 					if (subMonitor.isCanceled()) {
 						return;
 					}
-					formatFile(file, monitor);
+					try {
+						formatFile(file, monitor).get();
+					} catch (InterruptedException ex) {
+						LanguageServerPlugin.logError(ex);
+						Thread.currentThread().interrupt();
+					} catch (java.util.concurrent.ExecutionException ex) {
+						LanguageServerPlugin.logError(ex);
+					}
 					subMonitor.worked(1);
 				}
 				subMonitor.done();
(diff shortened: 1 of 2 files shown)


ahmedneilhussain:
HI @sebthom have a look at #401 - I've tried to port the file across but I was unsure exactly the source of the NPE that you were addressing as it was mixed up with your other changes. Can we discuss?

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/format/LSPFormatFilesHandler.java
```diff
@@ -69,37 +78,50 @@ public Object execute(final ExecutionEvent event) throws ExecutionException {
...
-			docProvider.disconnect(file);
+			return CompletableFuture.completedFuture(false);
 		}
+		final IDocument doc = docProvider.getDocument(file);
+		if (doc == null)
+			return CompletableFuture.completedFuture(false);
+
+		return LanguageServers.forDocument(doc) //
+				.withFilter(LSPFormatter::supportsFormatting) //
+				.computeFirst((lsw, ls) -> {
```
ahmedneilhussain, line 98: The new API is intended to provide the LS to you on a dedicated dispatch thread so it is safe to make a direct call to the LS. You shouldn't do arbitrary computations inside one of these lambda blocks. It will tie up the dispatch thread so that nothing else can be sent to the LS...

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/format/LSPFormatFilesHandler.java
```diff
@@ -69,37 +78,50 @@ public Object execute(final ExecutionEvent event) throws ExecutionException {
...
+					monitor.setTaskName(NLS.bind(Messages.LSPFormatFilesHandler_FormattingFile, file.getFullPath()));
+					final VersionedFormatRequest formatRequest = formatter.versionedRequestFormatting(doc,
```
ahmedneilhussain, line 100: You are supposed to be doing something like `ls.getTextDocumentService().format(...)` in one of these blocks, having been handed the language server(s) appropriate for your requested capabilities. You can then use the CompletableFuture returned by `computeFirst()` to schedule follow-up processing.

What you are doing is calling the legacy format method which in turn is part of the old API - it independently goes and gets the Language Server(s), and submits a request on a random thread. I'm quite surprised this works at all..!

ahmedneilhussain:
Hi @sebthom I have picked through this class and I think the problem is that batch processing such as you are trying to do doesn't play well at all with the automated shutdown-on-disconnect logic implemented in #103 

It will depend on which LS plugin you are using - I am trying with Corrosion which has the default timeout of zero - ie as soon as you close the last document associated with it, `stop()` is called on the server. See https://github.com/eclipse/lsp4e/blob/eef80dfc4e6e0d3f22567d73d285221a1fe01af9/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServerWrapper.java#L712

If you run a batch operation like this with no files open in the IDE, and on a LS with a short or zero timeout, then the LS will keep getting shut down and started up again. Depending on timings, the LS may be being shut down while a response is still pending. I think the start/stop code in LanguageServerWrapper probably has a couple of subtle outstanding race conditions - I wouldn't be surprised if there weren't some possible interleavings of calls to start and stop that caused problems.

However, even if this were correct it's obviously very inefficient to start the LS and stop for every document in a batch request. I think I can see a way of setting a batch mode that would keep the LS alive for the duration of a batch request. How does that sound?
@rubenporras do you have a view?

ahmedneilhussain:
I think this is a problem on master and in my branch and I don't think this PR will fix it - but I think in all three code streams the symptoms will be subtly different.

rubenporras:
> @rubenporras do you have a view?

To me disconnection immediately does not make any sense. That is why introduced the extension point to be able to define a higher timeout. Regardless of that extension point, probably we should increase the default to be something else than zero in a different PR.

ahmedneilhussain:
Hi @sebthom do you want to see if `master` has resolved this issue now Ruben's longer (non-zero!) default timeout fix is in place?

ahmedneilhussain:
I added a timeout for the format operation itself: as @rubenporras pointed out we should avoid calling `.get()` with no timeout on those futures as they could potentially block forever. However the 5s I randomly chose is maybe not enough.

See https://github.com/eclipse/lsp4e/issues/414 where I mention this.

sebthom:
@ahmedneilhussain I gave it a try, the recent changes are looking good!

PR #399, opened 2023-01-25
Use getLineOffset instead of getLineInformation().getOffset()

rubenporras:
Use getLineOffset instead of getLineInformation().getOffset() as it returns the same offset without the creation of an intermediate IRegion object.

Since DiagnosticsTest.testDiagnosticsRangeAfterDocument test that diagnostics outside of the document range are published with range start/end equals to the document length, and getLineOffset has an extra consistency check over getLineInformation().getOffset() which throws BadLocationExection, we need to handle this case when computing marker attributes.


mickaelistria:
Thanks!

PR #401, opened 2023-01-25
New api: Port formatting

ahmedneilhussain:
This ports the formatting code over to the new API.

Added in direct API support for optimistic locking [which had already been added for formatting by @rubenporras in earlier changes] as it's a generalised issue.

As was pointed out in other discussions, detecting changes that may invalidate server responses (because the server was using stale information to calculate that response) is something that may be of general use, not just for edits. For edits it is obviously critically important because otherwise document corruption may ensue. However even 'benign' requests that don't change any state may end up showing misleading data to the user.

The vagaries of type erasure meant that I've added special support for versioned edits specifically, and generic support for versioned data in general, although the latter doesn't do much other than bundle content with a version.

### DIFF
--- a/org.eclipse.lsp4e.test/src/org/eclipse/lsp4e/test/LanguageServersTest.java
+++ b/org.eclipse.lsp4e.test/src/org/eclipse/lsp4e/test/LanguageServersTest.java
@@ -14,15 +14,20 @@ package org.eclipse.lsp4e.test;
 
 import static org.eclipse.lsp4e.LanguageServiceAccessor.getActiveLanguageServers;
 import static org.eclipse.lsp4e.test.TestUtils.createUniqueTestFile;
+import static org.eclipse.lsp4e.test.TestUtils.numberOfChangesIs;
 import static org.eclipse.lsp4e.test.TestUtils.openEditor;
+import static org.eclipse.lsp4e.test.TestUtils.waitForAndAssertCondition;
 import static org.eclipse.lsp4e.test.TestUtils.waitForCondition;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertNotEquals;
 import static org.junit.Assert.assertNull;
 import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
 
+import java.util.ArrayList;
 import java.util.Collections;
+import java.util.ConcurrentModificationException;
 import java.util.List;
 import java.util.Optional;
 import java.util.Vector;
@@ -37,27 +42,33 @@ import java.util.stream.Collectors;
 import org.eclipse.core.resources.IFile;
 import org.eclipse.core.resources.IProject;
 import org.eclipse.core.runtime.CoreException;
+import org.eclipse.jface.text.BadLocationException;
 import org.eclipse.jface.text.IDocument;
 import org.eclipse.jface.text.ITextViewer;
 import org.eclipse.lsp4e.LSPEclipseUtils;
 import org.eclipse.lsp4e.LanguageServerWrapper;
 import org.eclipse.lsp4e.LanguageServers;
 import org.eclipse.lsp4e.LanguageServers.LanguageServerDocumentExecutor;
+import org.eclipse.lsp4e.LanguageServers.VersionedEdits;
 import org.eclipse.lsp4e.tests.mock.MockLanguageServer;
 import org.eclipse.lsp4e.tests.mock.MockTextDocumentService;
 import org.eclipse.lsp4j.DidChangeTextDocumentParams;
+import org.eclipse.lsp4j.DocumentFormattingParams;
+import org.eclipse.lsp4j.FormattingOptions;
 import org.eclipse.lsp4j.Hover;
 import org.eclipse.lsp4j.HoverParams;
 import org.eclipse.lsp4j.Position;
 import org.eclipse.lsp4j.Range;
 import org.eclipse.lsp4j.ServerCapabilities;
 import org.eclipse.lsp4j.TextDocumentSyncKind;
+import org.eclipse.lsp4j.TextEdit;
 import org.eclipse.lsp4j.jsonrpc.messages.Either;
 import org.eclipse.lsp4j.services.LanguageServer;
 import org.eclipse.swt.custom.StyledText;
 import org.eclipse.swt.widgets.Display;
 import org.eclipse.ui.IEditorPart;
 import org.eclipse.ui.texteditor.AbstractTextEditor;
+import org.eclipse.ui.texteditor.ITextEditor;
 import org.junit.Assume;
 import org.junit.Before;
 import org.junit.Rule;
@@ -777,6 +788,81 @@ public class LanguageServersTest {
 		final LanguageServerDocumentExecutor executor = LanguageServers.forDocument(document);
 
 		assertEquals(document, executor.getDocument());
+	}
+	
+	@Test
+	public void testVersionSupportSuccess() throws Exception {
+		List<TextEdit> formattingTextEdits = new ArrayList<>();
+		formattingTextEdits.add(new TextEdit(new Range(new Position(0, 0), new Position(0, 1)), "MyF"));
+		formattingTextEdits.add(new TextEdit(new Range(new Position(0, 10), new Position(0, 11)), ""));
+		formattingTextEdits.add(new TextEdit(new Range(new Position(0, 21), new Position(0, 21)), " Second"));
+		MockLanguageServer.INSTANCE.setFormattingTextEdits(formattingTextEdits);
+
+		IFile file = TestUtils.createUniqueTestFile(project, "Formatting Other Text");
+		IEditorPart editor = TestUtils.openEditor(file);
+		ITextViewer viewer = LSPEclipseUtils.getTextViewer(editor);
+
+		final var doc = viewer.getDocument();
+		final var docId = LSPEclipseUtils.toTextDocumentIdentifier(doc);
+
+		final var params = new DocumentFormattingParams();
+		params.setTextDocument(docId);
+		params.setOptions(new FormattingOptions(4, true));
+		
+		var ex = LanguageServers.forDocument(doc).withCapability(ServerCapabilities::getDocumentFormattingProvider);
+		
+		var result = ex.computeFirst(ls -> ls.getTextDocumentService().formatting(params).thenApply(ex::toVersionedEdits));
+		
+		VersionedEdits edits = result.join().get();
+		editor.getSite().getShell().getDisplay().syncExec(() -> {
+			try {
+				edits.apply();
+			} catch (ConcurrentModificationException | BadLocationException e) {
+				fail(e.getMessage());
+			}
+		});
+
+		ITextEditor textEditor = (ITextEditor) editor;
+		textEditor.getDocumentProvider().getDocument(textEditor.getEditorInput());
+		assertEquals("MyFormattingOther Text Second", viewer.getDocument().get());
+		
+		final long currentTS = LSPEclipseUtils.getDocumentModificationStamp(doc);
+		var ex2 = LanguageServers.forDocument(doc);
+		var dummyVersioned = ex2.computeFirst(ls -> CompletableFuture.completedFuture("Hello").thenApply(ex2::toVersioned)).join().get();
+		assertEquals(currentTS, dummyVersioned.getVersion());
+		assertEquals("Hello", dummyVersioned.get());
+
+		TestUtils.closeEditor(editor, false);
+	}
+	
+	@Test(expected=ConcurrentModificationException.class)
+	public void testVersionedEditsFailsOnModification() throws Exception {
+		List<TextEdit> formattingTextEdits = new ArrayList<>();
+		formattingTextEdits.add(new TextEdit(new Range(new Position(0, 0), new Position(0, 1)), "MyF"));
+		formattingTextEdits.add(new TextEdit(new Range(new Position(0, 10), new Position(0, 11)), ""));
+		formattingTextEdits.add(new TextEdit(new Range(new Position(0, 21), new Position(0, 21)), " Second"));
+		MockLanguageServer.INSTANCE.setFormattingTextEdits(formattingTextEdits);
+
+		IFile file = TestUtils.createUniqueTestFile(project, "Formatting Other Text");
+		IEditorPart editor = TestUtils.openEditor(file);
+		ITextViewer viewer = LSPEclipseUtils.getTextViewer(editor);
+
+		final var doc = viewer.getDocument();
+		final var docId = LSPEclipseUtils.toTextDocumentIdentifier(doc);
+
+		final var params = new DocumentFormattingParams();
+		params.setTextDocument(docId);
+		params.setOptions(new FormattingOptions(4, true));
+		
+		var ex = LanguageServers.forDocument(doc).withCapability(ServerCapabilities::getDocumentFormattingProvider);
+		
+		var result = ex.computeFirst(ls -> ls.getTextDocumentService().formatting(params).thenApply(ex::toVersionedEdits));
+		
+		VersionedEdits edits = result.join().get();
+		viewer.getDocument().replace(0, 0, "Hello");
+		waitForAndAssertCondition(1_000,  numberOfChangesIs(1));
+
+		edits.apply();
 	}
 
 	private static class LSWPair {
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LSPEclipseUtils.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LSPEclipseUtils.java
@@ -71,6 +71,7 @@ import org.eclipse.jdt.annotation.Nullable;
 import org.eclipse.jface.dialogs.MessageDialog;
 import org.eclipse.jface.text.BadLocationException;
 import org.eclipse.jface.text.IDocument;
+import org.eclipse.jface.text.IDocumentExtension4;
 import org.eclipse.jface.text.ITextOperationTarget;
 import org.eclipse.jface.text.ITextViewer;
 import org.eclipse.jface.text.RewriteSessionEditProcessor;
@@ -1244,4 +1245,19 @@ public final class LSPEclipseUtils {
 		}
 		return eitherCapability.isRight() || eitherCapability.getLeft();
 	}
+
+	/**
+ 	 * Gets the modificatino stamp for the supplied document, or returns -1 if not available.
+ 	 *
+ 	 * In practice just a sanity-checked downcast of a legacy API: should expect the platfom to be instantiating
+ 	 * Documents that implement the later interfaces.
+ 	 *
+ 	 * Should be called on UI thread
+ 	 *
+ 	 * @param document Document to check
+ 	 * @return Opaque version stamp, or -1 if not available
+ 	 */
+ 	public static long getDocumentModificationStamp(@Nullable IDocument document) {
+ 		return document instanceof IDocumentExtension4 ? ((IDocumentExtension4) document).getModificationStamp() : -1;
+ 	}
 }
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServers.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServers.java
@@ -15,6 +15,7 @@ package org.eclipse.lsp4e;
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.ConcurrentModificationException;
 import java.util.List;
 import java.util.Objects;
 import java.util.Optional;
@@ -28,8 +29,10 @@ import java.util.stream.Stream;
 import org.eclipse.core.resources.IProject;
 import org.eclipse.jdt.annotation.NonNull;
 import org.eclipse.jdt.annotation.Nullable;
+import org.eclipse.jface.text.BadLocationException;
 import org.eclipse.jface.text.IDocument;
 import org.eclipse.lsp4j.ServerCapabilities;
+import org.eclipse.lsp4j.TextEdit;
 import org.eclipse.lsp4j.jsonrpc.messages.Either;
 import org.eclipse.lsp4j.services.LanguageServer;
 
@@ -67,6 +70,7 @@ public abstract class LanguageServers<E extends LanguageServers<E>> {
 	 */
 	@NonNull
 	public <T> CompletableFuture<@NonNull List<@NonNull T>> collectAll(BiFunction<? super LanguageServerWrapper, LanguageServer, ? extends CompletionStage<T>> fn) {
+		computeVersion();
 		final CompletableFuture<@NonNull List<T>> init = CompletableFuture.completedFuture(new ArrayList<T>());
 		return executeOnServers(fn).reduce(init, LanguageServers::add, LanguageServers::addAll)
 			// Ensure any subsequent computation added by caller does not block further incoming messages from language servers
@@ -104,6 +108,7 @@ public abstract class LanguageServers<E extends LanguageServers<E>> {
 	 */
 	@NonNull
 	public <T> List<@NonNull CompletableFuture<@Nullable T>> computeAll(BiFunction<? super LanguageServerWrapper, LanguageServer, ? extends CompletionStage<T>> fn) {
+		computeVersion();
 		return getServers().stream()
 				.map(cf -> cf
 						.thenCompose(w -> w == null ? CompletableFuture.completedFuture(null) : w.executeImpl(ls -> fn.apply(w, ls)).thenApplyAsync(Function.identity())))
@@ -137,6 +142,7 @@ public abstract class LanguageServers<E extends LanguageServers<E>> {
 	 * non-empty response, and with an empty <code>Optional</code> if none of the servers returned a non-empty result.
 	 */
 	public <T> CompletableFuture<Optional<T>> computeFirst(BiFunction<? super LanguageServerWrapper, LanguageServer, ? extends CompletionStage<T>> fn) {
+		computeVersion();
 		final CompletableFuture<Optional<T>> result = new CompletableFuture<>();
 
 		// Dispatch the request to the servers, appending a step to each such that
@@ -203,12 +209,22 @@ public abstract class LanguageServers<E extends LanguageServers<E>> {
 
 		private final @NonNull IDocument document;
 
+		private long startVersion;
+
 		LanguageServerDocumentExecutor(final @NonNull IDocument document) {
 			this.document = document;
 		}
 
 		public @NonNull IDocument getDocument() {
 			return this.document;
+		}
+
+		public VersionedEdits toVersionedEdits(List<? extends TextEdit> edits) {
+			return new VersionedEdits(startVersion, edits, document);
+		}
+
+		public <T> Versioned<T> toVersioned(T t) {
+			return new Versioned<>(startVersion, t);
 		}
 
 		/**
@@ -237,6 +253,11 @@ public abstract class LanguageServers<E extends LanguageServers<E>> {
 			return LanguageServiceAccessor.getLSWrappers(this.document).stream()
 				.map(this::connectIf)
 				.toList();
+		}
+
+		@Override
+		protected void computeVersion() {
+			this.startVersion = LSPEclipseUtils.getDocumentModificationStamp(document);
 		}
 
 	}
@@ -282,6 +303,62 @@ public abstract class LanguageServers<E extends LanguageServers<E>> {
 		}
 	}
 
+	/**
+	 * Specialised POJO for document edits specifically
+	 *
+	 */
+	public static class VersionedEdits extends Versioned<List<? extends TextEdit>> {
+
+		private final IDocument document;
+
+		public VersionedEdits(long version, List<? extends TextEdit> data, IDocument document) {
+			super(version, data);
+			this.document = document;
+		}
+
+		/**
+		 * Apply the edits from the server, provided the document is unchanged since the request used
+		 * to compute the edits
+		 *
+		 * @throws BadLocationException
+		 * @throws ConcurrentModificationException
+		 */
+		public void apply() throws BadLocationException, ConcurrentModificationException {
+			if (getVersion() != LSPEclipseUtils.getDocumentModificationStamp(this.document)) {
+				throw new ConcurrentModificationException();
+			} else {
+				LSPEclipseUtils.applyEdits(this.document, get());
+			}
+		}
+
+ 	}
+
+	/**
+	 * Bundles together a result from a language server with the document version at the time it was run.
+	 * Supports optimistic locking.
+	 *
+	 * @param <T>
+	 */
+	public static class Versioned<T> {
+		private final long version;
+
+		private final T data;
+
+		public Versioned(final long version, final T data) {
+			this.version = version;
+			this.data = data;
+		}
+
+		public T get() {
+			return data;
+		}
+
+		public long getVersion() {
+			return version;
+		}
+	}
+
+
 
 	private static <T> boolean isEmpty(final T t) {
 		return t == null || ((t instanceof List) && ((List<?>)t).isEmpty());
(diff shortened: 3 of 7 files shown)


sebthom:
I gave your branch a try, two observations:

a) batch formatting became extremely slow, like one second per file.
b) this error was logged a lot during formatting:
```
an 25, 2023 1:39:42 PM org.eclipse.lsp4j.jsonrpc.RemoteEndpoint fallbackResponseError
SEVERE: Internal error: java.lang.NullPointerException: Cannot invoke "org.eclipse.lsp4j.ServerCapabilities.getCodeActionProvider()" because "this.serverCapabilities" is null
java.util.concurrent.CompletionException: java.lang.NullPointerException: Cannot invoke "org.eclipse.lsp4j.ServerCapabilities.getCodeActionProvider()" because "this.serverCapabilities" is null
	at java.base/java.util.concurrent.CompletableFuture.encodeThrowable(CompletableFuture.java:315)
	at java.base/java.util.concurrent.CompletableFuture.completeThrowable(CompletableFuture.java:320)
	at java.base/java.util.concurrent.CompletableFuture$AsyncRun.run(CompletableFuture.java:1807)
	at java.base/java.util.concurrent.CompletableFuture$AsyncRun.exec(CompletableFuture.java:1796)
	at java.base/java.util.concurrent.ForkJoinTask.doExec(ForkJoinTask.java:373)
	at java.base/java.util.concurrent.ForkJoinPool$WorkQueue.topLevelExec(ForkJoinPool.java:1182)
	at java.base/java.util.concurrent.ForkJoinPool.scan(ForkJoinPool.java:1655)
	at java.base/java.util.concurrent.ForkJoinPool.runWorker(ForkJoinPool.java:1622)
	at java.base/java.util.concurrent.ForkJoinWorkerThread.run(ForkJoinWorkerThread.java:165)
Caused by: java.lang.NullPointerException: Cannot invoke "org.eclipse.lsp4j.ServerCapabilities.getCodeActionProvider()" because "this.serverCapabilities" is null
	at org.eclipse.lsp4e.LanguageServerWrapper.lambda$16(LanguageServerWrapper.java:928)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1511)
	at org.eclipse.lsp4e.LanguageServerWrapper.registerCapability(LanguageServerWrapper.java:881)
	at org.eclipse.lsp4e.LanguageClientImpl.lambda$2(LanguageClientImpl.java:136)
	... 6 more
```

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/format/LSPFormatFilesHandler.java
```diff
@@ -81,15 +81,21 @@ protected void formatFile(final @NonNull IFile file, final IProgressMonitor moni
 				return;
 
 			monitor.setTaskName(NLS.bind(Messages.LSPFormatFilesHandler_FormattingFile, file.getFullPath()));
-			final VersionedFormatRequest formatRequest = formatter.versionedRequestFormatting(doc,
-					new TextSelection(0, 0));
+			final var formatting = formatter.requestFormatting(doc,
```
sebthom, line 84: My guess for the NPEs is, that the formatter does not wait correctly for an initialized LS. which may not be an issue if you do formatting of an open text document (where the LS is already loaded) but in case if this batch formatting handler, there might not be any LS running already. That is why in #398 I tried to run the formatter within the `computeFirst` callback.

ahmedneilhussain: > I gave your branch a try, two observations:
> 
> a) batch formatting became extremely slow, like one second per file. b) this error was logged a lot during formatting:
> 
> ```
> an 25, 2023 1:39:42 PM org.eclipse.lsp4j.jsonrpc.RemoteEndpoint fallbackResponseError
> SEVERE: Internal error: java.lang.NullPointerException: Cannot invoke "org.eclipse.lsp4j.ServerCapabilities.getCodeActionProvider()" because "this.serverCapabilities" is null
> java.util.concurrent.CompletionException: java.lang.NullPointerException: Cannot invoke "org.eclipse.lsp4j.ServerCapabilities.getCodeActionProvider()" because "this.serverCapabilities" is null
> 	at java.base/java.util.concurrent.CompletableFuture.encodeThrowable(CompletableFuture.java:315)
> 	at java.base/java.util.concurrent.CompletableFuture.completeThrowable(CompletableFuture.java:320)
> 	at java.base/java.util.concurrent.CompletableFuture$AsyncRun.run(CompletableFuture.java:1807)
> 	at java.base/java.util.concurrent.CompletableFuture$AsyncRun.exec(CompletableFuture.java:1796)
> 	at java.base/java.util.concurrent.ForkJoinTask.doExec(ForkJoinTask.java:373)
> 	at java.base/java.util.concurrent.ForkJoinPool$WorkQueue.topLevelExec(ForkJoinPool.java:1182)
> 	at java.base/java.util.concurrent.ForkJoinPool.scan(ForkJoinPool.java:1655)
> 	at java.base/java.util.concurrent.ForkJoinPool.runWorker(ForkJoinPool.java:1622)
> 	at java.base/java.util.concurrent.ForkJoinWorkerThread.run(ForkJoinWorkerThread.java:165)
> Caused by: java.lang.NullPointerException: Cannot invoke "org.eclipse.lsp4j.ServerCapabilities.getCodeActionProvider()" because "this.serverCapabilities" is null
> 	at org.eclipse.lsp4e.LanguageServerWrapper.lambda$16(LanguageServerWrapper.java:928)
> 	at java.base/java.util.ArrayList.forEach(ArrayList.java:1511)
> 	at org.eclipse.lsp4e.LanguageServerWrapper.registerCapability(LanguageServerWrapper.java:881)
> 	at org.eclipse.lsp4e.LanguageClientImpl.lambda$2(LanguageClientImpl.java:136)
> 	at java.base/java.util.concurrent.CompletableFuture$AsyncRun.run(CompletableFuture.java:1804)
> 	... 6 more
> ```
Thanks - I'll take a look. Can you give me any pointers as to how to use this class so I can try and reproduce the problems?

sebthom: just right click on a source folder of a project that uses a LS and select "source/format"

ahmedneilhussain: > My guess for the NPEs is, that the formatter does not wait correctly for an initialized LS. which may not be an issue if you do formatting of an open text document (where the LS is already loaded) but in case if this batch formatting handler, there might not be any LS running already. 

Hmm both new and old API are supposed to have the property that actions should wait for the running LS - that property, in conjunction with the fact we mostly lazily start LS on demand accounts for a lot of the complexity!

The pseudocode for how LSP4e mostly works is `getRunningLS().thenDoStuff()`.

sebthom: It was just a wild guess, but I am not seeing these exception with #398.

ahmedneilhussain: Are you running batch processing with a very short idle timeout and no documents open in the IDE? LSP4e is centred around documents open in the IDE - the lifecycle for the LS times it out and shuts it down (temporarily - it can restart) a certain time after the last open document attached the LS has closed.

This could explain both the timeout and the errors: if you run a batch command on a document without it being open in the IDE then the LS could get shut down immediately after the operation. If the LS gets lazily started then shut down per document in batch then that would explain the terrible throughput.

ahmedneilhussain: The code for LSPDiagnosticsToMarkers has special logic related to this kind of thing IIRC - when LS is launched for a project (triggered by a document opening) the LS can scan the whole project and send back diagnostics for files not open in the IDE, so LSP4e opens the document temporarily to create the markers.

ahmedneilhussain: Thanks, I've tried running using Rust as a LS formatting a project without any files open in the IDE and can see a lot of this sort of problem and symptoms of the server starting up and shutting down all the time. I think this is a more general problem with batch processing and the idle timeout - there's nothing to keep the server up. I'll check the behaviour on master to see how it looks there, but I'd be surprised if it wasn't an issue without my changes (I may have moved the symptoms).

If you keep open a file from within the project you are batch-formatting does it behave better?

sebthom: yes if I have one file open it is better or if I use the apporach of #398

ahmedneilhussain:
@sebthom @rubenporras what do you think of the latest commit? I added support for a 'keep-alive' on the LanguageServerWrapper and wired it in. It wasn't what I was initially planning to do today but it seems to work (on my machine...)!

It's a bit of a thrown-together solution. I thought about having a `BatchToken` class or something like that with a `done()` method, but that puts the onus on the user to use try-with-resources correctly. A `WeakHashMap` is a bit tricksy but doesn't have that issue. On the other hand, it might keep the server alive for longer than necessary if there is a lot of memory and a GC doesn't happen. I don't think that's usually an issue in a busy IDE though.

ahmedneilhussain:
Doesn't belong in this PR as such either. I can separate it out but it might be easier for you to experiment with @sebthom if I leave it in initially?

sebthom:
Couldn't we just keep all LSs by default running for e.g. 2-4 seconds after the last operation. is there the need to instantly close them with the last request?

rubenporras:
> @sebthom @rubenporras what do you think of the latest commit? I added support for a 'keep-alive' on the LanguageServerWrapper and wired it in. It wasn't what I was initially planning to do today but it seems to work (on my machine...)!
> 
> It's a bit of a thrown-together solution. I thought about having a `BatchToken` class or something like that with a `done()` method, but that puts the onus on the user to use try-with-resources correctly. A `WeakHashMap` is a bit tricksy but doesn't have that issue. On the other hand, it might keep the server alive for longer than necessary if there is a lot of memory and a GC doesn't happen. I don't think that's usually an issue in a busy IDE though.



> @sebthom @rubenporras what do you think of the latest commit? I added support for a 'keep-alive' on the LanguageServerWrapper and wired it in. It wasn't what I was initially planning to do today but it seems to work (on my machine...)!
> 
> It's a bit of a thrown-together solution. I thought about having a `BatchToken` class or something like that with a `done()` method, but that puts the onus on the user to use try-with-resources correctly. A `WeakHashMap` is a bit tricksy but doesn't have that issue. On the other hand, it might keep the server alive for longer than necessary if there is a lot of memory and a GC doesn't happen. I don't think that's usually an issue in a busy IDE though.

I think it is better to just set the default to five seconds. In the end you want to avoid the behaviour as well if the user is just opening/closing files on the UI one by one. I have prepared https://github.com/eclipse/lsp4e/pull/411, so that we can keep it outside this PR. If we still want to have something else, I would say, we do it anyway in a different PR.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServers.java
```diff
@@ -282,6 +313,62 @@ public LanguageServerProjectExecutor excludeInactive() {
 		}
 	}
 
+	/**
+	 * Specialised POJO for document edits specifically
+	 *
+	 */
+	public static class VersionedEdits extends Versioned<List<? extends TextEdit>> {
```
rubenporras, line 320: Can we extract the class `VersionedEdits` out of `LanguageServers` into `LSPFormatter`? I think we should keep this class generic.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/format/LSPFormatFilesHandler.java
```diff
@@ -81,15 +82,21 @@ protected void formatFile(final @NonNull IFile file, final IProgressMonitor moni
...
+			UI.getDisplay().syncExec(() -> {
+				try {
+					formatting.get().apply();
+				} catch (ConcurrentModificationException | BadLocationException e) {
+					LanguageServerPlugin.logError(e);
+
+				}
+			});
 			docProvider.changed(doc);
 			docProvider.saveDocument(monitor, file, doc, true);
```
rubenporras, line 101: Since we have optional instead of return and then get() on the optional, I think it is better to use ifPresent:
```
			formatting.ifPresent(edits -> {
				docProvider.aboutToChange(doc);
				UI.getDisplay().syncExec(() -> {
					try {
						edits.apply();
					} catch (ConcurrentModificationException | BadLocationException e) {
					}
				});
				docProvider.changed(doc);
				try {
					docProvider.saveDocument(monitor, file, doc, true);
				} catch (CoreException e) {
				}
			});

```

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServers.java
ahmedneilhussain: I moved it here because `VersionedEdits` *is* generic: other operations can produce edits so would benefit from this (eventually, when they are ported over...)

Having it in the formatter class ties it to formatting, and it isn't really specific to formatting.

I did ponder making it a top-level class rather than a nested class within `LanguageServers`. What do you think?

ahmedneilhussain:
> > Couldn't we just keep all LSs by default running for e.g. 2-4 seconds after the last operation. is there the need to instantly close them with the last request? For the formatting maybe even just 1 second may be enough.
> 
> Well, why be simple when you can be complicated ;-) I didn't know if there was a reason for having the timeout of zero as a possibility, let alone the default. Having a longer default timeout is still a little bit arbitrary, but would probably be fine in practice.

I see @rubenporras is already submitting a longer default timeout in another PR so I'll get rid of the batch keep alive stuff from this one.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/format/LSPFormatFilesHandler.java
```diff
@@ -81,15 +82,21 @@ protected void formatFile(final @NonNull IFile file, final IProgressMonitor moni
 				return;
 
 			monitor.setTaskName(NLS.bind(Messages.LSPFormatFilesHandler_FormattingFile, file.getFullPath()));
-			final VersionedFormatRequest formatRequest = formatter.versionedRequestFormatting(doc,
-					new TextSelection(0, 0));
+			final var formatting = formatter.requestFormatting(doc,
```
rubenporras, line 85: since the type is not obvious, I would prefer that we keep Optional<VersionedEdits>

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/format/LSPFormatFilesHandler.java
```diff
@@ -81,15 +82,21 @@ protected void formatFile(final @NonNull IFile file, final IProgressMonitor moni
...
+					new TextSelection(0, 0), batchToken).get();
```
rubenporras, line 86: should we not have a timeout on this get() operation?

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LSPEclipseUtils.java
```diff
@@ -1244,4 +1245,19 @@ public static boolean hasCapability(final @Nullable Either<Boolean, ? extends Ob
...
+ 	 *
+ 	 * In practice just a sanity-checked downcast of a legacy API: should expect the platfom to be instantiating
+ 	 * Documents that implement the later interfaces.
+ 	 *
+ 	 * Should be called on UI thread
+ 	 *
+ 	 * @param document Document to check
+ 	 * @return Opaque version stamp, or -1 if not available
+ 	 */
+ 	public static long getDocumentModificationStamp(@Nullable IDocument document) {
```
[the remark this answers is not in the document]
ahmedneilhussain: OK.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/format/LSPFormatFilesHandler.java
ahmedneilhussain: Sure

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/format/LSPFormatFilesHandler.java
ahmedneilhussain: One for @sebthom : his original submission just used `.get()` so I didn't change that. https://github.com/eclipse/lsp4e/blob/53defa0606b10b5a87d74e5cc5e41a6c1ae0e800/org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/format/LSPFormatFilesHandler.java#L87

ahmedneilhussain: Happy to put one in if you both agree.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/format/LSPFormatter.java
```diff
@@ -25,94 +23,63 @@
...
-		if (getLSPDocumentInfo(infos).getVersion() != version) {
-			throw new ConcurrentModificationException();
-		} else {
-			try {
-				LSPEclipseUtils.applyEdits(document, edits);
-			} catch (BadLocationException e) {
-				LanguageServerPlugin.logError(e);
+		DocumentFormattingParams params = new DocumentFormattingParams();
+		params.setTextDocument(docId);
+		params.setOptions(formatOptions);
```
rubenporras, line 61: Could we extract this to a method, for example `createDocumentFormattingParams`?

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LSPEclipseUtils.java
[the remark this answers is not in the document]
ahmedneilhussain: I've moved it, but need to keep it public for the test, unfortunately. I could duplicate the logic in the test I guess (it's only a 1-line function)...?

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServers.java
rubenporras: Yes, I think it makes sense.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/format/LSPFormatter.java
```diff
@@ -25,94 +23,77 @@
...
-			} catch (BadLocationException e) {
-				LanguageServerPlugin.logError(e);
+		// TODO: Could refine this algorithm: at present this grabs the first non-null response but the most functional
+		// implementation (if a text selection is present) would try all the servers in turn to see if they supported
+		// range formatting, falling back to a full format if unavailable
+		return executor.computeFirst((w, ls) -> {
+			final ServerCapabilities capabilities = w.getServerCapabilities();
+			if (isDocumentRangeFormattingSupported(capabilities)
+					&& (!isDocumentFormattingSupported(capabilities)
+							|| textSelection.getLength() != 0)) {
```
rubenporras, line 61: I think it would be slightly easier to understand if we write && !(textSelection.getLength() == 0 && isDocumentFormattingSupported(capabilities))

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/format/LSPFormatter.java
```diff
@@ -25,94 +23,77 @@
...
-		} else {
-			try {
-				LSPEclipseUtils.applyEdits(document, edits);
-			} catch (BadLocationException e) {
-				LanguageServerPlugin.logError(e);
+		// TODO: Could refine this algorithm: at present this grabs the first non-null response but the most functional
+		// implementation (if a text selection is present) would try all the servers in turn to see if they supported
+		// range formatting, falling back to a full format if unavailable
+		return executor.computeFirst((w, ls) -> {
+			final ServerCapabilities capabilities = w.getServerCapabilities();
```
rubenporras, line 58: I am not sure how to improve the code, but I do not like that we need to evaluate the capabilities twice, and prepare the parameters for document and range formatting in advance.
Would it be better to use first one executor with capability filter as in the if, call computeFirst for rangeFormatting on that executor, and only if the list is empty then prepare a second executor for documentFormatting?

What do you think, is that doable?

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServers.java
```diff
@@ -211,6 +218,14 @@ public static class LanguageServerDocumentExecutor extends LanguageServers<Langu
...
 		}
 
+		public VersionedEdits toVersionedEdits(List<? extends TextEdit> edits) {
+			return new VersionedEdits(startVersion, edits, document);
+		}
+
+		public <T> Versioned<T> toVersioned(T t) {
+			return new Versioned<>(startVersion, t);
+		}
+
```
rubenporras, line 228: I would move this methods to `VersionedEdits` and `Versioned` as factory methods that accept the executor?

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServers.java
```diff
@@ -360,6 +384,21 @@ private <T> void completeEmptyOrWithException(final CompletableFuture<Optional<T
...
+	 *
+	 * In practice just a sanity-checked downcast of a legacy API: should expect the platfom to be instantiating
+	 * Documents that implement the later interfaces.
+	 *
+	 * Should be called on UI thread
+	 *
+	 * @param document Document to check
+	 * @return Opaque version stamp, or -1 if not available
+	 */
+	public static long getDocumentModificationStamp(@Nullable IDocument document) {
```
rubenporras, line 398: should this not be private?

org.eclipse.lsp4e.test/src/org/eclipse/lsp4e/test/LanguageServersTest.java
```diff
@@ -778,6 +789,81 @@ public void testGetDocument() throws Exception {
 
 		assertEquals(document, executor.getDocument());
 	}
+	
+	@Test
+	public void testVersionSupportSuccess() throws Exception {
```
rubenporras, line 794: Since we have moved `VersionedEdits` out of `LanguageServers`, I would create as well a new test class for this two methods

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LSPEclipseUtils.java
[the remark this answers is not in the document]
rubenporras: After looking at it in more detail, I do not think it is actually a line function :( 

It should rather be
```
 		if (document instanceof IDocumentExtension4 ext) {
 			ext.getModificationStamp();
 		} else if (document != null){
 			IFile file = LSPEclipseUtils.getFile(document);
 			if (file != null) {
 				return file.getModificationStamp();
 			}
 		}
 		return -1;
```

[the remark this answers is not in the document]
rubenporras: I think I would make it private, and for the test, just cast the document to an IDocumentExtension4, if that passes the test.

It looks like we are missing a package to put internal utilities, but it is probably not worth for this little utility function.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/format/LSPFormatFilesHandler.java
rubenporras: We need to put one if we do not want to risk the thread hanging forever, don't we? Recently all the get() where replaced in the code base with get(timeout), so I think that is the standard unless there are very good reasons

rubenporras:
@ahmedneilhussain , I am finish with the review now. Thanks for addressing my comments on the go so quick.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/Versioned.java
```diff
@@ -0,0 +1,38 @@
...
+ *******************************************************************************/
+package org.eclipse.lsp4e;
+
+/**
+ * Bundles together a result from a language server with the document version at the time it was run.
+ * Supports optimistic locking.
+ *
+ * @param <T>
+ */
+public class Versioned<T> {
```
rubenporras, line 21: I have remembered that I have remembered that org.eclipse.lsp4e.operations.semanticTokens.SemanticHighlightReconcilerStrategy also has checks to skip part of the work if the document is changed in between request and its response.
Shall we name this class VersionedDocumentOperation and have a constructor such as
`public Versioned(final IDocument document, final T data)`
and move `getDocumentModificationStamp` inside the class and make in private?

rubenporras:
Also, could you take a look `SemanticHighlightReconcilerStrategy` and see how one would use the new versioned API introduced here? It would be nice if you could add it to this PR, it should not be a big change because I have already migrated it to your new `LanguageServers` API. Or if you prefer to keep this PR simple, just taking a brief look to see that it can be adapted to the versioned API easily would be enough, I can do the adaptations later.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServers.java
ahmedneilhussain: I think it works better (is more compact when used) as instance methods on the executor - a static factory would need two arguments, one for the result and one for the executor.

So
```
		var ex = LanguageServers.forDocument(doc).withCapability(ServerCapabilities::getDocumentFormattingProvider);
		
		var result = ex.computeFirst(ls -> ls.getTextDocumentService().formatting(params).thenApply(ex::toVersionedEdits));
```
would become
```
		var result = ex.computeFirst(ls -> ls.getTextDocumentService().formatting(params).thenApply(edits -> VersionedEdits.toVersionedEdits(edits, ex));
```
I think it is neater with the method reference rather than the lambda?

rubenporras: I believe in the end we need to have version checks for each request type, as recommended in 
https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#implementationConsiderations
where it recommends to check for outdated response in general (that is, for any response).

So yes, I understand that the code gets slight more verbose, but what do we do when we need versioned request for semantic highlight? Add it to the executor as well? And when we need it for hover? Add it there as well? We would end up with one method in the `LanguageServer` for each possible request type, right?

I would prefer to have those more local to the operations that use each request type, together with the subclasses of the `Versioned<T>` class.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/format/LSPFormatter.java
ahmedneilhussain: Would you mind if we did that as a future refinement? This implementation mirrors the existing behaviour. I need to get this initial port done as my boss wants me to investigate the resource exhaustion issue I raised yesterday, as it's having customer impact.

ahmedneilhussain: Evaluating the capabilities and creating the parameters is going to be very cheap. It's more of an issue that if the first LS to respond only supports full formatting then it will beat another server that could have done a more sophisticated format - so for that reason your change makes sense (I had given it some thought already). But at least the first to respond is a bit less arbitrary than just the first one in the list, as the old code did!

In practice, I doubt there will really be multiple LS offering formatting for a document. As we discussed before with Mickael, it's more likely that there will be a primary LS providing most of the brains, and *possibly* one or two others offering things like SonarCube integration.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LSPEclipseUtils.java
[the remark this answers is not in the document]
ahmedneilhussain: I have adopted your improved algorithm that falls back to the file stamp. In practice I've not seen a case  where the cast to `IDocumentExtension4` failed, as I suspect it'd only be legacy Eclipse code that would construct such a thing. However there are some corner cases to do with LS handling files outside Eclipse (that I've not had much to do with and don't fully understand), so I think your approach is much better.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServers.java
ahmedneilhussain: Ah, I see. I guess it depends on whether `Versioned<T>` (which is after all, just generic so can associated any result data with an initial timestamp) is likely to need more specialised subclasses which would be specific to the consuming class.

I think `VersionedEdits` is a special case that is likely to be all that is required for most requests that result in edits.

As a compromise, I'll create factory methods as you suggest but keep the `toVersionedEdits` on `LanguageServerDocumentExecutor` as a convenience. It's ok to have syntactic sugar for the common case, right? ;-)

rubenporras: I am not sure it will be much used, but fine for me :).

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/format/LSPFormatter.java
rubenporras: > Would you mind if we did that as a future refinement? This implementation mirrors the existing behaviour. I need to get this initial port done as my boss wants me to investigate the resource exhaustion issue I raised yesterday, as it's having customer impact.

Yes, that is fine for me, this is good enough, so let us merge it, and let us improve it. I can have a look in the meanwhile at the semantic highlight adaptation.

rubenporras (approved):
@ahmedneilhussain , shall I merge it?

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServers.java
ahmedneilhussain: Package-private now (`VersionedEdits.apply()` needs it)

org.eclipse.lsp4e.test/src/org/eclipse/lsp4e/test/LanguageServersTest.java
ahmedneilhussain: OK

ahmedneilhussain:
> @ahmedneilhussain , shall I merge it?

Hi - if you're OK with my latest changes, yes please. I think I've made all the changes we agreed on... as I said I need to take a look at the resource issue (and also look into cancel support) so would be good if we could pick up the semantic highlight strategy stuff in a few days if you haven't had a chance till then.

rubenporras (approved):

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/format/LSPFormatFilesHandler.java
```diff
@@ -81,17 +85,25 @@ protected void formatFile(final @NonNull IFile file, final IProgressMonitor moni
...
-			List<? extends TextEdit> formatEdits = formatRequest.edits().get();
-			if (formatEdits.isEmpty())
-				return;
-
-			docProvider.aboutToChange(doc);
-			UI.getDisplay().syncExec(() -> formatter.applyEdits(doc, formatEdits, formatRequest.version()));
-			docProvider.changed(doc);
-			docProvider.saveDocument(monitor, file, doc, true);
+			final Optional<VersionedEdits> formatting = formatter.requestFormatting(doc,
+					new TextSelection(0, 0)).get(SINGLE_FILE_TIMEOUT_MS, TimeUnit.MILLISECONDS);
```
rubenporras, line 89: note for the followup, we need to handle the TimeoutException

org.eclipse.lsp4e/src/org/eclipse/lsp4e/Versioned.java
```diff
@@ -0,0 +1,43 @@
...
+	private final long version;
+
+	private final T data;
+
+	public Versioned(final long version, final T data) {
```
mickaelistria, line 26: what version should that be here? The documentVersion from the Eclipse document? Please clarify that in name and javadoc.

PR #413, opened 2023-01-27
Port SemanticHighlight to getDocumentModificationStamp

rubenporras:


rubenporras:
@mickaelistria , okay with you to create a new package for `DocumentUtil`? I did not wanted to put the new method in `LSPEclipseUtils` as it is not related to the Language Server Protocol, and also if we make it public it is harder to do refactors, so I decided to create a place for utilities which we do not wish to export in the MANIFEST yet.

mickaelistria:
> okay with you to create a new package for DocumentUtil? I did not wanted to put the new method in LSPEclipseUtils as it is not related to the Language Server Protocol

Fine by me.

PR #420, opened 2023-01-30
[DO NOT MERGE] Diagnostics: Add torture test to provoke handle exhaustion

ahmedneilhussain:
Do not merge - will cause hang!

Abused the existing diagnostics UT to show that a large(ish) number of files with diagnostics + 'run in background' disabled + windows = handle exhaustion (max number of OS shells exceeded).


ahmedneilhussain:
Fix in #471 has been approved.

PR #422, opened 2023-01-30
Port SemanticHighlight to Versioned request pattern

rubenporras:


rubenporras:
@ahmedneilhussain , is this transformation to the Versioned<T> type what you would expect? Does the new code makes sense to you, from the readability point of view?

I was thinking on adding the `exceptionally` as an option the builder we offer in `LanguageServers`. Do you think as well that it would make sense?

ahmedneilhussain:
> @ahmedneilhussain , is this transformation to the Versioned type what you would expect? Does the new code makes sense to you, from the readability point of view?
> 
> I was thinking on adding the `exceptionally` as an option the builder we offer in `LanguageServers`. Do you think as well that it would make sense?

Looks reasonable to me. I haven't thought through the implications, but I'm not sure how an `exceptionally` handler stored on the executor would look - will you run into generic typing issues?

I'm going to take a look at cancel support later today/tomorrow, which is closely related. I think I can see how to do it but I don't think it'll be totally seamless because of the way cancellation works with lsp4j.

If you add an exceptionally handler could you sketch out the idea as a standalone PR so we can talk through the implications, just in case it interacts awkwardly with my attempt to add cancel support?

Thanks!

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServers.java
```diff
@@ -382,7 +387,7 @@ private <T> Stream<CompletableFuture<T>> executeOnServers(
 	 * forever...
 	 */
 	private <T> void completeEmptyOrWithException(final CompletableFuture<Optional<T>> completableFuture, final Throwable t) {
-		if (t != null) {
+		if (t != null && !isRequestCancelledException(t)) {
```
rubenporras, line 390: @ahmedneilhussain , I think this will do for the case computeFirst, which is enough for this PR, but it would be nice if it would work the same way for collectAll and computeAll, but I am not quite sure on how to code it. Do you have good ideas for that?

rubenporras:
the failing test is unrelated to this PR

PR #423, opened 2023-01-30
Fix Workspace Symbols Dialog

ahmedneilhussain:
This was originally submitted as part of porting Workspace symbols to the new API (now submitted standalone).

In the process of doing the port I realised that the Workspace Symbols dialog appears to have been broken some time ago: the code to call the server was upgraded to handle the newer return type, and indeed to convert any legacy `SymbolInformation` instances returned into the newer `WorkspaceSymbol` lsp4j type, but the UI code was still assuming the objects could be downcast to `SymbolInformation`.... the two types are not directly interconvertible.

I've got it working but the unit test could do with upgrading. I was also unsure what else used the old `SymbolInformation` code branches and whether I could get rid of it. Can anyone comment?

### DIFF
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/outline/SymbolsLabelProvider.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/outline/SymbolsLabelProvider.java
@@ -49,6 +49,7 @@ import org.eclipse.lsp4j.DocumentSymbol;
 import org.eclipse.lsp4j.Range;
 import org.eclipse.lsp4j.SymbolInformation;
 import org.eclipse.lsp4j.SymbolKind;
+import org.eclipse.lsp4j.WorkspaceSymbol;
 import org.eclipse.lsp4j.jsonrpc.messages.Either;
 import org.eclipse.swt.graphics.Image;
 import org.eclipse.ui.IMemento;
@@ -127,6 +128,8 @@ public class SymbolsLabelProvider extends LabelProvider
 		Image res = null;
 		if (element instanceof SymbolInformation info) {
 			res = LSPImages.imageFromSymbolKind(info.getKind());
+		} else if (element instanceof WorkspaceSymbol symbol) {
+			res = LSPImages.imageFromSymbolKind(symbol.getKind());
 		} else if (element instanceof DocumentSymbol symbol) {
 			res = LSPImages.imageFromSymbolKind(symbol.getKind());
 		} else if (element instanceof DocumentSymbolWithFile symbolWithFile) {
@@ -135,6 +138,8 @@ public class SymbolsLabelProvider extends LabelProvider
 		IResource file = null;
 		if (element instanceof SymbolInformation symbol) {
 			file = LSPEclipseUtils.findResourceFor(symbol.getLocation().getUri());
+		} else if (element instanceof WorkspaceSymbol symbol) {
+			file = LSPEclipseUtils.findResourceFor(getUri(symbol));
 		} else if (element instanceof DocumentSymbolWithFile symbolWithFile) {
 			file = symbolWithFile.file;
 		}
@@ -145,6 +150,8 @@ public class SymbolsLabelProvider extends LabelProvider
 			Range range = null;
 			if (element instanceof SymbolInformation symbol) {
 				range = symbol.getLocation().getRange();
+			} else if (element instanceof WorkspaceSymbol symbol && symbol.getLocation().isLeft()) {
+				range = symbol.getLocation().getLeft().getRange();
 			} else if (element instanceof DocumentSymbol documentSymbol) {
 				range = documentSymbol.getRange();
 			} else if (element instanceof DocumentSymbolWithFile symbolWithFile) {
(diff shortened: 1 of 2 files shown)


rubenporras:
> In the process of doing the port I realised that the Workspace Symbols dialog appears to have been broken some time ago: the code to call the server was upgraded to handle the newer return type, and indeed to convert any legacy `SymbolInformation` instances returned into the newer `WorkspaceSymbol` lsp4j type, but the UI code was still assuming the objects could be downcast to `SymbolInformation`.... the two types are not directly interconvertible.
> 
> I've got it working but the unit test could do with upgrading. I was also unsure what else used the old `SymbolInformation` code branches and whether I could get rid of it. Can anyone comment?

Could we fix this in a separate PR so that the fix and the actual port to the new API are clearer and can be better discussed? I would propose we do first the fix of the dialog.

ahmedneilhussain:
> > In the process of doing the port I realised that the Workspace Symbols dialog appears to have been broken some time ago: the code to call the server was upgraded to handle the newer return type, and indeed to convert any legacy `SymbolInformation` instances returned into the newer `WorkspaceSymbol` lsp4j type, but the UI code was still assuming the objects could be downcast to `SymbolInformation`.... the two types are not directly interconvertible.
> > I've got it working but the unit test could do with upgrading. I was also unsure what else used the old `SymbolInformation` code branches and whether I could get rid of it. Can anyone comment?
> 
> Could we fix this in a separate PR so that the fix and the actual port to the new API are clearer and can be better discussed? I would propose we do first the fix of the dialog.

Sure @rubenporras  - do you mean keep this PR as is and do follow-up work in a second PR, or should I only submit the first commit in this PR and submit the second separately? Without the second commit one cannot even do a manual test, and there isn't a UI test that covers this dialog, but I agree it is not part of the API port.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/outline/SymbolsLabelProvider.java
```diff
@@ -335,4 +350,11 @@ public void preferenceChange(PreferenceChangeEvent event) {
 		}
 	}
 
+	private static String getUri(WorkspaceSymbol symbol) {
```
rubenporras, line 353: This can just be
`return symbol.getLocation().map(Location::getUri, WorkspaceSymbolLocation::getUri)`

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/symbols/LSPSymbolInWorkspaceDialog.java
```diff
@@ -128,34 +129,31 @@ protected void fillContentProvider(AbstractContentProvider contentProvider, Item
...
-			}
-		}
+		final var params = new WorkspaceSymbolParams(itemsFilter.getPattern());
+		this.executor.computeAll((w, ls) -> ls.getWorkspaceService()
+ 				.symbol(params)).stream().map(s -> s.thenApply(LSPSymbolInWorkspaceDialog::eitherToWorkspaceSymbols))
+ 			.forEach(cf -> {
+ 				if (monitor.isCanceled()) {
+ 					return;
+ 				}
+ 				try {
```
rubenporras, line 139: since you talked about cancellation support, I have also been thinking about cancellation support and the get with timeout and its corresponding exception handling. 

I wonder if it would fit nicely addressed together. Now we have a rather arbitrary timeout and it the timeout is not enough, the answer to the request will never be read (but the request is not cancelled).
Shall it not be something like
```
while ("respose still useful") {
   try {
     get response with timeout
     do something with response
     exit while loop
   } catch (ExecutionException e) {
   } catch (InterruptedException e) {
      Thread.currentThread().interrupt();
   }  catch (TimeoutException) {
      if (!"respose still useful") {
         cancel request
         exit while loop
      }
     // else do nothing, retry
   }
}
```
The main point, is that the user should never see a timeout exception, either the answer is still useful and the we wait for it, or it is not, and in that case it is canceled. Of course the java code can be prettier than my sketched code :)

This is of course not for this PR, but for the one when we introduce cancellation support.

rubenporras:
> > > In the process of doing the port I realised that the Workspace Symbols dialog appears to have been broken some time ago: the code to call the server was upgraded to handle the newer return type, and indeed to convert any legacy `SymbolInformation` instances returned into the newer `WorkspaceSymbol` lsp4j type, but the UI code was still assuming the objects could be downcast to `SymbolInformation`.... the two types are not directly interconvertible.
> > > I've got it working but the unit test could do with upgrading. I was also unsure what else used the old `SymbolInformation` code branches and whether I could get rid of it. Can anyone comment?
> > 
> > 
> > Could we fix this in a separate PR so that the fix and the actual port to the new API are clearer and can be better discussed? I would propose we do first the fix of the dialog.
> 
> Sure @rubenporras - do you mean keep this PR as is and do follow-up work in a second PR, or should I only submit the first commit in this PR and submit the second separately? Without the second commit one cannot even do a manual test, and there isn't a UI test that covers this dialog, but I agree it is not part of the API port.

I would say submit the first commit of this PR as a PR without the draft, and then we can quickly review and merge it.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/symbols/LSPSymbolInWorkspaceDialog.java
ahmedneilhussain: Cool, I hope to have something ready to submit in draft later today (I keep getting sidetracked by other LSP4e issues that crop up!) then we can talk through the implications.

ahmedneilhussain:
> I would say submit the first commit of this PR as a PR without the draft, and then we can quickly review and merge it.

Done in #426

org.eclipse.lsp4e/src/org/eclipse/lsp4e/outline/SymbolsLabelProvider.java
ahmedneilhussain: Ooh, nice, I was trying to figure out how `Either.map(..)` was supposed to be used! Thanks.

I thought of moving the code back inline (since you've shown it can be indeed done in one line), but I think it still reads better out of line.

ahmedneilhussain:
Marked as ready for review as in theory could be merged as is and would get the Workspace Symbols dialog working again.

I expect we should probably tidy up and e.g. extend the tests, though.

PR #426, opened 2023-01-31
New API: Port workspace symbols

ahmedneilhussain:
This reworks onto the merged version of the new API example code originally submitted in #333 

Note that the Workspace Symbols dialog is currently not rendering and hot-linking correctly because of changes to the LSP spec. Originally attempted to fix this as well in #423 but not strictly related to the new API so have agreed to submit the API port in this PR as a standalone change.

@rubenporras FYI


rubenporras (approved):
Thanks

PR #430, opened 2023-01-31
New API: Port hyperlink detection

ahmedneilhussain:
Port to new API of demo code originally done in #251 

@angelozerr FYI: I'm not trying to subvert your work in #238 ...! I've been thinking about how best to implement cancel support in the new API, and in particular how to make it as general as possible. I think I know roughly how to do it so I will try and get the general design in place and then see how it works in this specific instance. Doing the API port separately should hopefully make the discussion a bit easier so we can see exactly how cancel support interacts with everything else.


rubenporras (approved):
LGTM

PR #431, opened 2023-02-01
[FOR DISCUSSION] New API: Outline of cancellation support

ahmedneilhussain:
Hi all, this is a sketch of how cancellation support can be added to the new API.

According to the LSP spec, cancellation works by sending a special notification `cancel` message containing the request ID of the pending request you want to cancel.

This is supported by LSP4j - according to https://github.com/eclipse/lsp4j/blob/main/documentation/jsonrpc.md if you have made a request to the language server, e.g.

```
CompletableFuture<Hover> pending = ls.hover(params);
```

then you can just call `pending.cancel(true)` to cancel it. The LSP4j remoting layer constructs CompletableFuture objects with an overridden cancel method that knows the id it is associated with.

The slightly complicating factor is that this only applies to the initial response objects constructed directly by lsp4j. The fluent API provided by `CompletableFuture` for chaining further actions will construct further `CompletableFuture` objects that LSP4j knows nothing about, and cancellation flows down not up.

E.g. if you have, say:

```
CompletableFuture<A> a = doRequest();
CompletableFuture<B> b = a.thenApply({ do stuff to a});
CompletableFuture<C> c = b.thenApply({do more stuff});
```

If you call `b.cancel(true)` then `b` will be cancelled (if yet to return), as will `c`, but `a` will be unaffected and run to completion.

For us this has the effect that
```
CompletableFuture<Hover> pending1 = ls.hover(params);
CompletableFuture<TransformedHover> pending2 = ls.hover(params).thenApply(h -> new TransformedHover(h));
pending1.cancel(true);
pending2.cancel(true);
```
the first request will have a cancel message sent to the server, the second will not. `pending2` will still be cancelled, i.e. it will throw a `CancellationException` if you call `get()` on it, but the server will not be informed it can stop processing.

The extension to the new API I outline here works by wrapping the lambda function you would otherwise have passed directly to the API in a new 'do around' block that stores the completable future prior to returning it. For this to work, the wrapped function *must return the LSP4j-constructed objects directly without any chained post processing*.

So
```
var result = LanguageServers.forDocument(doc).collectAll(canceller.wrap(ls -> ls.hover(params)).thenApply(xform)));
```
will work.
```
var result = LanguageServers.forDocument(doc).collectAll(canceller.wrap(ls -> ls.hover(params).thenApply(xform)));
```
won't.

### DIFF
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/Canceller.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/Canceller.java
@@ -1,0 +1,68 @@
+package org.eclipse.lsp4e;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.concurrent.CancellationException;
+import java.util.concurrent.CompletableFuture;
+import java.util.function.BiFunction;
+import java.util.function.Function;
+
+import org.eclipse.lsp4j.services.LanguageServer;
+
+public class Canceller {
+
+	private boolean cancelled;
+
+	private List<CompletableFuture<?>> running = new ArrayList<>();
+
+	private final Object lock = new Object();
+
+	public Canceller() {}
+
+
+	public boolean isCancelled() {
+		synchronized (this.lock) {
+			return this.cancelled;
+		}
+	}
+
+	public void cancel() {
+		synchronized(this.lock) {
+			if (!this.cancelled) {
+				this.cancelled = true;
+				this.running.forEach(cf -> cf.cancel(true));
+			}
+		}
+	}
+
+	protected void checkCancelled() {
+		if (this.cancelled) {
+			throw new CancellationException();
+		}
+	}
+
+	public <T> BiFunction<? super LanguageServerWrapper, LanguageServer, ? extends CompletableFuture<T>> wrap(BiFunction<? super LanguageServerWrapper, LanguageServer, ? extends CompletableFuture<T>> fn) {
+
+		return (w, ls) -> {
+			synchronized (this.lock) {
+				checkCancelled();
+				CompletableFuture<T> pending = fn.apply(w,  ls);
+				this.running.add(pending);
+
+				return pending;
+			}
+		};
+	}
+
+	public <T> Function<LanguageServer, ? extends CompletableFuture<T>> wrap(Function<LanguageServer, ? extends CompletableFuture<T>> fn) {
+		return ls -> {
+			checkCancelled();
+			CompletableFuture<T> pending = fn.apply(ls);
+			this.running.add(pending);
+
+			return pending;
+
+		};
+	}
+
+}
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/Canceller.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/Canceller.java
@@ -46,23 +46,29 @@ public class Canceller {
 		return (w, ls) -> {
 			synchronized (this.lock) {
 				checkCancelled();
-				CompletableFuture<T> pending = fn.apply(w,  ls);
-				this.running.add(pending);
-
-				return pending;
+				return fn.apply(w,  ls);
 			}
 		};
 	}
 
 	public <T> Function<LanguageServer, ? extends CompletableFuture<T>> wrap(Function<LanguageServer, ? extends CompletableFuture<T>> fn) {
 		return ls -> {
-			checkCancelled();
-			CompletableFuture<T> pending = fn.apply(ls);
-			this.running.add(pending);
-
-			return pending;
-
+			synchronized (this.lock) {
+				checkCancelled();
+				return fn.apply(ls);
+			}
 		};
+	}
+
+	public <T> CompletableFuture<T> wrap(CompletableFuture<T> raw) {
+		synchronized(this.lock) {
+			if (this.cancelled) {
+				raw.cancel(true);
+			} else {
+				this.running.add(raw);
+			}
+			return raw;
+		}
 	}
 
 }
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServers.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServers.java
@@ -413,7 +413,44 @@ public abstract class LanguageServers<E extends LanguageServers<E>> {
 		return new LanguageServerProjectExecutor(project);
 	}
 
+	/**
+	 *
+	 * @return True if the server call made using this executor has been cancelled
+	 */
+	public boolean isCancelled() {
+		return canceller.isCancelled();
+	}
+
+	/**
+	 * Cancel any running requests that have been wrapped with cancellation support
+	 */
+	public void cancel() {
+		canceller.cancel();
+	}
+
+	/**
+	 * Wrap a call to the LS such that it can be cancelled while running
+	 * @param <T>
+	 * @param fn
+	 * @return
+	 */
+	public <T> BiFunction<? super LanguageServerWrapper, LanguageServer, ? extends CompletableFuture<T>> wrapCancellable(BiFunction<? super LanguageServerWrapper, LanguageServer, ? extends CompletableFuture<T>> fn) {
+		return canceller.wrap(fn);
+	}
+
+	/**
+	 * Wrap a call to the LS such that it can be cancelled while running
+	 * @param <T>
+	 * @param fn
+	 * @return
+	 */
+	public <T> Function<LanguageServer, ? extends CompletableFuture<T>> wrapCancellable(Function<LanguageServer, ? extends CompletableFuture<T>> fn) {
+		return canceller.wrap(fn);
+	}
+
 	private @NonNull Predicate<ServerCapabilities> filter = s -> true;
+
+	private Canceller canceller = new Canceller();
 
 	private boolean isRequestCancelledException(final Throwable throwable) {
 		if (throwable instanceof final CompletionException completionException) {
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServers.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServers.java
@@ -75,6 +75,7 @@ public abstract class LanguageServers<E extends LanguageServers<E>> {
 	@NonNull
 	public <T> CompletableFuture<@NonNull List<@NonNull T>> collectAll(BiFunction<? super LanguageServerWrapper, LanguageServer, ? extends CompletableFuture<T>> fn) {
 		computeVersion();
+		fn = canceller.wrap(fn);
 		final CompletableFuture<@NonNull List<T>> init = CompletableFuture.completedFuture(new ArrayList<T>());
 		return executeOnServers(fn).reduce(init, LanguageServers::add, LanguageServers::addAll)
 			// Ensure any subsequent computation added by caller does not block further incoming messages from language servers
@@ -113,9 +114,10 @@ public abstract class LanguageServers<E extends LanguageServers<E>> {
 	@NonNull
 	public <T> List<@NonNull CompletableFuture<@Nullable T>> computeAll(BiFunction<? super LanguageServerWrapper, LanguageServer, ? extends CompletableFuture<T>> fn) {
 		computeVersion();
+		final BiFunction<? super LanguageServerWrapper, LanguageServer, ? extends CompletableFuture<T>> wrappedFn = canceller.wrap(fn);
 		return getServers().stream()
 				.map(cf -> cf
-						.thenCompose(w -> w == null ? CompletableFuture.completedFuture(null) : w.executeImpl(ls -> fn.apply(w, ls)).thenApplyAsync(Function.identity())))
+						.thenCompose(w -> w == null ? CompletableFuture.completedFuture(null) : w.executeImpl(ls -> wrappedFn.apply(w, ls)).thenApplyAsync(Function.identity())))
 				.toList();
 	}
 
@@ -147,6 +149,7 @@ public abstract class LanguageServers<E extends LanguageServers<E>> {
 	 */
 	public <T> CompletableFuture<Optional<T>> computeFirst(BiFunction<? super LanguageServerWrapper, LanguageServer, ? extends CompletableFuture<T>> fn) {
 		computeVersion();
+		fn = canceller.wrap(fn);
 		final CompletableFuture<Optional<T>> result = new CompletableFuture<>();
 
 		// Dispatch the request to the servers, appending a step to each such that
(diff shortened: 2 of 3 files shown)


ahmedneilhussain:
I was going to submit further revisions showing how this proposal could be polished up, but I've been asked to get some other stuff done so I won't have time for a day or two.

Briefly, the first improvement would be to ditch `Canceller` as a separate class and just fold it into the base `LanguageServers` class.

- The user would still have to call the wrap function explicitly, however (these would be instance methods on `LanguageServers` that you would call similarly to calling `wrapVersioned`)
- Some logic/assertions would need to be added to make it clear that the executor instances should only be used once: i.e. you should only ever call one of `collectAll`, `computeAll`, `computeFirst`, and not reuse the same executor for subsequent requests; you should construct a new executor instead. That's sort of the contract anyway, given the way versioning support works, but it'd probably be good to make this explicit and throw an `IllegalStateException` or suchlike if the user tries to reuse an executor.
-

ahmedneilhussain:
I also pondered building in the cancellation support so that it was *always* provided, i.e. the user never had to wrap the lambda explicitly, it was done for every request as part of the framework.

For that to work, however, we'd need to supply overloads of the request methods that took in post-processing as a second parameter.

E.g. rather than

```
var ex = LanguageServers.forDocument(doc);
ex.collectAll(ex.wrapCancellable(ls -> ls.hover(params)).thenApply(xform));
```
the user would write
```
LanguageServers.forDocument(doc).collectAll(ls -> ls.hover(params), result -> result.thenApply(xform));
```
The API can automagically wrap the first argument with the cancellation stuff, but only if it returns the raw result from LSP4j and doesn't do any post-processing. Because there is still a need for users to be able to do this kind of post-processing inside of the aggregation and so on, they would have to supply the post-processing transform to `collectAll` for the API to apply for them.

ahmedneilhussain:
Thoughts welcome, @rubenporras @mickaelistria @angelozerr

ahmedneilhussain:
PS I haven't tested this rigorously, but I did try it using Corrosion, hovering over a Rust symbol and then quickly moving my mouse to hover over another symbol. I could definitely see the cancel notifications being sent in the lsp4j console logging.

ahmedneilhussain:
Note @angelozerr that `Canceller` doesn't implement `CancelChecker`: although it does have a similar feel, I don't think `CancelChecker` is actually intended for what we're doing.

We would need to worry about `CancelChecker` if we were *generating* responses, rather than consuming them. I.e. if we were implementing methods on the client interface for situations where the server was making a request to the *client*. This is less common in the LSP spec but definitely allowed. In that case, `CancelChecker` would be something we'd need to take into consideration if we wanted to support the server sending a cancel notification to us telling us it was no longer interested in the result we were computing.

And of course it would be relevant if we were *implementing* a language server in java using LSP4j.

Do you agree? The documentation around `CancelChecker` isn't very clear so it did cause me a bit of head-scratching trying to figure it out!!

mickaelistria:
Isn't it overall simpler to just keep the reference to the LS request future and to just invoke `.cancel()` on it?

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/hover/LSPTextHover.java
```diff
@@ -200,10 +202,12 @@ private void initiateHoverRequest(@NonNull ITextViewer viewer, int offset) {
 		this.lastViewer = viewer;
 		try {
 			HoverParams params = LSPEclipseUtils.toHoverParams(offset, document);
+			this.canceller.cancel();
+			this.canceller = new Canceller();
 
 			this.request = LanguageServers.forDocument(document)
 				.withCapability(ServerCapabilities::getHoverProvider)
-				.collectAll(server -> server.getTextDocumentService().hover(params));
+				.collectAll(canceller.wrap(server -> server.getTextDocumentService().hover(params)));
```
rubenporras, line 210: According to our recent discussions, we could do away with the timeout of the get operation if we can cancel, right @mickaelistria? If this is so I believe we can get rid of the timeout in the method `getHoverInfo` in line 73, because it is run on a different thread than the UI thread, see org.eclipse.jface.text.TextViewerHoverManager.computeInformation(). Do you agree?

mickaelistria: >  we could do away with the timeout of the get operation if we can cancel, right @mickaelistria? 

For cases where the UI is not frozen, yes; but here it's not the case, I believe the the hover request (via Eclipse API) does happen in UI Thread. If it is so then the hover strategy does need to be reworked in Platform and/or LSP4E to become non-UI blocking. Then we can consider removing the timeout.

rubenporras: For the case I tested on my eclipse, it does not happen on the UI thread. I put a breakpoint on line 73 where I propose to remove the timeout, and the UI was not froozen.

Also the code I reference is quite explicit about it in org.eclipse.jface.text.TextViewerHoverManager.computeInformation()

```
		fThread= new Thread("Text Viewer Hover Presenter") { //$NON-NLS-1$
			@Override
			public void run() {
				// http://bugs.eclipse.org/bugs/show_bug.cgi?id=17693
				boolean hasFinished= false;
				try {
					if (fThread != null) {
						Object information;
						try {
							if (hover instanceof ITextHoverExtension2)
								information= ((ITextHoverExtension2)hover).getHoverInfo2(fTextViewer, region);
							else
								information= hover.getHoverInfo(fTextViewer, region);
```

so at that point, the code does not run in the UI thread. Maybe there are other consumers that do not create a new thread, so we could check if our thread is the UI one or not, and depending on that put the timeout. Does it look like a plan?

I am very keen removing the timeout because we have complex DSLs, and we actually run into timeouts, so it is a real problem us.

angelozerr:
@ahmedneilhussain if you need to check that cancel support is working, you could debug LemMinx which cancel any operation on server side as soon as possible. It misses just the hove operation (we need to implement it), but for other operation like diagnostics, code action, completion, etc, cancel support is implemented.

I think this cancel support will b every helpful for lemminx-maven //cc @vrubezhny @mickaelistria

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/hover/LSPTextHover.java
mickaelistria: OK, you're right. So yes, proper support for cancelling would make it OK to remove timeout in hover.

ahmedneilhussain:
I've embedded cancellation directly in the executor, rather than requiring a whole separate object. That simplifies the usage a bit - you have an executor object that can run a request, and also cancel it for you. I'll add a further refinement showing how the cancellation support can be made automatic, i.e. no explicit wrap stage needed.

Like versioning support, this makes the executor stateful. I think this is fine (given the gain in functionality), but I think it should be made explicit - the executor can only be used once without an explicit reset. See discussion on https://github.com/eclipse/lsp4e/pull/434#issuecomment-1425636214

mickaelistria:
It's still not clear to me why a new API for cancellation is necessary when CompletableFuture already offers .cancel?

mickaelistria:
IMO, instead of a new API, we should just retain the LSP request futures and use `cancel` on those. This will cascade to cancelling downstream futures.
For the futures that originate from LanguageServers API (which are combined with eg `.allOf()`), those futures can be made to react to cancellation with `combinedFuture.exceptionally(e -> if (e instanceof CancellationException) children.forEach(CompletableFuture::cancel))`.
With an extra API, we end up adding complexity for a problem that's already solved properly by the underlying framework.

ahmedneilhussain:
The other issue is that `cancel()` on the completable futures requires the requests to have been sent - if you decide to cancel before the initial request has been sent then you obviously don't want to send the request only to cancel it.

This is a complicating factor that @angelozerr addressed in #238 that I have also tried to take into account.

ahmedneilhussain:
> IMO, instead of a new API, we should just retain the LSP request futures and use `cancel` on those. This will cascade to cancelling downstream futures.

Well, that's what the `Canceller` object does, no? You need something to store the raw CFs returned by the LSP4j layer for when the consumer wants to cancel. 

> For the futures that originate from LanguageServers API (which are combined with eg `.allOf()`), those futures can be made to react to cancellation with `combinedFuture.exceptionally(e -> if (e instanceof CancellationException) children.forEach(CompletableFuture::cancel))`.
> With an extra API, we end up adding complexity for a problem that's already solved properly by the underlying framework.

Well, it's tricky to get right and anyone wanting cancellation support is going to need to do something similar, so IMHO an API should take care of low-level boilerplate and make it easy for consumers to get right without having to worry about details.

Can you maybe sketch out how you think it would look in practice without direct API support @mickaelistria ?

You didn't comment on #238 that it was all unnecessary! At the very least, we don't have any documentation on how to use LSP4j's cancel support from within LSP4e, and I don't think there is any existing code that does it...

ahmedneilhussain:
More to show what it would look like, in https://github.com/eclipse/lsp4e/pull/431/commits/b1ee38902ac8aa1fae79ac80ed9cd0bb14298033 I have added 'automagic' cancellation support to `computeAll` and `collectAll` (I didn't yet bother to do `computeFirst`, although the same principle applies since it shares most of its code with `collectAll`), and added https://github.com/eclipse/lsp4e/pull/431/commits/25216f8838cc424f863ddea835423614fb423eb8 to show how calling code would have to change.

It'd be nice in principle if every request made through the API could in theory be cancelled easily, but there is a price to pay. If the framework is to support cancellation automatically then the `fn` supplied to the request methods must return CompletableFutures from the LSP4j layer directly, without composing any further processing on top. But there is also a need to do some post processing inside the aggregation (e.g. because you want to construct a `ColorInformationMining` and pass in the `LanguageServerWrapper`). So the request methods need to take an optional post-processing function to do this work *after* storing the raw LSP4j returned future but *before* the aggregation.

@angelozerr @rubenporras FYI. It's possible to make cancellation automatically available for every request, but only by forcing the calling code to split out any post processing like this. I think maybe this is maybe messier than every consumer having to call `wrapCancellable` or something like it explicitly. Maybe it depends how widely we really think cancellation will be used if it is 'easily' available? I don't think anything does at the moment. Maybe a significant fraction of the codebase would make use of it, but only a significant minority of calls would need it.

mickaelistria:
Thanks for the example, I can see how cancellation is made harder for objects that require to store some reference to the particular wrapper/language server that has returned them; and this pattern is indeed making things harder.
I think we should first try to simplify the case of more basic cancellation (eg Hover with plain collectAll) which doesn't need to rely on those more complex patterns, so that user could just invoke `request.cancel()` to cancel the LS request; and then once we're done with those, we'll see what API we can provide for those patterns. I do believe that the proposed postProcessor is actually a good proposal.

rubenporras:
@ahmedneilhussain , just let you know, I plan to take a look at this PR tomorrow/early next week. 
Regards

mickaelistria:
I've spent more time looking at it and I believe that it's better to let the client keep the `request` future apart to cancel it when needed (that will cancel both the request and downstream consumers).
With the current API, it's indeed quite tricky to keep a reference to the cancellable future. I think we need some new `CompletableFuture<List<U>> LanguageServers.collectAll(BiFunction<LanguageServerWrapper, LanguaegServer, CompletableFuture<T>> lsRequestor,  BiFunction<LanguageServerWrapper, T, U>  remapper)` that would allow processing cases where we still need a reference to the wrapper in the "terminal" objects while still being able to call `.cancel()` on this future to cancel all requests.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/Canceller.java
```diff
@@ -0,0 +1,74 @@
...
+import java.util.ArrayList;
+import java.util.List;
+import java.util.concurrent.CancellationException;
+import java.util.concurrent.CompletableFuture;
+import java.util.function.BiFunction;
+import java.util.function.Function;
+
+import org.eclipse.lsp4j.services.LanguageServer;
+
+public class Canceller {
```
mickaelistria, line 12: I think there is indeed a need for such logic somewhere when we create the "compound" futures from the request to propagate cancellation. In my draft attempt locally, I try to have it as a method, but a dedicated class may be a better fit.
Couldn't it be made package visible only (ie not public)?

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServers.java
```diff
@@ -413,8 +416,36 @@ private <T> void completeEmptyOrWithException(final CompletableFuture<Optional<T
 		return new LanguageServerProjectExecutor(project);
 	}
 
+	/**
+	 *
+	 * @return True if the server call made using this executor has been cancelled
```
mickaelistria, line 421: Does that mean that an executor can only run a single operation and then need to be discarded?

ahmedneilhussain:
Hi @mickaelistria @rubenporras @angelozerr what do you think of this as a compromise option?

I've simplified it a bit. The user now only has to wrap the *result* of calling the language server, not the whole function that gets passed to the executor.

Internally, the function does get wrapped, but this is only for the (minor, but still important for correctness) case where the user cancels before the function calling the LS reaches the front of the executor's queue, in which case we want to not send the request at all.

I quite like this. Making it totally transparent to the calling code is really hard (the solution where the user has to pass *two* lambda functions to the executor was a bit messy). So I think the best solution is to make the caller aware they need to cooperate in order for cancellation to work but make it nice and easy for them to do so.

Wrapping the *result* rather than the function means that the user can enchain postprocessing naturally.

```
ex.computeAll(ls -> ex.cancellable(ls.getTextDocumentServer().hover(params)).thenApply(hover -> toProcessedHover(hover));
```

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/hover/LSPTextHover.java
```diff
@@ -61,6 +62,8 @@ public class LSPTextHover implements ITextHover, ITextHoverExtension {
 	private static final MarkupParser MARKDOWN_PARSER = new MarkupParser(new MarkdownLanguage(true));
 	private static final int GET_TIMEOUT_MS = 1000;
 
+	private LanguageServerDocumentExecutor lastExecutor;
```
mickaelistria, line 65: Also, this shows that with this proposal, the client needs to keep a reference to the executor instead of a reference to the request future. It's not a clear win.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/Canceller.java
ahmedneilhussain: Yes, I don't think there is any need for this to be public API.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/hover/LSPTextHover.java
ahmedneilhussain: Indeed. It's not perfect.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/hover/LSPTextHover.java
```diff
@@ -200,10 +203,14 @@ private void initiateHoverRequest(@NonNull ITextViewer viewer, int offset) {
 		this.lastViewer = viewer;
 		try {
 			HoverParams params = LSPEclipseUtils.toHoverParams(offset, document);
+			if (this.lastExecutor != null) {
+				this.lastExecutor.cancel();
+			}
```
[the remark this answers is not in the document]
mickaelistria: > unless they really know how it all works and take great care.

I think it's fair to expect that people who use LSP4E as API need to learn about it, about LSP4J and CompletableFuture (and Streams, and Java...) and take great care in how to use those.
Adding wrappers or custom objects to the existing upstream APIs isn't really making things simpler for them: it does only add an extra layer of indirection (ie "put lipstick on a pig") that users need to learn about. ie someone who want cancellations will need to learn about storing the executor, wrapping in cancellable, and calling cancel on the executor; while the typical issue of cancelling futures already has solutions and patterns with LSP4J and CompletableFuture APIs.
This really looks like a problem where "less is more" to me. I find an extra layer of indirection much more difficult to maintain and use than just being able to cancel the futures returned by LanguageServers to cancel the underlying LS requests.

mickaelistria:
With recent changes in cancellation (latest being #777 , #778 and #780 ) cancelling the requests can be achieved without extra API, just invoking `.cancel()` on the futures returned by LanguageServers API.
So it seems like we don't need extra API. However, the various operation still need to keep a reference to the future and cancel it whenever necessary.

PR #432, opened 2023-02-01
New api/port code actions menu

ahmedneilhussain:
This ports QuickAssist and the code-actions menu to the new API.

Followed up @rubenporras idea for a single-server executor variant, but also incorporated the document. This makes `SingleServerDocumentExecutor` sort of a direct replacement for `DocumentInfo`.


org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServers.java
```diff
@@ -263,6 +264,42 @@ public long getStartVersion() {
...
+		SingleServerDocumentExecutor(final @NonNull IDocument document, final @NonNull LanguageServerWrapper serverWrapper) {
+			super(document);
+			this.serverWrapper = serverWrapper;
+		}
+
+		@Override
+		@SuppressWarnings("null")
+		protected @NonNull List<@NonNull CompletableFuture<@Nullable LanguageServerWrapper>> getServers() {
+			// Compute list of servers from document & filter
+			return Collections.singletonList(CompletableFuture.completedFuture(this.serverWrapper));
```
rubenporras, line 289: just style, but I think that `serverWrapper` reads better than `this.serverWrapper`

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServers.java
```diff
@@ -263,6 +264,42 @@ public long getStartVersion() {
...
+		public <T> CompletableFuture<T> execute(@NonNull Function<LanguageServer, ? extends CompletionStage<T>> fn) {
+			return this.serverWrapper.execute(fn);
+		}
+
+		public boolean supports(final Predicate<ServerCapabilities> predicate) {
+			return predicate.test(this.serverWrapper.getServerCapabilities());
+		}
+
+		public LanguageServerWrapper getServer() {
+			return this.serverWrapper;
```
rubenporras, line 301: do we actually need this method if we have already a public execute?

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServers.java
ahmedneilhussain: Sorry, force of habit!

ahmedneilhussain:
Hi @rubenporras how should I move this forward? Shall I rewrite it without the `LanguageServerDocumentExecutor`, since we are not so certain if it is needed or what API it should have?

rubenporras:
Hi @ahmedneilhussain , what about https://github.com/eclipse/lsp4e/pull/434#issuecomment-1428105434 or just passing the `LanguageServerWrapper` instead of a `LanguageServerDocumentExecutor`? I slowly get the impression that the wrapper is just good enough.

ahmedneilhussain:
> Hi @ahmedneilhussain , what about [#434 (comment)](https://github.com/eclipse/lsp4e/pull/434#issuecomment-1428105434) or just passing the `LanguageServerWrapper` instead of a `LanguageServerDocumentExecutor`? I slowly get the impression that the wrapper is just good enough.

Thanks @rubenporras , I've just gone with using LanguageServerWrapper directly: as it turned out the code looks OK like that.

What do you think?

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServers.java
```diff
@@ -300,7 +300,6 @@ public long getStartVersion() {
 			return this.startVersion;
 		}
 
-
```
rubenporras, line 303: can we remove LanguageServers from the PR? By now it is only a line which has been removed :)

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/codeactions/LSPCodeActionQuickAssistProcessor.java
```diff
@@ -158,22 +151,23 @@ public ICompletionProposal[] computeQuickAssistProposals(IQuickAssistInvocationC
...
+				// as a side effect of this request
+				futures = executor.computeAll((w, ls) -> ls.getTextDocumentService()
+						.codeAction(params)
+						.thenAccept(actions -> LanguageServers.streamSafely(actions)
+								.forEach(action -> processNewProposal(invocationContext, new CodeActionCompletionProposal(action, w)))));
 
 				CompletableFuture<?> aggregateFutures = CompletableFuture
 						.allOf(futures.toArray(new CompletableFuture[futures.size()]));
+
+				//
```
rubenporras, line 170: can we remove this comment?

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/codeactions/LSPCodeActionQuickAssistProcessor.java
```diff
@@ -158,22 +151,23 @@ public ICompletionProposal[] computeQuickAssistProposals(IQuickAssistInvocationC
...
-				futures = infos.stream()
-						.map(info -> info.getInitializedLanguageClient()
-								.thenComposeAsync(ls -> ls.getTextDocumentService().codeAction(params)
-										.thenAcceptAsync(actions -> actions.stream().filter(Objects::nonNull)
-												.map(action -> new CodeActionCompletionProposal(action, info))
-												.forEach(p -> processNewProposal(invocationContext, p)))))
-								.toList();
+				// Start all the servers computing actions - each server will append any code actions to the ongoing list of proposals
+				// as a side effect of this request
+				futures = executor.computeAll((w, ls) -> ls.getTextDocumentService()
```
rubenporras, line 162: could we not use here collectAll and then save as the trouble of calling 
```
CompletableFuture<?> aggregateFutures = CompletableFuture
						.allOf(futures.toArray(new CompletableFuture[futures.size()]))
```
right after?

rubenporras:
@ahmedneilhussain , I am finished with the review. Just some minor details. If you address them the extra space and the extra comment, (I think they can just reverted) and comment/address the other one, we can merge.
Thanks

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/codeactions/LSPCodeActionQuickAssistProcessor.java
ahmedneilhussain: I've enhanced the comment instead to try and clarify better what is going on (it isn't obvious)! See below.

rubenporras (approved):

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/codeactions/LSPCodeActionQuickAssistProcessor.java
ahmedneilhussain: No, because the intermediate (per-LS) futures are needed at line 178.

The logic for this function is not obvious, but this is one example (of many) where we have to compromise because we have to couple a 'push' (the asynchronous calls to the LS, which complete and then push their results to whatever next steps are enchained) to a 'pull' - where the UI is calling a method and expecting the result synchronously.

The way this method works is to start a load of calls to the possible servers in the foreground. We're not interested in a result from those CompletableFutures - they have `void` return type and do their work as a mutating side-effect, adding their intermediate results into a field on the class.

Their individual futures are combined into a single one that completes when they all do. If this single future completes before the timeout, then we're good: the whole result is available and can be returned synchronously from this method. If not, then this method will return the result so far, and chain follow-up actions onto the individual uncompleted futures so that they trigger a refresh.

It's a bit convoluted, but I have tried to be faithful to the original behaviour and do a straight port to the new API without attempting to improve or alter the logic. If we used `collectAll` as you suggest then we wouldn't be able to do multiple updates if we got several more results at separate times after the initial timeout. We could make the change you suggest but in that case the user would get the initial result up to that point, but would not see an update until *all* the remaining results had arrived. That might be sufficient but I have confined myself to porting the code and not trying to change the behaviour.

PR #434, opened 2023-02-02
New API: Add single-server document executor

ahmedneilhussain:
This covers the main uses for DocumentInfo: following a server request, bundle the document and the server that responded into an object that can follow up further requests.
Extracted at reviewer's request from an API port into a standalone PR

### DIFF
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServers.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServers.java
@@ -15,6 +15,7 @@ package org.eclipse.lsp4e;
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.List;
 import java.util.Objects;
 import java.util.Optional;
@@ -41,7 +42,11 @@ import org.eclipse.lsp4j.services.LanguageServer;
 
 /**
  * Main entry point for accessors to run requests on the language servers, and some utilities
- * for manipulating the asynchronous response objects in streams
+ * for manipulating the asynchronous response objects in streams.
+ *
+ * Note that versioning support makes the executor classes stateful. Attempting to call request
+ * methods multiple times on the same executor object will throw an <code>IllegalStateExeception</code>.
+ * Either call <code>reset()</code> between intervening executions or <code>clone()</code> a fresh object.
  */
 public abstract class LanguageServers<E extends LanguageServers<E>> {
 
@@ -74,6 +79,7 @@ public abstract class LanguageServers<E extends LanguageServers<E>> {
 	 */
 	@NonNull
 	public <T> CompletableFuture<@NonNull List<@NonNull T>> collectAll(BiFunction<? super LanguageServerWrapper, LanguageServer, ? extends CompletionStage<T>> fn) {
+		checkHasRun();
 		computeVersion();
 		final CompletableFuture<@NonNull List<T>> init = CompletableFuture.completedFuture(new ArrayList<T>());
 		return executeOnServers(fn).reduce(init, LanguageServers::add, LanguageServers::addAll)
@@ -112,6 +118,7 @@ public abstract class LanguageServers<E extends LanguageServers<E>> {
 	 */
 	@NonNull
 	public <T> List<@NonNull CompletableFuture<@Nullable T>> computeAll(BiFunction<? super LanguageServerWrapper, LanguageServer, ? extends CompletionStage<T>> fn) {
+		checkHasRun();
 		computeVersion();
 		return getServers().stream()
 				.map(cf -> cf
@@ -146,6 +153,7 @@ public abstract class LanguageServers<E extends LanguageServers<E>> {
 	 * non-empty response, and with an empty <code>Optional</code> if none of the servers returned a non-empty result.
 	 */
 	public <T> CompletableFuture<Optional<T>> computeFirst(BiFunction<? super LanguageServerWrapper, LanguageServer, ? extends CompletionStage<T>> fn) {
+		checkHasRun();
 		computeVersion();
 		final CompletableFuture<Optional<T>> result = new CompletableFuture<>();
 
@@ -203,6 +211,12 @@ public abstract class LanguageServers<E extends LanguageServers<E>> {
 		// Use CF::getNow with a default null?
 		return getServers().stream().map(CompletableFuture::join).anyMatch(Objects::nonNull);
 	}
+
+	/**
+	 * Creates a new executor from this one, with the same settings and filter, so a further request can be run
+	 */
+	@Override
+	public abstract E clone();
 
 
 	/**
@@ -268,6 +282,72 @@ public abstract class LanguageServers<E extends LanguageServers<E>> {
 			return this.startVersion;
 		}
 
+		public SingleLanguageServerDocumentExecutor toExecutor(final @NonNull LanguageServerWrapper serverWrapper) {
+			return new SingleLanguageServerDocumentExecutor(this.document, serverWrapper);
+		}
+
+		@Override
+		public LanguageServerDocumentExecutor clone() {
+			return new LanguageServerDocumentExecutor(this.document).withFilter(getFilter());
+		}
+
+		@Override
+		public void reset() {
+			super.reset();
+			this.startVersion = 0L;
+		}
+	}
+
+	/**
+	 * Executor for a single server - for follow-up requests that should go to the server that returned an earlier response.
+	 * The multi-LS dispatch methods will work, and will take into account the filter, so will return exactly 1 or 0 response.
+	 * The additional <code>execute()</code> method will bypass the filter and just run the request directly on the single wrapped
+	 * language server.
+	 *
+	 */
+	public static class SingleLanguageServerDocumentExecutor extends LanguageServerDocumentExecutor {
+
+		private final @NonNull LanguageServerWrapper serverWrapper;
+
+		SingleLanguageServerDocumentExecutor(final @NonNull IDocument document, final @NonNull LanguageServerWrapper serverWrapper) {
+			super(document);
+			this.serverWrapper = serverWrapper;
+		}
+
+		@Override
+		@SuppressWarnings("null")
+		protected @NonNull List<@NonNull CompletableFuture<@Nullable LanguageServerWrapper>> getServers() {
+			// Take into account filter
+ 			if (getFilter().test(serverWrapper.getServerCapabilities())) {
+ 				return Collections.singletonList(CompletableFuture.completedFuture(serverWrapper));
+ 			}
+			return Collections.emptyList();
+		}
+
+		/**
+		 * Runs a request directly on the (single) wrapped language server. No filter is applied, but version checking is supported.
+		 * @param <T>
+		 * @param fn
+		 * @return
+		 */
+		public <T> CompletableFuture<T> execute(@NonNull Function<LanguageServer, ? extends CompletionStage<T>> fn) {
+			checkHasRun();
+			computeVersion();
+			return serverWrapper.execute(fn);
+		}
+
+		public boolean supports(final Predicate<ServerCapabilities> predicate) {
+			return predicate.test(serverWrapper.getServerCapabilities());
+		}
+
+		public LanguageServerWrapper getServer() {
+			return serverWrapper;
+ 		}
+
+		@Override
+		public SingleLanguageServerDocumentExecutor clone() {
+			return new SingleLanguageServerDocumentExecutor(getDocument(), serverWrapper);
+		}
 
 	}
 
@@ -309,6 +389,13 @@ public abstract class LanguageServers<E extends LanguageServers<E>> {
 				wrappers.add(wrapper.getInitializedServer().thenApply(ls -> wrapper));
 			}
 			return wrappers;
+		}
+
+		@Override
+		public LanguageServerProjectExecutor clone() {
+			LanguageServerProjectExecutor cloned = new LanguageServerProjectExecutor(this.project).withFilter(getFilter());
+			cloned.restartStopped = this.restartStopped;
+			return cloned;
 		}
 	}
 
@@ -413,7 +500,28 @@ public abstract class LanguageServers<E extends LanguageServers<E>> {
 		return new LanguageServerProjectExecutor(project);
 	}
 
+	/**
+	 *
+	 * @return True if this executor has been used for a request
+	 */
+	public final boolean hasRun() {
+		return hasRun;
+	}
+
+	public void reset() {
+		hasRun = false;
+	}
+
+	protected void checkHasRun() {
+		if (hasRun) {
+			throw new IllegalStateException("Executor has already been used for a request"); //$NON-NLS-1$
+		}
+		hasRun = true;
+	}
+
 	private @NonNull Predicate<ServerCapabilities> filter = s -> true;
+
+	private boolean hasRun = false;
 
 	private boolean isRequestCancelledException(final Throwable throwable) {
 		if (throwable instanceof final CompletionException completionException) {
--- a/org.eclipse.lsp4e.test/src/org/eclipse/lsp4e/test/LanguageServersTest.java
+++ b/org.eclipse.lsp4e.test/src/org/eclipse/lsp4e/test/LanguageServersTest.java
@@ -38,11 +38,13 @@ import org.eclipse.core.resources.IFile;
 import org.eclipse.core.resources.IProject;
 import org.eclipse.core.runtime.CoreException;
 import org.eclipse.jface.text.IDocument;
+import org.eclipse.jface.text.IDocumentExtension4;
 import org.eclipse.jface.text.ITextViewer;
 import org.eclipse.lsp4e.LSPEclipseUtils;
 import org.eclipse.lsp4e.LanguageServerWrapper;
 import org.eclipse.lsp4e.LanguageServers;
 import org.eclipse.lsp4e.LanguageServers.LanguageServerDocumentExecutor;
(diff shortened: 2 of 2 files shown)


ahmedneilhussain:
@rubenporras as requested! I even added a test ;-)

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServers.java
```diff
@@ -268,6 +269,42 @@ public long getStartVersion() {
...
+		public SingleServerDocumentExecutor toExecutor(final @NonNull LanguageServerWrapper serverWrapper) {
+			return new SingleServerDocumentExecutor(this.document, serverWrapper);
+		}
+	}
+
+	/**
+	 * Executor for a single server - for follow-up requests that should go to the server that returned an earlier response
+	 *
+	 */
+	public static class SingleServerDocumentExecutor extends LanguageServerDocumentExecutor {
```
rubenporras, line 281: I think that to be consistent with the names of the other classes, it should be named `SingleLanguageServerDocumentExecutor`, in the name of the other classes we use `LanguageServer` and not just `Server`

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServers.java
```diff
@@ -268,6 +269,42 @@ public long getStartVersion() {
...
+		protected @NonNull List<@NonNull CompletableFuture<@Nullable LanguageServerWrapper>> getServers() {
+			// Compute list of servers from document & filter
+			return Collections.singletonList(CompletableFuture.completedFuture(this.serverWrapper));
+		}
+
+		public <T> CompletableFuture<T> execute(@NonNull Function<LanguageServer, ? extends CompletionStage<T>> fn) {
+			return this.serverWrapper.execute(fn);
+		}
+
+		public boolean supports(final Predicate<ServerCapabilities> predicate) {
```
rubenporras, line 301: I think it not right that this executor does not respect the `withFilter` or `withPredicate` (even though one could call it) and instead offers a new method to do the filter

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServers.java
```diff
@@ -268,6 +269,42 @@ public long getStartVersion() {
...
+
+	/**
+	 * Executor for a single server - for follow-up requests that should go to the server that returned an earlier response
+	 *
+	 */
+	public static class SingleServerDocumentExecutor extends LanguageServerDocumentExecutor {
+
+		private final @NonNull LanguageServerWrapper serverWrapper;
+
+		SingleServerDocumentExecutor(final @NonNull IDocument document, final @NonNull LanguageServerWrapper serverWrapper) {
```
rubenporras, line 285: the LanguageServerDocumentExecutor uses the document to find out the wrappers and connect them if needed, but I do not think this executor uses the document for anything.

rubenporras:
@ahmedneilhussain, I have prepared https://github.com/eclipse/lsp4e/pull/435, it is only the API in LanguageServers extracted out of https://github.com/eclipse/lsp4e/pull/396, it have added a getServer method as you propose and have also taken care of handling  the filter if given.
If you like it I can take over your test and merge. If not we can keep the discussion and see where we converge.

ahmedneilhussain:
HI @rubenporras I think you raise some good points. I proposed this class for various reasons, but it still feels a bit unsatisfactory. Although I was following your model for an executor for a single server/wrapper, this was something I had considered myself anyway as part of the original design for the new API.

The general need for this sort of class is for following up an initial request/response with further requests. There are quite a few existing cases within the old codebase where the language server is stored on e.g. a UI class for subsequent requests. We obviously don't want this, which is why I initially proposed using the wrapper (with a public `execute`) method on it.
Having an executor based around this wrapper takes it one step further but it does have some drawbacks:

- The core executor methods `collectAll` `computeAll` `computeFirst` don't make a lot of sense: they are there to handle the complexities that more than one LS could respond to the request - and the calling code cannot know in advance how many might apply. With an executor for a single LS we already know there will be exactly one LS, so using those three methods for followup requests adds a lot of extra baggage for no gain.
- If we expose a single `execute` method to avoid the pointless complexity of the core executor methods, then what exactly does a single-ls executor offer over using the wrapper itself? The answer might be versioning and cancel support - if and when they get added.
- I like your idea of taking the filter into account. Very often in the codebase we can see that after the initial request, some of the follow up requests depend on the server offering additional capabilities. However that doesn't play very well with having the executor as a final field on the class and being reused - since setting the filter changes it for subsequent requests.

rubenporras:
> HI @rubenporras I think you raise some good points. I proposed this class for various reasons, but it still feels a bit unsatisfactory. Although I was following your model for an executor for a single server/wrapper, this was something I had considered myself anyway as part of the original design for the new API.

Yes, it has its drawbacks, but it has the upside that you can actually tell that the pattern of how to do a remote call to one or many LS, just use `LanguageServers` and the methods in it. Then you can even remove the public method from the LanguageServerWrapper as I did in https://github.com/eclipse/lsp4e/pull/396. That is to me very appealing.

> 
> The general need for this sort of class is for following up an initial request/response with further requests. There are quite a few existing cases within the old codebase where the language server is stored on e.g. a UI class for subsequent requests. We obviously don't want this, which is why I initially proposed using the wrapper (with a public `execute`) method on it. Having an executor based around this wrapper takes it one step further but it does have some drawbacks:
> 
> * The core executor methods `collectAll` `computeAll` `computeFirst` don't make a lot of sense: they are there to handle the complexities that more than one LS could respond to the request - and the calling code cannot know in advance how many might apply. With an executor for a single LS we already know there will be exactly one LS, so using those three methods for followup requests adds a lot of extra baggage for no gain.

Yes, agree with that. I did not look at the overall code, so I do not know how many of these consumers do we actually have. If they are only a few, I think it is fine, they will just not use these APIs. To me it would not be worth having a new interface. It is similar to using `Collections.singletoList, Collections.singletoSet, Collections.singletoMap`. Yes, there is no way of getting one element out of them, you still need to iterate or just call get(0) or similar. It is not perfect, yet it is good enough.

> * If we expose a single `execute` method to avoid the pointless complexity of the core executor methods, then what exactly does a single-ls executor offer over using the wrapper itself? The answer might be versioning and cancel support - if and when they get added.

Yes, I agree with that, that is why in my original version in https://github.com/eclipse/lsp4e/pull/396 I did not have it, to me to use `computeFirst()` or iterate through `collectAll()` as one would do with  `Collections.singletoList`. I added it now to my proposal because I imagine you would like it ;)

> * I like your idea of taking the filter into account. Very often in the codebase we can see that after the initial request, some of the follow up requests depend on the server offering additional capabilities. However that doesn't play very well with having the executor as a final field on the class and being reused - since setting the filter changes it for subsequent requests.

Okay, I see, then we could add to the `forWrapper` method, so that it can only be provided only once, and the implement `withFilter` and `withCapability` to throw an `UnsupportedOperationException` (Like again `SingletonList` does).

I can adapt my PR with any of this changes if you like.

rubenporras:
> I think there are two issues to resolve:
> 
> * Perhaps the single-server executors should not inherit from LanguageServers directly - either they should be a separate (but similar) class hierarchy, or perhaps we should refactor so that there is a common root with e.g. the filtering support:
> 
> ```
>             LanguageServers
>               /                          \
> MultiLSExecutors             SingleLSExecutors
>               |                                                               |
> existing leaf classes                           SingleLanguageServerDocumentExecutor, SingleLanguageServerExecutor
> ```
> 
> What do you think?

I do not know how much code would profit from a better API, but my feeling now is that not many (as I said above), and that we provide more value by porting the code to the API we have (with minor improvements here and there like what we are doing).

rubenporras:
> * I like your idea of taking the filter into account. Very often in the codebase we can see that after the initial request, some of the follow up requests depend on the server offering additional capabilities. However that doesn't play very well with having the executor as a final field on the class and being reused - since setting the filter changes it for subsequent requests.

Actually now that I think about it, is this not a problem with the existing executors ?

I can already write

```
		LanguageServerDocumentExecutor ex = LanguageServers.forDocument(document);
		ex = ex.withFilter(x -> true);
		ex.computeAll(...);
		ex = ex.withFilter(x -> false);
		ex.computeAll(...);
```

and it is undefined what will happen. I think the LanguageServerProjectExecutor will behave as the `LanguageServerWrapperExecutor`, the document one will ignore it. Something we need to improve already aparently :(

ahmedneilhussain:
An interesting point about `Collections.singletonList()`!

Just to let you know I have to work on something else for a couple of days, so will follow this up later in the week.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServers.java
ahmedneilhussain: Done. The names are becoming a bit cumbersome but I guess that's what `var` was invented for ;-)

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServers.java
ahmedneilhussain: See summary of revisions below

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServers.java
ahmedneilhussain: You are (partly) right - one of the motivations for this class was as a replacement for `DocumentInfo`, which effectively bundles a reference to a single LS with a document - it's virtually a POJO with no brains. In some of the code I was trying to port, the existing logic passed through a `DocumentInfo` as a single field. My first attempt replaced this with passing around a `LanguageServerWrapper` and an `IDocument` as individual arguments. But then I thought of taking your `LanguageServerWrapperExecutor` idea and marrying it with `LanguageServerDocumentExecutor`.

I think this gives a nicer way of rewriting the existing code, and also means versioning support is available.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServers.java
```diff
@@ -204,6 +212,12 @@ public boolean anyMatching() {
 		return getServers().stream().map(CompletableFuture::join).anyMatch(Objects::nonNull);
 	}
 
+	/**
+	 * Creates a new executor from this one, with the same settings and filter, so a further request can be run
+	 */
+	@Override
```
rubenporras, line 218: I do not remember the details, but I remember that Effective Java (2nd Edition) recommends avoiding `clone` and use copy constructors instead, and I think it is a usually accepted pattern.
For example
`	public static LanguageServerDocumentExecutor forExecutor(final @NonNull LanguageServerDocumentExecutor executor) {`

and similar for the other executor types.

Could we follow that recommendation?

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServers.java
```diff
@@ -41,7 +42,11 @@
 
 /**
  * Main entry point for accessors to run requests on the language servers, and some utilities
- * for manipulating the asynchronous response objects in streams
+ * for manipulating the asynchronous response objects in streams.
+ *
+ * Note that versioning support makes the executor classes stateful. Attempting to call request
+ * methods multiple times on the same executor object will throw an <code>IllegalStateExeception</code>.
+ * Either call <code>reset()</code> between intervening executions or <code>clone()</code> a fresh object.
```
rubenporras, line 49: I do not think we need the `reset` method, for the number of times we will actually want to reuse executors, I think just creating a new object should be good enough.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServers.java
```diff
@@ -268,6 +282,72 @@ public long getStartVersion() {
...
+		@SuppressWarnings("null")
+		protected @NonNull List<@NonNull CompletableFuture<@Nullable LanguageServerWrapper>> getServers() {
+			// Take into account filter
+ 			if (getFilter().test(serverWrapper.getServerCapabilities())) {
+ 				return Collections.singletonList(CompletableFuture.completedFuture(serverWrapper));
+ 			}
+			return Collections.emptyList();
+		}
+
+		/**
```
rubenporras, line 327: I think we should remove this three methods for the time being and add then later if needed, it is really weird that they promote different patterns, like an executor which does not respect the filter.

I would propose that we go without `execute`, `supports` and `getServer`, change a consumers, and the create a new PR showing how those consumers would be indeed simplified.

Could we do that? Adding the methods is indeed simple if really needed.

rubenporras:
thanks for the newer version of the PR, I have left some comments.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServers.java
ahmedneilhussain: If you want, but that's partly because `clone` and `Clonable` predate generics and are thus broken. I've not actually implemented the 'official' `Clonable` interface, just provided a method called `clone()` that is generic and thus properly covariantly typed.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServers.java
ahmedneilhussain: You are perhaps right but I can foresee getting the reverse complaint as well. Certainly it's more of a convenience thing than a necessity thing if clone/copy is implemented.

rubenporras: I agree, but since adding is easier than removing, I think we should start small. It is safer :)

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServers.java
rubenporras: Alright, I see, I have overlooked that. I still think the copy constructor is preferable, as then there is still only pattern to get new executors (constructor) vs two pattern (close + constructor)

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServers.java
```diff
@@ -268,6 +282,72 @@ public long getStartVersion() {
 			return this.startVersion;
 		}
 
+		public SingleLanguageServerDocumentExecutor toExecutor(final @NonNull LanguageServerWrapper serverWrapper) {
```
rubenporras, line 285: This I would also remove, by the way, and keep the API small. It is only a bit of syntactic sugar, and we should only add those if it really pays. I am not sure yet it does.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServers.java
ahmedneilhussain: The methods were all driven by necessity - the design choices were prompted by the issues I encountered porting code. See for example `CodeActionCompletionProposal.apply()`.

I like the idea that if you want to follow up your initial request with a further action - but only if the server supports the more refined capability - then you can call `clone().withFilter(...).computeFirst()` and get a result only if the server supports it.

In the case of trying to port the apply method above, there are two criteria for proceeding, only one of which is a server predicate, so just using the filter method doesn't work.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServers.java
ahmedneilhussain: Yeah, but it does make the api more convenient, especially as a lot of these methods will get used in lambdas. LSP4e is already somewhat rich in lambda functions that ramble on for multiple lines and are hard to read..!

ahmedneilhussain:
Hi @rubenporras you started re-reviewing before I had a chance to finish typing up my summary of what I had changed and why :-)

Just for the record:

- I have made the single-use aspect of the executors explicit, so the executors throw if used more than once without an intervening called to `reset()`. This was already the case because of versioning support, and cancel support will only make them more stateful, so better to make this aspect of the contract explicit.
- I didn't split the class hierarchy into two (as I initially proposed) with separate executors for a single server. I've kept them with the multi-server `collectAll()` methods and so on, but also kept the single-server `execute()` method. The answer as to what the executor offers over just calling wrapper.execute() would be versioning (and shortly, cancellation) support.

I quite like this as a compromise. I know you've already come back with some comments - can I ask you take a look at the other two PRs I have in flight porting stuff to the new API, as that might give some better rationale for the choices?

ahmedneilhussain:
I don't think statefulness is a big deal if it's made explicit, and if you get something in return. It's more a problem if it's implicit and sporadic and so people introduce bugs by misusing the class because they're not aware of that aspect of the contract. Lots of classes like eclipse `Job` implementations and e.g. the `LSSearchQuery` classes we have in this codebase are essentially 'one-shot' classes that should only be run once: you don't rerun them, you create a new object.

Versioning and cancel support will IMHO make the trade off well worth it, especially as it's trivial to implement `clone` and `reset` methods (or copy constructors!) that make the statefulness no inconvenience. Cancel support could be done outside the executors directly (but less conveniently), but versioning makes sense as built in. I think it is worth having a rich API that makes common stuff easy and concise, then that means consumers are less likely to make mistakes and can also focus more on their domain-level logic. I think the fact that most of the time the code is shrinking when we port it to the new API is a good sign.

rubenporras:
Hi @ahmedneilhussain , I have the felling we have done quite well without an executor like this one or the one I proposed in another PR. I already closed my PR, shall we close this one as well?

ahmedneilhussain:
> Hi @ahmedneilhussain , I have the felling we have done quite well without an executor like this one or the one I proposed in another PR. I already closed my PR, shall we close this one as well?

Sure @rubenporras I think we can bin this and reinstate one or other idea if a real need arises in future.

PR #440, opened 2023-02-08
New API: Port HighlightReconcilingStrategy

rubenporras:


rubenporras:
@ahmedneilhussain, does it look good to you?

ahmedneilhussain:
The change looks great to me.

ahmedneilhussain:
Unrelated to this change, is document highlight working correctly at the moment? I wasn't aware it was even implemented as an LSP feature until I dug into this code recently, because I've never seen it working. AIUI, this is the feature that should show e.g. other instances of a variable in scope when you click on one of them (as it does in Java). If I try this using corrosion then I can see the requests firing and the response coming back but never see any highlights.

rubenporras:
It works for me in the latest released LSP4E, but we use our own server, I cannot tell with corrosion.

PR #441, opened 2023-02-08
New API: port FoldingReconcilingStrategy

rubenporras:
@ahmedneilhussain , does it makes sense?


ahmedneilhussain:
I might go with `computeAll` rather than `collectAll` when the results are all being used independently. That might be a bit pedantic!

For extra pedantry
```
LanguageServers.forDocument(document).withCapability(ServerCapabilities::getFoldingRangeProvider).computeAll(
 				server -> server.getTextDocumentService().foldingRange(params))
                                .forEach(result -> result.thenAccept(this::applyFolding));
```
(the async is already baked into the futures returned from `computeAll`).

It's not really a big deal - and your code is more succinct anyway. But assuming we go ahead and merge in cancellation support then it's going to be necessary to avoid tacking follow-up computations directly onto the CF returned from the LSP4j layer.

rubenporras:
Thanks, I have applied your suggestion

rubenporras:
flaky test.

PR #442, opened 2023-02-08
New API: Port find references

rubenporras:
@ahmedneilhussain : what about this one?

### DIFF
--- a/org.eclipse.lsp4e.test/src/org/eclipse/lsp4e/test/references/FindReferencesTest.java
+++ b/org.eclipse.lsp4e.test/src/org/eclipse/lsp4e/test/references/FindReferencesTest.java
@@ -97,21 +97,22 @@ public class FindReferencesTest {
 		long delay = System.currentTimeMillis() - time;
 		// TODO re-use the UI freeze monitoring org.eclipse.ui.monitoring instead
 		assertTrue("Find references blocked UI for " + delay + "ms", delay < uiFreezeThreesholdreezeThreeshold);
-		AtomicInteger runCount = new AtomicInteger();
+		AtomicInteger blockedCount = new AtomicInteger();
 		Thread uiThreadActiveChecker = new Thread(() -> {
 			while (!Thread.currentThread().isInterrupted()) {
 				long triggerTime = System.currentTimeMillis();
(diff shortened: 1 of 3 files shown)


ahmedneilhussain:
@rubenporras  do you know what's causing the test failures?

org.eclipse.lsp4e.test/src/org/eclipse/lsp4e/test/references/FindReferencesTest.java
```diff
@@ -111,7 +108,6 @@ public void findReferencesNonBlocking() throws Exception {
 		uiThreadActiveChecker.start();
 		try {
 			assertNotNull("Search Result view not found", findSearchResultView(5000));
-			assertTrue("UI Thread was frozen", runCount.intValue() > 1000);
```
rubenporras, line 114: I have checked the new code and I think the reason why I need to set this to a lower value is that the new code blocks the UI less than the previous one, and findSearchResultView returns faster than before, thus the uiThreadActiveChecker does not have time to increase the time high enough.

I have put some breakpoints to make sure of that, and I am quite sure that creating the lambda in the body of execturor.collectAll in LSSearchQuery.run is not being run in the UI Thread. Also we run the query in the background with NewSearchUI.runQueryInBackground(query).

Do you agree that the test is not actually measuring frozen UI, but rather the number of iterations in the while loop in the uiThreadActiveChecker, and that being small is not the same as a frozen UI?

rubenporras:
> @rubenporras do you know what's causing the test failures?

Yes, I think I do: see https://github.com/eclipse/lsp4e/pull/442/files#r1101661929

rubenporras:
@mickaelistria , the failing test trigger `CastClassException `in `org.eclipse.search2.internal.ui.SearchViewManager.findLRUSearchResultView(IWorkbenchPage, boolean)`, I have looked at the code and I think it is a bug that we should fix, but I have trouble finding this code in Github. Could you point me to the project in _https://github.com/orgs/eclipse-platform/repositories_ that have the code? Then I would create a PR with a fix.
Thanks

rubenporras:
Finally I could figure out the problem with the test :)

PR #450, opened 2023-02-09
Fix NPE if a location is null that happens during execution of the tests

rubenporras:
Fix NPE if a location is null that happens during execution of the LSP4E tests


mickaelistria:
Thanks.

PR #454, opened 2023-02-09
Do not connect documents in anyMatching

rubenporras:
`anyMatching` will connect the document if not already connected, while
`LanguageServiceAccessor.getLanguageServers(..., capabilities -> ...)
would not do that.

So if the document is not connected, this call will take longer and will
have a side effect.

I think anyMatching should not connect documents, they should be
connected on demand when any of the collect/compute methods are called.

### DIFF
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServers.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServers.java
@@ -209,6 +209,7 @@ public abstract class LanguageServers<E extends LanguageServers<E>> {
 	 * Executor that will run requests on the set of language servers appropriate for the supplied document
 	 *
 	 */
+	@SuppressWarnings("null")
 	public static class LanguageServerDocumentExecutor extends LanguageServers<LanguageServerDocumentExecutor> {
 
 		private final @NonNull IDocument document;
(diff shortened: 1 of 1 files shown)


rubenporras:
@ahmedneilhussain, fyi.

ahmedneilhussain:
I think splitting out filtering and connection and making `anyMatching` avoid connecting the document is a great idea. Just a small - but critical! - mistake in the way you have refactored the calling of `LanguageServerWrapper.connect()`

ahmedneilhussain:
PS I think you'll need to override the timeout-overload of `anyMatching` that you have submitted independently?

rubenporras:
> PS I think you'll need to override the timeout-overload of `anyMatching` that you have submitted independently?

Yes, if we merge this, I will update the other PR, or if we merge the other one first, then I update this one ;)

rubenporras:
> I think splitting out filtering and connection and making `anyMatching` avoid connecting the document is a great idea. Just a small - but critical! - mistake in the way you have refactored the calling of `LanguageServerWrapper.connect()`

which one? I am not sure I know...

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServers.java
```diff
@@ -227,34 +228,48 @@ public static class LanguageServerDocumentExecutor extends LanguageServers<Langu
...
-					try {
-						return wrapper.connect(document);
-					} catch (IOException ex) {
-						LanguageServerPlugin.logError(ex);
-					}
-				}
-				return CompletableFuture.completedFuture(null);
-			}).thenApply(server -> server == null ? null : wrapper);
+		private @NonNull CompletableFuture<@Nullable LanguageServerWrapper> connectIf(@NonNull CompletableFuture<@Nullable LanguageServerWrapper> wrapper) {
+			return wrapper.thenCombine(wrapper.thenApply(this::connect), (w, server) -> w);
```
rubenporras, line 245: this thenCombine looks cumbersome to me, but I did not find anything better. @ahmedneilhussain , is there anything simpler to achieve this? I am quite struggling with the CompletableFuture API, so probably there is something simple which I have missed

ahmedneilhussain:
Hi @rubenporras I think there isn't any need for `thenCombine()` - it should just be a small tweak to the original code to split the filter on the server capabilities out from the call to `wrapper.connect()`.

How about something like this:

```
		@NonNull CompletableFuture<@Nullable LanguageServerWrapper> filter(@NonNull LanguageServerWrapper wrapper) {
			return wrapper.getInitializedServer().thenApply(server -> {
				if (server != null && getFilter().test(wrapper.getServerCapabilities())) {
					return wrapper;
				}
				return null;
			});
		}

		@NonNull CompletableFuture<@Nullable LanguageServerWrapper> connect(@NonNull CompletableFuture<@Nullable LanguageServerWrapper> wrapperFuture) {
			return wrapperFuture.thenCompose(wrapper -> {
				try {
					if (wrapper != null) {
						CompletableFuture<LanguageServer> tmp = wrapper.connect(document);
						if (tmp != null) {
							return tmp.thenApply(ls -> wrapper);
						}
					}
				} catch (IOException ex) {
					LanguageServerPlugin.logError(ex);
				}
			});
		}

		@Override
		protected @NonNull List<@NonNull CompletableFuture<@Nullable LanguageServerWrapper>> getServers() {
			// Compute list of servers from document & filter
			return LanguageServiceAccessor.getLSWrappers(this.document).stream()
				.map(this::filter)
				.map(this::connect)
				.toList();
		}
```

ahmedneilhussain:
In fairness, I thought this was a trivial change but it took me a few minutes, because there were some issues with null-ness to worry about. The original code may have had a bug because `wrapper.connect(IDocument)` can return `null`, as distinct from `CompletableFuture.completedFuture(null)`!! For some reason the `@Nullable` annotation processor did not spot this in the original code but (correctly) complained about it in the modified version above until I figured out what was going on.

ahmedneilhussain:
The CF api does take some getting used to... I found https://www.baeldung.com/java-completablefuture quite helpful.

The main thing to get comfortable with is that
`cf.thenApply(x -> y)` is identical to `cf.thenCompose(x -> CompletableFuture.completedFuture(y))`

The way we use them in `LanguageServiceAccessor` (and here) is a little odd, in some ways - normally you would have a construct like  `CompletableFuture<LanguageServerWrapper>` because the wrapper was not yet available as an object; in our case, we already have access to the underlying wrapper object, we are using the completable future to access the object *when it is guaranteed to be in some particular state* (e.g. language server started; document connected).

rubenporras:
@ahmedneilhussain. Thanks, I have applied your suggestion and I will read that page carefully.

rubenporras:
I think this is now good, so I will merge it, if you think there is something I have not consider, I can adapt the code based on your comments.

ahmedneilhussain:
Thanks - one final comment to fix.

On this topic, do you think maybe some of these low-level methods could benefit from some comments making clear what is going on and why? I think between us we have demonstrated that it is not at all obvious! We have worked very hard to make the API provide robust ordering guarantees but it was quite tricky to get right and thus the resulting code is a bit brittle: it will be very easy for a future maintainer to make reasonable-looking changes that introduce subtle bugs.

rubenporras:
> do you think maybe some of these low-level methods could benefit from some comments making clear what is going on and why

Yes, that would be very nice. I for myself still do not understand for example why
```
			// Ensure any subsequent computation added by caller does not block further incoming messages from language servers
			.thenApplyAsync(Function.identity());

```
Even when it is already commented :(

PR #455, opened 2023-02-10
New API: Port documentLink

rubenporras:


rubenporras:
@ahmedneilhussain , looks good to you?

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/documentLink/DocumentLinkDetector.java
```diff
@@ -80,35 +80,18 @@ public IHyperlink[] detectHyperlinks(ITextViewer textViewer, IRegion region, boo
...
-										return null;
-									} catch (TimeoutException e) {
-										LanguageServerPlugin.logWarning("Could not detect hyperlinks due to timeout after 2 seconds in `document/Link`", e); //$NON-NLS-1$
-										return null;
-									}
-								}).filter(Objects::nonNull).flatMap(List<DocumentLink>::stream).map(link -> {
+			return LanguageServers.forDocument(document)
+					.withFilter(capabilities -> capabilities.getDocumentLinkProvider() != null)
+					.collectAll(languageServer -> languageServer.getTextDocumentService().documentLink(params))
+					.thenApplyAsync(links -> {
```
ahmedneilhussain, line 86: `async` not needed - already done by `collectAll`.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/documentLink/DocumentLinkDetector.java
```diff
@@ -80,35 +80,18 @@ public IHyperlink[] detectHyperlinks(ITextViewer textViewer, IRegion region, boo
...
+						IHyperlink[] res = links.stream().flatMap(List<DocumentLink>::stream).filter(Objects::nonNull)
+								.filter(link -> link.getTarget() != null).map(link -> {
```
ahmedneilhussain, line 88: Not your doing, but this is a bit of a monster lambda expression.

Could you pull some of it out into a method while you're in the code? Maybe a (lsp4j) DocumentLink -> (eclipse) DocumentHyperLink converter?

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/documentLink/LSPDocumentLinkPresentationReconcilingStrategy.java
```diff
@@ -70,18 +70,17 @@ private void underline() {
...
-							if (control != null) {
-								control.getDisplay().asyncExec(() -> underline(links));
-							}
-						})).toArray(CompletableFuture[]::new)));
+		cancel();
+		final var params = new DocumentLinkParams(LSPEclipseUtils.toTextDocumentIdentifier(uri));
+		request = LanguageServers.forDocument(document)
+				.withFilter(capabilities -> capabilities.getDocumentLinkProvider() != null)
+				.collectAll(languageServer -> languageServer.getTextDocumentService().documentLink(params))
+				.thenAcceptAsync(links -> {
```
ahmedneilhussain, line 78: Again, async not needed .... unless using the display as an executor

ahmedneilhussain: .. which you can't do if control is null. Although whether that is something that should be guarded against is debatable. Could extract the display at the top of this method and then use it as the async executor.

ahmedneilhussain:
Couple of points, but LGTM broadly.

rubenporras:
@ahmedneilhussain , good now? 

Yes, I knew about the big method, but I wanted to keep my PR as small as possible. Now it is part of a second commit inside the PR.

ahmedneilhussain:
> @ahmedneilhussain , good now?
> 
> Yes, I knew about the big method, but I wanted to keep my PR as small as possible. Now it is part of a second commit inside the PR.

Yeah, thanks. I think you're right to separate out the commits and keep the API changes distinct from unrelated (but welcome) code improvements.

rubenporras:
I will merge this as well, if comments come, I will address them as well

sebthom:
@rubenporras did you on purpose increase the timeout from 2 to 4 seconds in the detectHyperlinks method? the problem is that this runs in the UI thread and freezes the GUI. and with slow language servers it freezes a lot.

rubenporras:
I do not know for sure, but most probably intentional. We have some cases where our LS is not so fast, and we prefer it to be slower than not to present any information.
I think we should make this timeouts configurable.

PR #456, opened 2023-02-10
New API: Port link edit

rubenporras:
@ahmedneilhussain , ok for you?

### DIFF
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/linkedediting/LSPLinkedEditingBase.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/linkedediting/LSPLinkedEditingBase.java
@@ -12,20 +12,24 @@
 package org.eclipse.lsp4e.operations.linkedediting;
 
 import java.net.URI;
+import java.util.Objects;
+import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
-import java.util.concurrent.atomic.AtomicReference;
 
 import org.eclipse.core.runtime.preferences.IEclipsePreferences;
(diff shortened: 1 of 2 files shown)


org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/linkedediting/LSPLinkedEditingBase.java
```diff
@@ -66,18 +67,10 @@ protected CompletableFuture<LinkedEditingRanges> collectLinkedEditingRanges(IDoc
...
-				.thenComposeAsync(languageServers ->
-					CompletableFuture.allOf(languageServers.stream()
-							.map(ls -> ls.getTextDocumentService().linkedEditingRange(LSPEclipseUtils.toLinkedEditingRangeParams(params))
-								.thenAcceptAsync(result -> {
-									if (result != null && result.getRanges() != null && result.getRanges().size() > 1) {
-										range.set(result);
-									}
-							})).toArray(CompletableFuture[]::new)))
-				.thenApplyAsync(theVoid -> range.get());
+		return LanguageServers.forDocument(document).withCapability(ServerCapabilities::getLinkedEditingRangeProvider)
```
mickaelistria, line 70: The issue is that there is no filtering on the result: sometimes, some LS may return nothing or a single range and we need to use the next LS. This seems lost in translation. Maybe instead we should `computeAll` and filter the output to choose the most interesting one?

rubenporras: Good catch, My code looked too simple :) I will fix that.

PR #459, opened 2023-02-13
New api: port rename

ahmedneilhussain:
Ported rename to the new API. We need to make a slightly arbitrary choice of LS in the (fairly unlikely, in practice) case where more than one LS could do the rename. The old logic just picked the first LS that supported rename; I've changed this  slightly to pick the first LS that responds to the prepareRename phase and stick with that one.

I ran into some logic that had not been updated to account for the fact that `WorkspaceEdit`s can now consist of `DocumentEdit`s rather than `Edit`s, so have fixed that.

### DIFF
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/rename/LSPRenameHandler.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/rename/LSPRenameHandler.java
@@ -29,6 +29,8 @@ import org.eclipse.jface.viewers.ISelection;
 import org.eclipse.jface.viewers.ISelectionProvider;
 import org.eclipse.lsp4e.LSPEclipseUtils;
 import org.eclipse.lsp4e.LanguageServerPlugin;
+import org.eclipse.lsp4e.LanguageServers;
+import org.eclipse.lsp4e.LanguageServers.LanguageServerDocumentExecutor;
 import org.eclipse.lsp4e.LanguageServiceAccessor;
 import org.eclipse.lsp4e.ui.Messages;
 import org.eclipse.lsp4e.ui.UI;
@@ -55,27 +57,23 @@ public class LSPRenameHandler extends AbstractHandler implements IHandler {
 				IDocument document = LSPEclipseUtils.getDocument(textEditor);
 				if (document != null) {
 					Shell shell = part.getSite().getShell();
-					return LanguageServiceAccessor.getLanguageServers(document, LSPRenameHandler::isRenameProvider)
-							.thenAcceptAsync(languageServers -> {
-								if (languageServers.isEmpty()) {
-									return;
-								}
-								int offset = textSelection.getOffset();
-								// TODO consider better strategy to pick LS, or iterate over LS until one gives
-								// 	a good result
-								final var processor = new LSPRenameProcessor(document, languageServers.get(0), offset);
-								final var refactoring = new ProcessorBasedRefactoring(processor);
-								final var wizard = new LSPRenameRefactoringWizard(refactoring);
-								final var operation = new RefactoringWizardOpenOperation(wizard);
-								shell.getDisplay().asyncExec(() -> {
-									try {
-										operation.run(shell, Messages.rename_title);
-									} catch (InterruptedException e1) {
-										LanguageServerPlugin.logError(e1);
-										Thread.currentThread().interrupt();
-									}
-								});
-							});
+					LanguageServerDocumentExecutor executor = LanguageServers.forDocument(document).withFilter(LSPRenameHandler::isRenameProvider);
+					if (executor.anyMatching()) {
+						int offset = textSelection.getOffset();
+
+						final var processor = new LSPRenameProcessor(document, executor, offset);
+						final var refactoring = new ProcessorBasedRefactoring(processor);
+						final var wizard = new LSPRenameRefactoringWizard(refactoring);
+						final var operation = new RefactoringWizardOpenOperation(wizard);
+						shell.getDisplay().asyncExec(() -> {
+							try {
+								operation.run(shell, Messages.rename_title);
+							} catch (InterruptedException e1) {
+								LanguageServerPlugin.logError(e1);
+								Thread.currentThread().interrupt();
+							}
+						});
+					}
 				}
 			}
 		}
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/rename/LSPRenameHandler.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/rename/LSPRenameHandler.java
@@ -61,7 +61,7 @@ public class LSPRenameHandler extends AbstractHandler implements IHandler {
 					if (executor.anyMatching()) {
 						int offset = textSelection.getOffset();
 
-						final var processor = new LSPRenameProcessor(document, executor, offset);
+						final var processor = new LSPRenameProcessor(document, offset);
 						final var refactoring = new ProcessorBasedRefactoring(processor);
 						final var wizard = new LSPRenameRefactoringWizard(refactoring);
 						final var operation = new RefactoringWizardOpenOperation(wizard);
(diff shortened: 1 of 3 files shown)


rubenporras:
> I ran into some logic that had not been updated to account for the fact that `WorkspaceEdit`s can now consist of `DocumentEdit`s rather than `Edit`s, so have fixed that.

Could we have that in a different PR that we can review independently?

ahmedneilhussain:
@rubenporras I've pulled the bug fix out into #460 so this just has the API port commit.

ahmedneilhussain:
I wondered whether I should add versioning to this port. Versioning isn't quite as necessary for rename as e.g. format as it is invoked via a modal dialog. What do you think, @rubenporras  (and others)?

mickaelistria:
> I wondered whether I should add versioning to this port

Can you please elaborate about what version you'd like to change and why? I think this is internal stuff, so I don't think a version should be bumped more than usual.

ahmedneilhussain:
> > I wondered whether I should add versioning to this port
> 
> Can you please elaborate about what version you'd like to change and why? I think this is internal stuff, so I don't think a version should be bumped more than usual.

No, no, not the version of LSP4e! I meant optimistic locking in the same was as formatting now uses `VersionedEdits` - so if the document changes between the request and applying the resulting `WorkspaceEdit` then the rename is aborted. At the moment no such check is made.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/rename/LSPRenameProcessor.java
```diff
@@ -102,19 +105,22 @@ public RefactoringStatus checkInitialConditions(IProgressMonitor pm)
...
-						status.addFatalError(Messages.rename_invalidated);
-					}
-				}
+			final var identifier = LSPEclipseUtils.toTextDocumentIdentifier(document);
+			final var params = new PrepareRenameParams();
+			params.setTextDocument(identifier);
+			params.setPosition(LSPEclipseUtils.toPosition(offset, document));
+
+
+			Optional<Pair<LanguageServerWrapper, Either3<Range, PrepareRenameResult, PrepareRenameDefaultBehavior>>> tmp = initialExecutor.withFilter(LSPRenameProcessor::isPrepareRenameProvider)
```
rubenporras, line 114: I understand that this works, but I do not quite like to change executors like this. Do you think is it a good pattern to promote? I would rather use LanguageServers.forDocument(document).withFilter(...) and do not pass the executor to the method at all. We already pass the document and reusing the executor does not bring anything here. What do you think?

rubenporras (commented):
Hi @ahmedneilhussain , just a small comment, which I do not know if you agree with or not. I would be happy if we do not reset executors like this (I was actually thinking about creating a PR to forbid it) but if you want to keep it just tell me and I will approve & merge the change. It looks good to me otherwise.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/rename/LSPRenameProcessor.java
ahmedneilhussain: You're right. Passing an executor and then changing the filter amounts to no more than passing the document in a needlessly elaborate way...

rubenporras:
@ahmedneilhussain , I would merge it now but there are conflicts on the `LSPRenameProcessor`, could you resolve them?

ahmedneilhussain:
> @ahmedneilhussain , I would merge it now but there are conflicts on the `LSPRenameProcessor`, could you resolve them?

Yes, I'm working on that ATM. I will also change the `isEnabled` implementation to use the new API now you have added a timeout to `anyMatch`.

ahmedneilhussain:
@rubenporras how does this look?

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/rename/LSPRenameHandler.java
```diff
@@ -97,19 +95,21 @@ public void setEnabled(Object evaluationContext) {
 			IDocument document = LSPEclipseUtils.getDocument((ITextEditor) part);
 			if (document != null && isEnable) {
 				try {
-					isEnable = !LanguageServiceAccessor.getLanguageServers(document, LSPRenameHandler::isRenameProvider)
-							.get(50, TimeUnit.MILLISECONDS).isEmpty();
+					// See if there is a server that supports rename. We don't use LanguageServers.anyMatching in this
+					// case because we want to time out and not block the UI if we have to wait for servers to start
+					// but we also want to be able to detect the timeout and schedule a handler update for when
+					// the definitive info on server capabilities is available
+					isEnable = LanguageServers.forDocument(document).withFilter(LSPRenameHandler::isRenameProvider)
```
rubenporras, line 102: Why do we rewrite this code? Does it have an advantage over the previous one? If not, and I would not know of any, I would leave the code as it was.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/rename/LSPRenameHandler.java
```diff
@@ -97,19 +95,21 @@ public void setEnabled(Object evaluationContext) {
...
-								final var handleEvent = new HandlerEvent(this, enabled, false);
-								fireHandlerChanged(handleEvent);
-							});
-
+					// Timed out, schedule a handler update from when all servers are started and have provided their capabilities
+					LanguageServers.forDocument(document).withFilter(LSPRenameHandler::isRenameProvider)
+						.computeFirst(ls -> CompletableFuture.completedFuture(true))
+						.thenAccept(o -> o.ifPresent(b -> {
+							final var handleEvent = new HandlerEvent(this, true, false);
+							fireHandlerChanged(handleEvent);
```
rubenporras, line 111: Same here: Why do we rewrite this code? Does it have an advantage over the previous one? If not, and I would not know of any, I would leave the code as it was.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/rename/LSPRenameHandler.java
```diff
@@ -97,19 +95,21 @@ public void setEnabled(Object evaluationContext) {
...
-					// in case the language servers take longer to kick in, defer the enablement to
-					// a later time
-					LanguageServiceAccessor.getLanguageServers(document, LSPRenameHandler::isRenameProvider)
-							.thenAccept(languageServer -> {
-								boolean enabled = !languageServer.isEmpty();
-								final var handleEvent = new HandlerEvent(this, enabled, false);
-								fireHandlerChanged(handleEvent);
-							});
-
+					// Timed out, schedule a handler update from when all servers are started and have provided their capabilities
```
rubenporras, line 106: No your mistake, but should we not handle only `TimeoutException` with a retry, and not the `ExecutionException`?

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/rename/LSPRenameHandler.java
ahmedneilhussain: I was going to use your anyMatching(timeout, unit) overload, but then I saw you ended up not submitting that.

I know we are rewriting things in the new API so that they get the benefit of the concurrency guarantees (which are not needed here), but the medium-term goal is that LanguageServiceAccessor becomes internal only and that the many confusing overloads get deprecated and removed or made package-private.

Since we want to deprecate direct access to these low-level methods, and given I was in the class anyway, I thought I might as well port this method so that we have this area of the code now completely decoupled from the old API.

rubenporras: Okay, I have looked at the code a bit more, and I understand the use case, it is a good idea to offer a consistent way to deal with that, but it is a separate issue that the request ordering. The pattern of "look for matches and if timeout enable on the background" seems useful but would apply to all the handlers (for example org.eclipse.lsp4e.operations.format.LSPFormatHandler.isEnabled()). There are more in the code base.
I would be for taking it out of this PR and creating a new PR for that, so that we can merge the rest or if we do it here, have something that we can apply on the other handlers as well, in the second case, what about adding a Consumer to `anyMatching` that will be called if timeout before returning `false`? In my opinion that would be the pattern to follow, but just to do one bit at a time, I would still put that in a different PR

ahmedneilhussain: OK if you want I will back this change out

rubenporras (approved):
Thanks

rubenporras:
@ahmedneilhussain , should I do the new method for anyMatching or would you like to do it?

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/rename/LSPRenameHandler.java
ahmedneilhussain: I think you're right - if there's an execution exception thrown here then it'll be thrown on retry (probably) anyway. Pulled that out into a pure error clause.

PR #460, opened 2023-02-13
Rename: Fix issue with rename wizard

ahmedneilhussain:
The LS protocol now allows (and indeed prefers) a WorkspaceEdit to consist of DocumentChanges instead of Changes. Some old code was assuming that if there were no Changes then there was nothing to do. This mean a load of spurious 'no edits' warnings in the UI before the rename succeeded anyway.

Split out of #459 as not strictly related to the API port

@rubenporras FYI


rubenporras:
thanks

PR #462, opened 2023-02-14
If prepare rename timeouts, the rename operation can still be done,

rubenporras:
"newName" will be used as placeholder. Adding a fatal error is not necessary.

### DIFF
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/rename/LSPRenameProcessor.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/rename/LSPRenameProcessor.java
@@ -16,6 +16,7 @@ package org.eclipse.lsp4e.operations.rename;
 import java.util.List;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
 
 import org.eclipse.core.runtime.Assert;
 import org.eclipse.core.runtime.CoreException;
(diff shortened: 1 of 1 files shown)


org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/rename/LSPRenameProcessor.java
```diff
@@ -110,7 +111,11 @@ public RefactoringStatus checkInitialConditions(IProgressMonitor pm)
 					final var params = new PrepareRenameParams();
 					params.setTextDocument(identifier);
 					params.setPosition(LSPEclipseUtils.toPosition(offset, document));
-					prepareRenameResult = languageServer.getTextDocumentService().prepareRename(params).get(1000, TimeUnit.MILLISECONDS);
+					try {
+						prepareRenameResult = languageServer.getTextDocumentService().prepareRename(params).get(1000, TimeUnit.MILLISECONDS);
+					} catch (TimeoutException e) {
+						LanguageServerPlugin.logInfo("Could not prepare rename due to timeout after 1 seconds in `textDocument/prepareRename`. 'newName' will be used"); //$NON-NLS-1$
```
mickaelistria, line 117: This should be a warning as it's abnormal behavior.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/rename/LSPRenameProcessor.java
```diff
@@ -110,7 +111,11 @@ public RefactoringStatus checkInitialConditions(IProgressMonitor pm)
...
 					params.setTextDocument(identifier);
 					params.setPosition(LSPEclipseUtils.toPosition(offset, document));
-					prepareRenameResult = languageServer.getTextDocumentService().prepareRename(params).get(1000, TimeUnit.MILLISECONDS);
+					try {
+						prepareRenameResult = languageServer.getTextDocumentService().prepareRename(params).get(1000, TimeUnit.MILLISECONDS);
+					} catch (TimeoutException e) {
+						LanguageServerPlugin.logInfo("Could not prepare rename due to timeout after 1 seconds in `textDocument/prepareRename`. 'newName' will be used"); //$NON-NLS-1$
+					}
 					if (prepareRenameResult == null) {
 						status.addFatalError(Messages.rename_invalidated);
```
mickaelistria, line 120: Isn't it still blocking the operation in such case?

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/rename/LSPRenameProcessor.java
rubenporras: Okay, I have changed it. It is only that this warnings with an exception attached look a bit scary on the console, but that is another topic.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/rename/LSPRenameProcessor.java
rubenporras: done

PR #463, opened 2023-02-14
Implement anyMatching as a non blocking call using a default timeout

rubenporras:
If the default timeout is not met, we return that a server is matching. Usually the pattern is to ask anyMatching before enabling a functionality on a given document, but if the functionality is enabled and there will always be follow-up request to use the functionality.

Usually the call is done to show/hide some menu item on the UI. It is better to return quick and let the action do nothing if the user actually runs the action than to block for a (potentially) long time the creation on the menu items.

### DIFF
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServers.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServers.java
@@ -21,6 +21,8 @@ import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionException;
 import java.util.concurrent.CompletionStage;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
 import java.util.function.BiFunction;
 import java.util.function.Function;
 import java.util.function.Predicate;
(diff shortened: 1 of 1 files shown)


rubenporras:
@ahmedneilhussain , does it makes sense to you? I think this is better than having anyMatches without timeout hanging potentially forever and another one with a timeout.
Alternatively, we could return Optional<Boolean> and let the optional be empty if there are timeouts + no matching, but I think returning true is a sane default that would fix all the consumers we adapted so far

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServers.java
```diff
@@ -267,7 +284,7 @@ public static class LanguageServerDocumentExecutor extends LanguageServers<Langu
 		@Override
 		public boolean anyMatching() {
 			return LanguageServiceAccessor.getLSWrappers(document).stream()
-					.map(this::filter).map(CompletableFuture::join).anyMatch(Objects::nonNull);
+					.map(this::filter).map(this::matches).anyMatch(bool -> bool);
```
mickaelistria, line 287:
```suggestion
			return !LanguageServiceAccessor.getLSWrappers(document).stream()
					.map(this::filter).map(this::matches).isEmpty();
```

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServers.java
```diff
@@ -194,14 +196,29 @@ public <T> CompletableFuture<Optional<T>> computeFirst(BiFunction<? super Langua
...
+	 *  the server would be matching, and we rely on the next call of to executor to filter the server if needed.
 	 *
-	 * @return True if there is a language server for this project/document & server capabilities
+	 * @return True if there is a language server for this project/document & server capabilities.
 	 */
 	public boolean anyMatching() {
-		// TODO: should maybe have a default timeout for this...?
-		// Use CF::getNow with a default null?
-		return getServers().stream().map(CompletableFuture::join).anyMatch(Objects::nonNull);
+		return getServers().stream().map(this::matches).anyMatch(bool -> bool);
```
mickaelistria, line 221:
```suggestion
		return !getServers().stream().map(this::matches).isEmpty();
```

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServers.java
rubenporras: thanks, this kind of test makes more sense. There is no isEmpty on an stream but I have used now `.count() != 0`

mickaelistria: Sorry, it's just `.empty()`. Please avoid using `.count()` as it will evaluate all content of the stream, the complexity is `O(n)` instead of `O(1)`

rubenporras: My IDE tells me there is no `empty()` either. Am I doing something wrong? 

So far I have used .findFirst().isPresent() in https://github.com/eclipse/lsp4e/pull/467

mickaelistria: My bad again, `empty()` is an empty constructor; .findFirst().isPresent() is good.

PR #466, opened 2023-02-14
New API: Port Outline

rubenporras:


rubenporras:
@ahmedneilhussain , does it makes sense to you?

mickaelistria (approved):
This looks good to me; and I believe it showcases well some benefits of newer API.

ahmedneilhussain:
> @ahmedneilhussain , does it makes sense to you?

@rubenporras there's a potential problem with calling `get()` on an Optional, but otherwise LGTM.

PR #468, opened 2023-02-15
Avoid get() on optional

rubenporras:


rubenporras:
thanks @ahmedneilhussain for finding the problem

PR #469, opened 2023-02-16
New API: Port Code Action Marker Resolution

rubenporras:


rubenporras:
@ahmedneilhussain , does it makes sense?

rubenporras:
I need to take a look at the failing tests

rubenporras:
As usual, just let me know if something is not right.

PR #470, opened 2023-02-16
New API: covariantly return null on failure path

ahmedneilhussain:
@rubenporras this was a comment I made on #454 that I think arrived too late for you to notice.

If connect fails or returns null, then we should return null also in this method, not wrapper.

It's actually quite hard to provoke this error path when wrapper is not itself null, but for paranoia's sake I think we should still correct the logic here.

### DIFF
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServers.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServers.java
@@ -257,7 +257,7 @@ public abstract class LanguageServers<E extends LanguageServers<E>> {
 						LanguageServerPlugin.logError(e);
 					}
 				}
-				return CompletableFuture.completedFuture(wrapper);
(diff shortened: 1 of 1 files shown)


org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServers.java
```diff
@@ -257,7 +257,7 @@ public static class LanguageServerDocumentExecutor extends LanguageServers<Langu
 						LanguageServerPlugin.logError(e);
 					}
 				}
-				return CompletableFuture.completedFuture(wrapper);
+				return CompletableFuture.completedFuture(null);
```
rubenporras, line 260: Hi @ahmedneilhussain , but t this point `wrapper` will be `null` because we checked on line 250, right?

rubenporras:
thanks, this covers the exception case which I did not think about.

Issue #392, opened 2023-01-23
Resource exhaustion with many files with annotations


ahmedneilhussain:
We've just experienced a problem (or more likely managed to reproduce reliably a sporadic bug) where we experience a weird UI sluggishness and then a load of out-of-OS-handle errors on Windows.

We think the culprit is 
https://github.com/eclipse/lsp4e/blob/5a75b544c083ab6b89ef11851dc07d2fcbe556ed/org.eclipse.lsp4e/src/org/eclipse/lsp4e/ConnectDocumentToLanguageServerSetupParticipant.java#L68

When the LS starts up for a project, it can report project-wide diagnostics, even for files that are not open. To do so the file is temporarily opened at https://github.com/eclipse/lsp4e/blob/5a75b544c083ab6b89ef11851dc07d2fcbe556ed/org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/diagnostics/LSPDiagnosticsToMarkers.java#L145

It seems that when a buffer opens, even temporarily, `ConnectDocumentToLanguageServerSetupParticipant` will get run and submit an Eclipse Job. If the user is running without 'always run in background' set, then these jobs each pop up a progress dialog.
Often these dialogs are never seen because the job completes so fast.

However, in the circumstances we have witnessed, with many many 'offline' files getting temporarily opened and progress dialogs created in a short space of time, that Eclipse can exhaust the OS handles for open shells.

ahmedneilhussain:
```
org.eclipse.swt.SWTError: No more handles
      at org.eclipse.swt.SWT.error(SWT.java:4944)
      at org.eclipse.swt.SWT.error(SWT.java:4833)
      at org.eclipse.swt.SWT.error(SWT.java:4804)
      at org.eclipse.swt.widgets.Widget.error(Widget.java:447)
      at org.eclipse.swt.widgets.Control.createHandle(Control.java:718)
      at org.eclipse.swt.widgets.Scrollable.createHandle(Scrollable.java:145)
      at org.eclipse.swt.widgets.Composite.createHandle(Composite.java:294)
      at org.eclipse.swt.widgets.Decorations.createHandle(Decorations.java:362)
      at org.eclipse.swt.widgets.Shell.createHandle(Shell.java:613)
      at org.eclipse.swt.widgets.Control.createWidget(Control.java:746)
      at org.eclipse.swt.widgets.Scrollable.createWidget(Scrollable.java:160)
[...]

PR #471, opened 2023-02-16
Experimental: disable document setup participant

ahmedneilhussain:
Following on from #420 and the discussion here https://github.com/eclipse/lsp4e/issues/392 I have done some investigation into the way documents get connected to the language server when you open them in Eclipse.

I added a couple of hooks to key methods to show what was calling them by walking down the stack trace when they were called, filtering out the general boilerplate classes to look for the underlying domain operation that triggered them.
I ran a few manual test cases in the IDE using Rust/Corrosion, just opening `.rs` files and scraping the stderr messages out of the console logs.

I then tried disabling `ConnectDocumentToLanguageServerSetupParticipant` in the plugin.xml

### DIFF
--- a/org.eclipse.lsp4e/plugin.xml
+++ b/org.eclipse.lsp4e/plugin.xml
@@ -359,14 +359,13 @@
          </includes>
       </viewerContentBinding>
    </extension>
-   <!--
    <extension
          point="org.eclipse.core.filebuffers.documentSetup">
       <participant
             class="org.eclipse.lsp4e.ConnectDocumentToLanguageServerSetupParticipant"
             contentTypeId="org.eclipse.core.runtime.text">
       </participant>
-   </extension> -->
+   </extension>
    <extension point="org.eclipse.ui.editors.annotationTypes">
       <type name="org.eclipse.lsp4e.read"></type>
       <type name="org.eclipse.lsp4e.write"></type>
--- a/org.eclipse.lsp4e/plugin.xml
+++ b/org.eclipse.lsp4e/plugin.xml
@@ -359,13 +359,14 @@
          </includes>
       </viewerContentBinding>
    </extension>
+   <!--
    <extension
          point="org.eclipse.core.filebuffers.documentSetup">
       <participant
             class="org.eclipse.lsp4e.ConnectDocumentToLanguageServerSetupParticipant"
             contentTypeId="org.eclipse.core.runtime.text">
       </participant>
-   </extension>
+   </extension> -->
    <extension point="org.eclipse.ui.editors.annotationTypes">
       <type name="org.eclipse.lsp4e.read"></type>
       <type name="org.eclipse.lsp4e.write"></type>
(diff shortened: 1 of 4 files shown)


ahmedneilhussain:
Here is a the console log for opening a rust file (having first made sure the LS is running):

image

ahmedneilhussain:
From the above observations, for Corrosion at least, `ConnectDocumentToLanguageServerSetupParticipant` is not needed:

1. There are many hooks added to the standard editor by LSP4e that fire when a document is opened.
2. All of them follow the core pattern in LSP4e (whether they use the old or new API) which has the pseudocode `getInitializedServer(doc).thenCompose(ensureConnected(doc)).thenApply(ls -> doSomethingWithLs())` and so will ensure the document is connected - whichever one fires first will trigger the actual connection.
3. `ConnectDocumentToLanguageServerSetupParticipant` is one of the last to trigger connection, probably because it has an extra layer of asynchronous delay by virtue of triggering via an Eclipse UI Job.

ahmedneilhussain:
This might not be true for other plugins, particularly linters and things like sonar cube. The functionality that indirectly triggers connection is optional as far as I can tell from a quick scan of the code: in all of the classes except `ConnectDocumentToLanguageServerSetupParticipant` there is a filter on `ServerCapabilities` present. Whilst primary language servers would probably offer at least some of these services (otherwise the user experience would not be very rich), presumably things like SonarCube and so on, whose main purpose is to just supply extra diagnostic markers, would not implement most of that functionality and so would not be connected without a class like `ConnectDocumentToLanguageServerSetupParticipant`.

ahmedneilhussain:
I thought that it might also be needed for programmatic connection, where the file buffer is loaded but an actual document editor not materialised in the GUI, but this appears not to be the case. I used the new 'format files' on a directory that had no files open in the UI and the `ConnectDocumentToLanguageServerSetupParticipant` was not triggered at all:

image

ahmedneilhussain:
I also note that the debug plugin wires in the `ConnectDocumentToLanguageServerSetupParticipant` - presumably for similar reasons.

ahmedneilhussain:
I've rewritten this class to be more lightweight and just use the standard `CompletableFuture` async calls to trigger connection in the background. This should avoid the weird cursor flickering/handle exhaustion problems that occur when the connect is wrapped in a UI job - I don't think the eclipse job really adds anything.

@rubenporras @mickaelistria what do you think?

mickaelistria:
The documentSetupParticipant is indeed intended to handle the `didOpen` early, so diagnostics or other notifications can be sent when opening the document, without user performing any actions. If some other strategy can be implemented to achieve the same result, it would be welcome; by semantics of the LSP, it would better be a workbench listener checking opened editors; the only tricky thing is that the same document can serve multiple editors (ie it can be opened simultaneously N times in Eclipse IDE while LSP assumes it's opened only once), but I suspect it's a problem that can be worked around by checking all editors.

In the meantime, I'll try to review this particular PR soon.

mickaelistria (approved):
This proposal seems good to me. I agree the Job is more noise than value.

ahmedneilhussain:
That's great @mickaelistria . Is it good to merge?

mickaelistria:
> Is it good to merge?

No objection from me; feel free to proceed.

PR #475, opened 2023-02-17
Maintenance/tidy ls startup

ahmedneilhussain:
The LS lifecycle code in `LanguageServerWrapper` has a lot of chaining of completable futures that mutate fields, which makes the class and its lifecycle quite tricky to reason about. Some monster lambda functions in the CF chaining makes it really hard to see the overall flow.

As a first step to improvements, I've taken the (straightforward, linear) domain code for declaring the client-side options that we support and split it out. Hopefully that makes LSW a bit more focused as well.

### DIFF
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServerWrapper.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServerWrapper.java
@@ -21,7 +21,6 @@ import java.io.File;
 import java.io.IOException;
 import java.net.URI;
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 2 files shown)


org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServerWrapper.java
```diff
@@ -439,20 +322,33 @@ public synchronized void start() throws IOException {
 		}
 	}
 
+	private CompletableFuture<InitializeResult> initServer(final URI rootURI) {
+		String name = "Eclipse IDE"; //$NON-NLS-1$
```
rubenporras, line 326: I would rather declare name as final and assign the default value in the else branch

org.eclipse.lsp4e/src/org/eclipse/lsp4e/SupportedFeatures.java
```diff
@@ -0,0 +1,143 @@
...
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *  Ahmed Hussain (Cocotec Ltd) - initial implementation
+ *
+ *******************************************************************************/
+package org.eclipse.lsp4e;
```
rubenporras, line 13: I would rather have the class in the packagae` org.eclipse.lsp4e.internal`

rubenporras (changes requested):
Just to minor comments, otherwise, I think it makes sense. Thanks

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServerWrapper.java
ahmedneilhussain: Used a ternary expression

org.eclipse.lsp4e/src/org/eclipse/lsp4e/SupportedFeatures.java
ahmedneilhussain: Sure, done.

ahmedneilhussain:
@rubenporras @mickaelistria I've rebased on master and addressed the minor comments.

mickaelistria (approved):
I didn't review in details; but the proposal looks much cleaner and worth it.

rubenporras (approved):
I also did not checked thoroughly that the code was moved right, but I trust you on that

rubenporras:
@ahmedneilhussain , shall we merge?

PR #476, opened 2023-02-20
Do not handle cancellation exceptions in LanguageServers

rubenporras:
Do not handle cancellation exceptions in LanguageServers, as each
consumers might want to react different to a cancellation

### DIFF
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServers.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServers.java
@@ -19,7 +19,6 @@ import java.util.List;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
-import java.util.concurrent.CompletionException;
 import java.util.concurrent.CompletionStage;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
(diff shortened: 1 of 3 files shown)


org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServers.java
```diff
@@ -80,7 +81,14 @@
 		final CompletableFuture<@NonNull List<T>> init = CompletableFuture.completedFuture(new ArrayList<T>());
 		return executeOnServers(fn).reduce(init, LanguageServers::add, LanguageServers::addAll)
 			// Ensure any subsequent computation added by caller does not block further incoming messages from language servers
-			.thenApplyAsync(Function.identity());
+			.thenApplyAsync(Function.identity())
+			.exceptionally(e -> {
+				if (isRequestCancelledException(e)) {
+					return Collections.emptyList();
+				}
```
mickaelistria, line 88: I don't think it's a good pattern to try to hide the cancellation or other exceptions to clients by default; particularly if we want to implement proper cancellation later: in some cases, a cancelled future should cause a retrigger the computation with updated parameters (eg document changed in between).

Also, as mentioned in other tickets, we'll most likely want the cancellation exception to be caught and propagate to call `.cancel()` on the parent futures.

rubenporras: @mickaelistria , yes, I agree with you, I have adapted the PR accordingly and undo a previous mistake that I did by putting logic in `LanguageServers`. Could you take a look again?

mickaelistria (approved):
I think it's good now. If you're OK with it, we can merge when CI is back to life and validates it.

rubenporras:
Yes, let us do that

PR #482, opened 2023-02-21
Deprecate methods in LanguageServerAccessor allowing access to a LanguageServer

rubenporras:
a LanguageServer


mickaelistria (changes requested):
Do we still use those in LSP4E? Usually removing all references inside the project is a requirement for proper deprecation.
Also, another rule we should try to follow is that all deprecated methods should have in their Javadoc an example or a snippet of how they should be replaced.

rubenporras:
I will add the documentation.

After all the PRs we did, not, we do not use them any longer, only in tests. Does it qualify as something we must also not do before setting them obsolete?

If I understand you right, if we do not use them and add documentation, this PR would make sense, right?

mickaelistria:
> After all the PRs we did, not, we do not use them any longer,

Good.
For `@Deprecated(forDeletion = true)` it is.

> only in tests. Does it qualify as something we must also not do before setting them obsolete?

Ideally yes.

> If I understand you right, if we do not use them and add documentation, this PR would make sense, right?

Yes.

mickaelistria (approved):

PR #483, opened 2023-02-21
New API: Port LSContentAssistProcessor

rubenporras:


rubenporras:
@ahmedneilhussain , does it makes sense for you? 

PS: I think this is the last PR in the saga, or do you think I have missed something?

ahmedneilhussain:
> @ahmedneilhussain , does it makes sense for you?
> 
> PS: I think this is the last PR in the saga, or do you think I have missed something?

Sorry @rubenporras I was knee-deep in our private fork merging in all the new stuff and back porting to Java 8 (it wasn't fun...). I'm taking a look now.

ahmedneilhussain:
I think it's OK as a straight port of the original code. I've noticed that this (and several of the other classes we've ported recently) have slightly esoteric logic for what essentially boils down to coupling a push to a pull: the Eclipse classes we're wiring in try and pull some data (presumably on the UI thread), but the LSP protocol is based around async 'push' of response data. To cater for this, these classes start a request, then block for a short period. If the request doesn't fully complete (on all LS) in that time, then they try and return any data 'so far' synchronously to the UI calling them, but schedule some follow-up for when the data is fully present. All of these classes do it slightly differently, and the resulting code isn't easy to follow.

It'd be nice to have a good standard pattern for doing this sort of thing. Having a think about it...

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/completion/LSContentAssistProcessor.java
```diff
@@ -65,10 +65,10 @@ public class LSContentAssistProcessor implements IContentAssistProcessor {
 	private IDocument currentDocument;
 	private String errorMessage;
 	private final boolean errorAsCompletionItem;
-	private CompletableFuture<List<@NonNull LanguageServer>> completionLanguageServersFuture;
+	private CompletableFuture<@NonNull List<@NonNull Void>> completionLanguageServersFuture;
```
mickaelistria, line 68: `@NonNull Void` seems strange as the only value accepted by a reference of type `Void` is null.

mickaelistria:
> It'd be nice to have a good standard pattern for doing this sort of thing. Having a think about it...

I think the best (and maybe even easiest in some cases) approach is to fix Platform so that it becomes better at handling async or long-running tasks.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/completion/LSContentAssistProcessor.java
rubenporras: Yes, it is weird but without it the JDT null analysis creates markers, so I think it is better like this, to avoid the marker.

Thanks for the review

rubenporras:
> > It'd be nice to have a good standard pattern for doing this sort of thing. Having a think about it...
> 
> I think the best (and maybe even easiest in some cases) approach is to fix Platform so that it becomes better at handling async or long-running tasks.

I still do not have an opinion, but will keep it as something to think about

PR #486, opened 2023-02-22
New API: Port CommandExecutor

rubenporras:


org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServiceAccessor.java
```diff
@@ -485,6 +484,13 @@ private static LanguageServerWrapper getLSWrapper(@Nullable IProject project,
 		}
 	}
 
+	public static LanguageServerWrapper getLSWrapperForConnection(@NonNull IDocument document,
```
rubenporras, line 487: @mickaelistria , the main question is we want to expose this new API or something else, I see that the existing `getLSWrapperForConnection` in this package has been marked as deprecated.

mickaelistria: I used to consider this method as a bad smell, something we'd like to get rid of; so making it API doesn't make me comfortable. However, now that we've established that LanguageServerWrapper is a proper API element, maybe it makes sense now.
But couldn't we instead add something like `LanguageServers.forDocument().withId(languageServerId)` to be more consistent with other operations and get rid of this odd method?

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServiceAccessor.java
```diff
@@ -337,7 +337,7 @@ public static Collection<LanguageServerWrapper> getLSWrappers(@NonNull final IFi
 	}
 
 	@NonNull
-	static Collection<LanguageServerWrapper> getLSWrappers(@NonNull final IDocument document) {
+	public  static Collection<LanguageServerWrapper> getLSWrappers(@NonNull final IDocument document) {
```
rubenporras, line 340: @mickaelistria , is this better thatn the method aI added in the last version?

rubenporras:
@mickaelistria , also in my latest commit, I tried to remove usage of `CommandExecutor` in favor of more straight forward code, as it is deprecated, but now there is a test failing. Is the use case that fails one that we need to support? If so it does not look right that `CommandExecutor` is deprecated, as there is no replacement.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServiceAccessor.java
rubenporras: So far @ahmedneilhussain and I discussed several times how have an executor with a single server, and it is a bit weird for the given API that expects consumers to deal with n servers, so far we refrained from such executor/filter. I could create such a filter but if it is acceptable to export the method I make now public, I think that is preferable.

mickaelistria:
Yes, that looks good to me. Is there a particular reason why this is still a draft?

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServiceAccessor.java
ahmedneilhussain: Yeah, I also wondered about a `forDocument().withId(languageServerId)` style method. But I think it was for this very use-case! When I went through the existing API usage by usage (see document attached https://github.com/eclipse/lsp4e/pull/251#issuecomment-1337648823 ) this class `CommandExecutor` was an awkward outlier.

I think if there's a rare one-off need then a one-off solution is OK. `CommandExecutor` is part of the framework so I think it's OK for it to have privileged access to low-level API that we don't want 'ordinary' users to be doing. It would be nice if this could be package-private but we'd need to move classes around a bit for that to work.

ahmedneilhussain: Actually you have now made the change I was about to suggest: all that this code is trying to do is make sure that the server that issued a command is the one that executed it.... but in fact we can just do that by storing the wrapper in the code mining. 👏

ahmedneilhussain:
@mickaelistria one thing about this class (nothing to do with @rubenporras changes) is why with the command executor we fall back to assuming a workspace edit and then blindly apply it! It seems quite a dangerous thing to have as a fallback... "we don't really recognise this message so let's turn it into a destructive operation and then hope for this best"!!!

Is there some historic explanation why this might have been a reasonable thing for the code to assume about an unknown command?

mickaelistria:
> Is there some historic explanation why this might have been a reasonable thing for the code to assume about an unknown command?

It's what most commands do in practice (return a WorkapceEdit), so it's a heuristic that works OK so far to support most command by default without specific handler for each ones (we're talking about dozens in case of JDT-LS); but indeed if we can refine it to prevent bad things from happening without requiring addition of dozens of handles, that'd be good.

rubenporras:
> Yes, that looks good to me. Is there a particular reason why this is still a draft?

I was not sure this was the right approach, I still had the feeling that it should be possible to make the code work without making any new API public. I have achieve that in https://github.com/eclipse/lsp4e/pull/491, so I will close this PR and focus on the other one (I think it will pass the test suite, and then it would be ready to merge.)

PR #487, opened 2023-02-22
Enable cancel-able requests via LanguageServerWrapper.execute

mickaelistria:


ahmedneilhussain:
Unfortunately I think this kind of approach just 'kicks the can further down the road', i.e. moves the problems slightly further.

This will only work if `fn`, which is entirely under the user's control, returns a `CompletableFuture` that has been constructed by the lsp4j layer with no further processing.

It also won't work if the user (or any other logic) chains further operations on the value returned from this function - calling cancel on the `CompletableFuture` objects that result from the chaining operations won't have any effect.

I think it's really hard to make this completely seamless because of the way cancellation propagates down, not up (although that aspect of the CompletableFuture design is a Good Thing overall: we wouldn't want cancellation to cancel the ls `start()` operations... I was a bit worried about that until I picked through the logic).

mickaelistria:
> This will only work if fn, which is entirely under the user's control, returns a CompletableFuture that has been constructed by the lsp4j layer with no further processing.

Yes, and this is why the documentation has been enriched to clarify it.

> It also won't work if the user (or any other logic) chains further operations on the value returned from this function - calling cancel on the CompletableFuture objects that result from the chaining operations won't have any effect.

Indeed, but it's semantically correct to expect consumers to cancel the request future if they want to cancel the LSP request. This approach has the benefit that cancellation becomes only usage of existing `.cancel()` methods, users don't need new API, they just need to cancel the request future if they want to get the LSP request cancelled. It's pretty consistent and thus relatively easy to grok; in the usual cases, people who don't get the LSP request cancelled will just realize "oh, I'm cancelling the wrong future", it's not so difficult (once people got trapped a few times) and it's really a very usual pattern of CompletableFuture so we're not inventing any new problem nor solution here.

> although that aspect of the CompletableFuture design is a Good Thing overall

This is more or less a dependency management problem: if you want to cut one leaf (eg the 2nd future), you don't want to cut the whole branch (ie it's parents and siblings). I don't think it's a problem we should try to hide, on the contrary, I would be in favor of strongly leveraging this specified semantics of CompletableFuture.


So despite being not ideal, despite CompletableFuture not being smart enough to guess what we actually want to cancel ("The annoying thing with computers is that they don't do what you want to do but what you ask them to do"), do you think this patch can be perceived as an improvement?

angelozerr:
Please see https://github.com/eclipse/lsp4mp/blob/6f2e1830b2a5ad12bf7956f8ce392f20db000bb8/microprofile.ls/org.eclipse.lsp4mp.ls/src/main/java/org/eclipse/lsp4mp/ls/MicroProfileWorkspaceService.java#L68

With this code when lsp client cancel the workspace symbol on server side it cancel the returned future and other future that we add with cancelIfNeed coing from the cancel checker that you can use it to know if lsp client has canceled the workppace symbol.

What do yiu think about this idea?

PR #489, opened 2023-02-23
Use CompletableFuture instead of CompletionStage

mickaelistria:
Facilitates chaining and cancellation.


mickaelistria:
@ahmedneilhussain IIRC you performed the same change in some other patch. I also had to deal with it. So I suggest we just use CompletableFuture everywhere instead of CompletionStage

mickaelistria:
My bad, this was included in https://github.com/eclipse/lsp4e/commit/f543bf6deeffaabc6f23ae4cd47c26076ffb38b8 (both commits were supposed to be separate but I failed at something apparently)

PR #491, opened 2023-02-26
New API: Port CodeActionMarkerResolution

rubenporras:


org.eclipse.lsp4e/src/org/eclipse/lsp4e/command/CommandExecutor.java
```diff
@@ -243,11 +143,11 @@ private static ParameterizedCommand createEclipseCoreCommand(@NonNull Command co
 	 * Very empirical and unsafe heuristic to turn unknown command arguments into a
 	 * workspace edit...
 	 */
-	private static WorkspaceEdit createWorkspaceEdit(List<Object> commandArguments, IDocument document) {
-		final var res = new WorkspaceEdit();
+	private static WorkspaceEdit createWorkspaceEdit(List<Object> commandArguments, @NonNull IResource resource) {
```
mickaelistria, line 146: We'd still want it to work on any IDocument, not only IResource. Imagine one opens a .js file in Eclipse IDE with Wild Web Developer, then the language server starts and can provide code actions with workspace edits. This change here would cause the codeAction to stop functioning (or at least it would give that impression in the API).

mickaelistria (changes requested):
Making resource required at API-level for toWorkspaceEdit is too restrictive.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/command/CommandExecutor.java
rubenporras: I have added an API that accepts a IDocument just to have a clear API, but I think the resource works fine, because the only thing we take from the document is the path, and we only support file based documents, and that we can take from the resource as well.

mickaelistria: > I have added an API that accepts a IDocument just to have a clear API

I don't see it.

> I think the resource works fine,

Yes, but there is not always a resource, while there is always a document or a path, so the API should express it works with any document/path. If path/URI is more convenient, it's good to use it as well.

rubenporras: There is now `public static CompletableFuture<Object> executeCommandClientSide(@NonNull Command command, @NonNull IDocument document)` in the `CommandExecutor` which accepts a IDocument.

mickaelistria: Ah OK sorry, and this one is private. So it's OK as it.

mickaelistria (approved):

Issue #492, opened 2023-02-27
NPE when using multiple languages client for same file extension 


sbouchet:
using latest snapshot from https://download.eclipse.org/lsp4e/snapshots/. when opening a simple java file, i get the following npe.
```
Thread [main] (Suspended (exception java.lang.NullPointerException))	
	owns: org.eclipse.swt.widgets.RunnableLock  (id=327)	
	java.lang.String.compareTo(java.lang.String) line: 2013	
	java.lang.String.compareTo(java.lang.Object) line: 140	
	java.util.Comparator<T>.lambda$comparing$77a9974f$1(java.util.function.Function, java.lang.Object, java.lang.Object) line: 473	
	java.util.Comparator$$Lambda$69.0x0000000800cc0908.compare(java.lang.Object, java.lang.Object) line: not available	
	java.util.Comparators$NullComparator<T>.compare(T, T) line: 85	
	java.util.TimSort<T>.countRunAndMakeAscending(T[], int, int, java.util.Comparator<? super T>) line: 355	
	java.util.TimSort<T>.sort(T[], int, int, java.util.Comparator<? super T>, T[], int, int) line: 220	
	java.util.Arrays.sort(T[], int, int, java.util.Comparator<? super T>) line: 1307	
	java.util.ArrayList<E>.sort(java.util.Comparator<? super E>) line: 1721	
	org.eclipse.lsp4e.LanguageServers$LanguageServerProjectExecutor.getServers() line: 347	
	org.eclipse.lsp4e.LanguageServers$LanguageServerProjectExecutor(org.eclipse.lsp4e.LanguageServers<E>).anyMatching() line: 222	
	org.eclipse.lsp4e.operations.codeactions.LSPCodeActionMarkerResolution.checkMarkerResoultion(org.eclipse.core.resources.IMarker) line: 128	
	org.eclipse.lsp4e.operations.codeactions.LSPCodeActionMarkerResolution.hasResolutions(org.eclipse.core.resources.IMarker) line: 214	
	org.eclipse.ui.internal.ide.registry.MarkerHelpRegistry.hasResolution(org.eclipse.core.resources.IMarker, org.eclipse.core.runtime.IConfigurationElement, org.eclipse.ui.IMarkerResolutionGenerator) line: 264	
	org.eclipse.ui.internal.ide.registry.MarkerHelpRegistry.hasResolutions(org.eclipse.core.resources.IMarker) line: 234	
[...]

PR #493, opened 2023-02-28
Fix NPE

rubenporras:
for https://github.com/eclipse/lsp4e/issues/492

### DIFF
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServers.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServers.java
@@ -15,7 +15,7 @@ package org.eclipse.lsp4e;
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Collection;
-import java.util.Comparator;
+import java.util.Collections;
 import java.util.List;
 import java.util.Objects;
 import java.util.Optional;
@@ -164,8 +164,8 @@ public abstract class LanguageServers<E extends LanguageServers<E>> {
 		return result.thenApplyAsync(Function.identity());
 	}
 
-	public @NonNull E withComparator(final @Nullable Comparator<? super LanguageServerWrapper> comparator) {
-		this.comparator = comparator;
+	public @NonNull E withPreferredServer(final @Nullable String serverId) {
+		this.serverId = serverId;
 		return (E)this;
 	}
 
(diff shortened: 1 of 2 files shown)


org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServers.java
```diff
@@ -458,5 +463,6 @@ private <T> void completeEmptyOrWithException(final CompletableFuture<Optional<T
 
 	private @NonNull Predicate<ServerCapabilities> filter = s -> true;
 
-	protected @Nullable Comparator<? super LanguageServerWrapper> comparator;
+	protected @Nullable String serverId;
+;
```
mickaelistria, line 467:
```suggestion

```

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServers.java
```diff
@@ -164,11 +164,12 @@ public <T> CompletableFuture<Optional<T>> computeFirst(BiFunction<? super Langua
 		return result.thenApplyAsync(Function.identity());
 	}
 
-	public @NonNull E withComparator(final @Nullable Comparator<? super LanguageServerWrapper> comparator) {
-		this.comparator = comparator;
+	public @NonNull E withPreferredServer(final @Nullable String serverId) {
```
mickaelistria, line 167: Could we use a LanguageServerDefiniton here? Stronger typing usually makes code less error-prone.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServers.java
rubenporras: thanks.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServers.java
rubenporras: the consumer does not have the `LanguageServerDefiniton`, it only has the id, so I think that for this PR is it fine to leave it as this.

mickaelistria: The consumer can get the languageServerDefinition from the id (`LanguageServerRegistry.getDefinition(id)`). Really, one never regrets stronger typing.

rubenporras: Ok, done

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/diagnostics/LSPDiagnosticsToMarkers.java
```diff
@@ -207,7 +216,7 @@ private IMarker getExistingMarkerFor(IDocument document, Diagnostic diagnostic,
 				if (LSPEclipseUtils.toOffset(diagnostic.getRange().getStart(), document) == MarkerUtilities.getCharStart(marker)
 						&& LSPEclipseUtils.toOffset(diagnostic.getRange().getEnd(), document) == MarkerUtilities.getCharEnd(marker)
 						&& Objects.equals(marker.getAttribute(IMarker.MESSAGE), diagnostic.getMessage())
-						&& Objects.equals(marker.getAttribute(LANGUAGE_SERVER_ID), this.languageServerId)) {
+						&& Objects.equals(marker.getAttribute(LANGUAGE_SERVER_ID), languageServerDefinition)) {
```
mickaelistria, line 219: I believe marker do require Serializable objects of something of that sort as they are persisted in the workspace. If so, we'll need to stick to id.

rubenporras: Okay, then back to the roots. If it passes the tests I will merge it. We can always improve later if wanted

Issue #38, opened 2022-01-03
Conflicting keybinding for Open Type


trancexpress:
On Eclipse bugzilla the issue was: https://bugs.eclipse.org/bugs/show_bug.cgi?id=527785

A conflict occurred for CTRL+SHIFT+T:
Binding(CTRL+SHIFT+T,
	ParameterizedCommand(Command(org.eclipse.lsp4e.symbolinworkspace,Go to Symbol in Workspace,
		,
		Category(org.eclipse.lsp4e.category,Language Servers,null,true),
		org.eclipse.ui.internal.WorkbenchHandlerServiceHandler@101eb529,
		,,true),null),
	org.eclipse.ui.defaultAcceleratorConfiguration,
	org.eclipse.ui.contexts.window,,,system)
Binding(CTRL+SHIFT+T,
	ParameterizedCommand(Command(org.eclipse.jdt.ui.navigate.open.type,Open Type,
		Open a type in a Java editor,
		Category(org.eclipse.ui.category.navigate,Navigate,null,true),
		org.eclipse.ui.internal.WorkbenchHandlerServiceHandler@6a9b8f33,
		,,true),null),
	org.eclipse.ui.defaultAcceleratorConfiguration,
	org.eclipse.ui.contexts.window,,,system)

garretwilson:
I installed Eclipse EE 2022-12, and now when I try to find a Java class using `Ctrl+Shift+T`, I first get a menu asking if I want to "Go to Symbol in Workspace" or "Open Type". I can keep doing this over and over and it shows these conflicting mapped commands.

But oddly if I `Ctrl+Click` on a class to open another file, and then close the file going back to the original class source code, then `Ctrl+Shift+T` brings up "Open Type" as I would expect.

My error log shows

```
eclipse.buildId=4.26.0.20221201-1200
java.version=17.0.5
java.vendor=Eclipse Adoptium
BootLoader constants: OS=win32, ARCH=x86_64, WS=win32, NL=en_US
Framework arguments:  -product org.eclipse.epp.package.jee.product
Command-line arguments:  -os win32 -ws win32 -arch x86_64 -product org.eclipse.epp.package.jee.product

org.eclipse.jface
Warning
Thu Jan 05 09:41:46 BRT 2023
Keybinding conflicts occurred.  They may interfere with normal accelerator operation.

eclipse.buildId=4.26.0.20221201-1200
java.version=17.0.5
java.vendor=Eclipse Adoptium
[...]

PR #494, opened 2023-02-28
Disable LSP rename handler for Java file

angelozerr:
This PR disable LSP rename handler for Java file. 

The problem has been reported here https://github.com/jbosstools/jbosstools-quarkus/issues/199

The JBoss Quarkus maps the MicroProfile language server to a Java file. Once the plugin is installed, it is not possible to use `Alt+Shift+R` because I think there is a conflict between LSP4E and JDT which is really annoying.

I have never seen some language server which support rename in Java file, it is the reason why this PR disable rename handler for Java file.

### DIFF
--- a/org.eclipse.lsp4e/plugin.xml
+++ b/org.eclipse.lsp4e/plugin.xml
@@ -104,7 +104,24 @@
             class="org.eclipse.lsp4e.operations.rename.LSPRenameHandler"
             commandId="org.eclipse.ui.edit.rename">
          <activeWhen>
-            <reference definitionId="org.eclipse.lsp4e.textSelectionHasLanguageServer" />
+            <and>
+               <!-- LSP rename handler is active when: -->
+               <!-- 1. The file is not a Java file to avoid conlict with JDT UI rename handler -->
+               <not>
+                 <with variable="editorInput">
+                    <or>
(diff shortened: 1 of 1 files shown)


mickaelistria (changes requested):
We won't create a particular dependency in LSP4E between LSP4E and JDT. We need to find another solution.

angelozerr:
This problem is very annoying because when you install JBoss Quarkus, we loose the capability to use `Alt+Shift+R` in Java editor.

Instead of excluding JDT Java source, we could include only editor which is the generic editor. What do you think @mickaelistria ?

angelozerr:
@JessicaJHee I asked at https://github.com/microsoft/vscode-languageserver-node/issues/1198

sbouchet:
that's an already known issue : https://github.com/eclipse/lsp4e/issues/38#issuecomment-1399985169

sbouchet:
see also #38

mickaelistria:
I'm closing this in LSP4E, it's more properly actionable on JDT end: https://github.com/eclipse-platform/eclipse.platform.ui/issues/513#issuecomment-1468778600 , or even in the jbosstools-quarkus plugin to redefine the binding to force one over the other.

PR #495, opened 2023-02-28
Remove last references to LSPDocumentInfo outside of the test plugin

rubenporras:


rubenporras:
This are the last usages of LSPDocumentInfo, maybe I am mistaken but it looks to me like dead code that we can just remove. Is this truly just dead code by now?

mickaelistria:
LSPDocumentInfo could then be made `@Deprecated(forRemoval=true)` just like https://github.com/eclipse/lsp4e/blob/master/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServiceAccessor.java#L629

rubenporras:
> LSPDocumentInfo could then be made `@Deprecated(forRemoval=true)` just like https://github.com/eclipse/lsp4e/blob/master/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServiceAccessor.java#L629

I have already made it `@Deprecated(forRemoval=true)` in the same PR you mentioned, so if we can just remove it here, I think we are done.

mickaelistria:
> I have already made it @Deprecated(forRemoval=true) in the same PR you mentioned, so if we can just remove it here, I think we are done.

I don't think we should remove it in this release as it was part of the API in previous release.
Let's stick to deprecated for removal at the moment; we'll remove it some time later.

PR #499, opened 2023-02-28
Remove usages of getLanguageServers

rubenporras:


rubenporras:
@ahmedneilhussain , as promised. Does it make sense for you?

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServers.java
```diff
@@ -212,6 +213,21 @@ protected Boolean matches(@NonNull CompletableFuture<@Nullable LanguageServerWra
 		return Boolean.FALSE;
 	}
 
+	protected Boolean matches(@NonNull CompletableFuture<@Nullable LanguageServerWrapper> wrapperFuture, Consumer<Boolean> consumer) {
```
mickaelistria, line 216: what's the consumer here? What about just returning a `CompletableFuture<Boolean>` and consumer can decide to wait or not, and for how long?

rubenporras: @mickaelistria , @ahmedneilhussain, I think we have several `AbstractHandler` that would all benefit from having support that allows consistent behavior   without handling timeouts and futures themselves, so to me it is worth having this small API. I have now updated the PR with the consumers that I think should use the new API.

Probably LSPFormatFilesHandler should also use the same logic but there the refactor was not straight forward so I have not done it to save myself the work if you still do not like the approach.

rubenporras: @mickaelistria , @ahmedneilhussain , I have a second proposal https://github.com/eclipse/lsp4e/pull/507, which puts the logic outside of `LanguageServers` and also consolidates all the `AbstractHandlers` in a common implementation. I think it is better than this PR. Would you like to take a look at https://github.com/eclipse/lsp4e/pull/507?

PR #505, opened 2023-02-28
Restore used API method

mickaelistria:
And mark it for deprecation with a migration hint

### DIFF
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/command/CommandExecutor.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/command/CommandExecutor.java
@@ -27,10 +27,10 @@ import org.eclipse.core.commands.NotHandledException;
 import org.eclipse.core.commands.ParameterType;
 import org.eclipse.core.commands.ParameterizedCommand;
 import org.eclipse.core.commands.common.NotDefinedException;
-import org.eclipse.core.resources.IFile;
 import org.eclipse.core.resources.IResource;
 import org.eclipse.core.resources.ResourcesPlugin;
 import org.eclipse.core.runtime.IPath;
+import org.eclipse.core.runtime.Path;
 import org.eclipse.jdt.annotation.NonNull;
 import org.eclipse.jdt.annotation.Nullable;
 import org.eclipse.jface.text.IDocument;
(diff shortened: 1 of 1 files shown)


rubenporras:
I am sorry that the refactor caused some trouble.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/command/CommandExecutor.java
```diff
@@ -64,23 +64,25 @@ public class CommandExecutor {
...
-		IFile file = LSPEclipseUtils.getFile(document);
-		if (file != null) {
-			return CommandExecutor.executeCommandClientSide(command, file);
-		}
-		return null;
+	/**
+	 * @deprecated use {@link #executeCommandClientSide(Command, IDocument)}
+	 */
+	@Deprecated(forRemoval = true)
+	public static CompletableFuture<Object> executeCommand(@Nullable Command command, @Nullable IDocument document,
```
rubenporras, line 71: If we restore this method, I think it should also do the server request, otherwise we keep the API stable but it does not do the same as before, which I think it is more error prone than the breaking change + asking consumer to refactor. Since I broke it, If you want that I can do that, but given that the class was deprecated already and that we want to promote the clean pattern which the old code did not support, I would be rather in favor of breaking the API.

mickaelistria: I agree with it; but we can proceed as you prefer: either we merge it and then re-add server-side execution; or we try to fix it before merging, or we decide it's not that critical to support server-side command executions here. FWIW, I checked on GitHub search and in the workspace, and I think all consumers I've noticed do assume it's client-side, so loosing server-side commands here will not be perceived as a regression by those, ie there is most likely no actual problem to solve...

org.eclipse.lsp4e/src/org/eclipse/lsp4e/command/CommandExecutor.java
```diff
@@ -64,23 +64,25 @@ public class CommandExecutor {
...
+			@Nullable String languageServerId) {
+		return executeCommandClientSide(command, document);
 	}
 
-	@SuppressWarnings("unused") // ECJ compiler for some reason thinks handlerService == null is always false
-	public static CompletableFuture<Object> executeCommandClientSide(@NonNull Command command, @NonNull IResource resource) {
+
+	public static CompletableFuture<Object> executeCommandClientSide(@NonNull Command command, @Nullable IDocument document) {
```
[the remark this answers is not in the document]
mickaelistria: I see it in compilation errors (some JDT-LS issue cause "build automatically" to be often disabled, so I miss a lot of errors in my workspace). I'll work on fixing those.

mickaelistria:
> I am sorry that the refactor caused some trouble.

No need to be sorry about that! This is a deprecated class, I got reviewed approval, there are no tests for it... It's not this particular change that caused the breakage, but a sum of issues or weakness that have stacked and made your change showcase a symptom of all that, not the only cause.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/command/CommandExecutor.java
rubenporras: If all consumers assume it's client-side, then what about just https://github.com/eclipse/lsp4e/pull/506?

mickaelistria: This will restore the method indeed. But we also need the command execution to work with non-IResource document, as I believe it used to (although I'm not 100% sure) and to be able to generate a workspaceEdit from a URI without requiring an IResource.

rubenporras: Okay, I will look at how to restore that behavior as well, later. I think it should be a small adaptation to #506

rubenporras:
If you are happy with https://github.com/eclipse/lsp4e/pull/506 I would submit it.

mickaelistria:
Replaced by #508

PR #508, opened 2023-03-01
Deprecate for removal methods which grant access to the LanguageServer

rubenporras:


org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServerWrapper.java
```diff
@@ -480,23 +480,47 @@ public synchronized void stop() {
...
+	@Deprecated(forRemoval = true)
 	public @Nullable CompletableFuture<LanguageServer> connect(IDocument document) throws IOException {
+		CompletableFuture<LanguageServerWrapper> connect2 = connect2(document);
+		if (connect2 != null) {
+			return connect2.thenApply(theVoid -> languageServer);
+		}
+		return null;
+	}
+
+	public @Nullable CompletableFuture<LanguageServerWrapper> connect2(IDocument document) throws IOException {
```
rubenporras, line 519: since java does not support overloading return values, I could not find anything better than this. In the next release we can deprecate this one and just change the return value of connect.

mickaelistria: or we can just call it `connectDocument` ?

rubenporras: good idea. I have renamed it.

Issue #504, opened 2023-02-28
Align progress handler to VSCode language-client


BoykoAlex:
LSP4E `LSPProgressManager` expects the following sequence of messages:
1. Begin message that has total amount of work in percentage field not necessarily 100, could be any number.
2. Progress report message that has the work done units in the percentage field, i.e. some intger number typically less or equal the value in 1.

VSCode language client expects the following sequence of messages
1. Begin message with 0 as initial percentage value
2. Progress report message that has a percentage value between 0 and 100
(See https://github.com/microsoft/vscode-languageserver-node/blob/main/client/src/common/progressPart.ts#L81)

I'd probably keep the implementation the same as in the VSCode language client. Percentage value hints at the value to be between 0 and 100.

In addition for the indefinite progress where instead of percentage string messages are passed I never saw the string message. Likely because in this case i didn't supply the percentage value. Probably because of the initial check in `void report(final WorkDoneProgressReport report, final IProgressMonitor monitor)` method from the `LSPProgressManager`

BoykoAlex:
Fixed with 9a5e1bcbd3d148ba54ee509277c2200facf2eff1

PR #509, opened 2023-03-01
Allow progress report messages to be applied for indefinite progress

BoykoAlex:
Partially address #504  - the last part of it.

I can submit the full PR for #504. Essentially the total work for the progress would be hardcoded to 100 and the begin message percentage would be assumed 0 if it is a non null integer.

### DIFF
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/progress/LSPProgressManager.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/progress/LSPProgressManager.java
@@ -128,7 +128,11 @@ public class LSPProgressManager {
 	private void begin(final WorkDoneProgressBegin begin, final IProgressMonitor monitor) {
 		Integer percentage = begin.getPercentage();
 		if (percentage != null) {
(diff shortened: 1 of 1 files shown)


rubenporras (approved):

rubenporras:
Would the rest of #504 be backwards compatible?

BoykoAlex:
> Would the rest of #504 be backwards compatible?

@rubenporras  I wasn't thinking about backward compatibility, but i think it can be made backward compatible...

BoykoAlex:
@rubenporras  Do I need to wait for someone to merge this PR or merge it myself?

BoykoAlex:
@rubenporras i have pushed the commit that I think resolves #504 with backward compatibility support

rubenporras (approved):

rubenporras:
> > Would the rest of #504 be backwards compatible?
> 
> @rubenporras I wasn't thinking about backward compatibility, but i think it can be made backward compatible...

Thanks, I made the same mistake in the Xtext Language Server as in LSP4E, so not being backward compatible would have broken at least that server

PR #510, opened 2023-03-01
Add Icon for Struct Item Kind

rubenporras:
Taken from
https://github.com/eclipse-cdt/cdt/tree/7c6bd5bdcb4cf8c31050aeeb102250a27a157728/core/org.eclipse.cdt.ui/icons/obj16


rubenporras:
Is it okay to take an icon from the CTD? Does the license allow that?

mickaelistria:
> Is it okay to take an icon from the CTD? Does the license allow that?

Yes, we can freely copy code and resources between Eclipse projects that have the same license.

PR #513, opened 2023-03-03
Basic call hierarchy view support added.

andrewL-avlq:
Basic call hierarchy view added that uses the LSP prepareCallHierarchy and callHierarchyIncomingCalls APIs to provide content.


mickaelistria (approved):
This seems like a good addition!

rubenporras:
This is copy of what we tested internally on the current released LSP4E adapted to use the new API in `LanguageServers`. We will soon update our product to Eclipse 2022-12 and will check if some error got in when migrating to the new API. We also plan to write some tests in the future.

PR #517, opened 2023-03-03
Semantic Highlighting: Merge rather than replace styles

ahmedneilhussain:
@rubenporras what do you think of this?

Semantic highlighting is causing problems with Coco's eclipse plugin. We don't use TM4e on the eclipse side, we have our own editor subclass and implement the various highlighting and bracket-matching hooks directly. Semantic highlighting assumes text mate highlighting is active (not unreasonably, since there needs to be some kind of assumed presentation for the server's tokenisation!) and clobbers everything else.

However, I've also observed similar behaviour with Corrosion (the Rust plugin, which I use in my lsp4e dev workspace as a de facto test harness). When you open a document you initially get colors from the TM highlighting, but when the semantic highlighting kicks in after a second or so, almost all of the colours get replaced as black.

Before:
image

After short delay:
image

If I make the change in this PR then the result looks like this:

image

### DIFF
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/semanticTokens/SemanticHighlightReconcilerStrategy.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/semanticTokens/SemanticHighlightReconcilerStrategy.java
@@ -285,7 +285,7 @@ public class SemanticHighlightReconcilerStrategy
 		documentTimestampAtLastAppliedTextPresentation = DocumentUtil.getDocumentModificationStamp(document);
 		IRegion extent = textPresentation.getExtent();
 		if (extent != null) {
-			textPresentation.replaceStyleRanges(styleRangeHolder.overlappingRanges(extent));
+			textPresentation.mergeStyleRanges(styleRangeHolder.overlappingRanges(extent));
 		}
 	}
 }


ahmedneilhussain:
I'm not suggesting this fix for merge, necessarily, I'd just like some help understanding what's going on and advice about what the best behaviour should be. I think for Coco's case (non-tm4e based editor) there's no sensible thing the semantic highlighting can really do - it needs to get the colouring from somewhere! In this case, perhaps there should be a better way of disabling semantic highlighting per content type? In the LSP4e README it does say that people 'generally' pair lsp4e with tm4e, but it doesn't mandate it.

I'm more unsure why I'm seeing the behaviour I am with Corrosion, since that *does* use tm4e AIUI. @mickaelistria can you comment, as I believe you are involved with both Corrosion and tm4e?

From what the class comments say, @rubenporras , it appears the intention is that the semantic highlighting gets merged with the 'vanilla' highlighting, but the line I change looks as though it will clear any existing styling. Is it possible that the order in which the two reconciliation algorithms are applied makes a difference, and is not always predictable?

Is it also possible the extent to which this is a problem will depend on the server & language: if the server provides semantic highlighting for almost every token in the document then the 'vanilla' highlighting won't ever make much difference.

mickaelistria:
Semantic Highlighting uses TM4E "just" to turn the response into some styling. TM4E, when used with a grammar, also uses a TextMate grammar to everaluate the styles to apply.
So TM4E TextMate are 2 distinct sources of styling/highlighting and are possibly conflicting. I think the policy so far is that one fully replaces the other, but merging may be a better default. I defer to @rubenporras on what seems best here.

rubenporras:
Hi @ahmedneilhussain 

my first impression is that if with Corrosion almost all colors are replaced as black, that is because no stylesheet has been configured. What do you think of https://github.com/eclipse/lsp4e/pull/522/files? If you think it is a valid approach, would you mind testing it with corrosion?

> We don't use TM4e on the eclipse side, we have our own editor subclass and implement the various highlighting and bracket-matching hooks directly.

Regarding this, what do you think of having the possibility to contribute your own org.eclipse.lsp4e.operations.semanticTokens.TokenTypeMapper so that Semantic Highlight would work without TM4E?

Regards

rubenporras:
@ahmedneilhussain , I think that semantic tokens is a more refined "tokenizer" than syntactic tokens, so if they differ on what token the editor should display, I think the semantic information should replace the syntactic one, so replace is wanted here. I will merge my PR which I think will avoid the behavior as well except for the case where an item is obsolete, there the code will still create an empty style if none is found.
I think that to solve that, we would need to add the possibility to contribute an own token type mapper, be it with an adapter or with a different method.

rubenporras:
We can close this one now, right?

ahmedneilhussain:
> We can close this one now, right?

Sure.

Issue #518, opened 2023-03-05
[Feature Request]: "Language Server Processes" view


sebthom:
I'd like to suggest an Eclipse view that shows all spawned language server processes, showing language server name, initial project, pid, command line and a button to terminate a language server.

Sometimes language servers start misbehaving and only restarting them helps. This means either closing all editor windows and waiting for X seconds, or restarting eclipse, or manually killing the respective processes in the task manager.

mickaelistria:
Why a view? OSs generally provide the feature of listing processes and children processes very well and a user who can know whether/why/which language server to kill should usually be able to identify the process.

sebthom:
> Why a view? OSs generally provide the feature of listing processes and children processes very well and a user who can know whether/why/which language server to kill should usually be able to identify the process.

I sometimes see half a dozen node.exe processes in the task manager and have no idea which one to kill.

sebthom:
My personal intention is general Eclipse users.

sebthom:
But I now also find it very convenient while troubleshooting a language server.

martinlippert:
From my point of view, this view should not be visible by default in any of the usual perspectives and the user should not see this if everything just goes well.

But if something goes wrong or behaves in a strange way, I would find such a view quite useful in order to get an idea about all the associated and running language server processes. For that purpose (troubleshooting) I would find it even useful to enhance this view with a few additional actions, like the ability to see/jump to the log output of the language server process (might require some work behind the scenes to have a homogenous way to get the log output), stop the process (maybe to get it restarted), see the consumed memory, or even more.

In general, all this information can already be found somewhere else (using `jps`, other command line tools), but such a view would make it a lot easier to troubleshoot issues when users report issues, I think.

sebthom:
> From my point of view, this view should not be visible by default in any of the usual perspectives and the user should not see this if everything just goes well.
 
Definitely, this view needs to be manually opened.

> For that purpose (troubleshooting) I would find it even useful to enhance this view with a few additional actions, like the ability to see/jump to the log output of the language server process (might require some work behind the scenes to have a homogenous way to get the log output), stop the process (maybe to get it restarted), see the consumed memory, or even more.

I implemented stopping of processes in #519. This is actually the main reason I would like to have such a view.
I also wondered about displaying memory consumption but I don't really really know how to get this metric from external processes in a portable way.

PR #519, opened 2023-03-06
Add Language Servers View

sebthom:
This PR addresses #518 and introduces a new "Language Server Processes" view, looking like this:

image

image


sebthom:
This PR currently has the following issues, where I would like to ask for feedback/suggestions:
1. The view is refreshed via a polling job every two seconds. Would be nice to have a way to get notified about started/stopped LS processes from LSP somehow
2. The view needs access to the list of currently running LS processes, as a quick hack I temporarily set visibility of `LanguageServiceAccessor.startedServers` to public.
3. The view also needs access to the `LanguageServerWrapper.lspStreamProvider` field to retrieve PID, command args etc. - I again temporarily set the visibility to public.

sebthom:
@ahmedneilhussain since you are doing a lot of refactoring lately maybe you have some ideas regarding the open issues?

ahmedneilhussain:
I'm looking at the LS lifecycle code at the moment in an attempt to tidy it up and make the started/stopped/error states a little bit more explicit and atomic. I'll see what can be done.

ahmedneilhussain:
Hi @sebthom I've been busy wading through the LS lifecycle code (in `LanguageServerWrapper` - there's some more in `LanguageServiceAccessor`). See #520 for the gory details.

I think it would be relatively straightforward to have LSW instances call a method on `LanguageServiceAccessor` to notify it of lifecycle events. The latter could then support subscribers (probably just your view, in practice!) to have these events forwarded on from all wrappers in the standard Observable pattern.

Start and stop are obviously explicitly called methods that are easy to hook into. I think we could also add a 'whenComplete' hook onto `launcherFuture` since that completes when the stream to the server is closed, which is a way of detecting crashes.

I think we could also expose some of the server info on `ServerConnection` in my PR.

sebthom:
The fields on `ServerConnection` look very useful!

angelozerr:
Just for your informartion, I have ported LSP4E classes (LanguageServerWrapper, etc) in IJ and I provide a LSP consoles which show both the started language server and the console. IMHO I think it is better to have processes and console in the same view.

Here a screenshot:

image

I have managed too lifecycle (starting, stopping, started, stopped) of language server to update UI according the sate:


IJLSPConsole

sebthom:
@angelozerr that looks really nice... unfortunately you implemented it for the "wrong" IDE :-)

I don't have the resources to match this in Eclipse atm. I still believe that the language server view offered by this PR is a worthwhile addition and can be extended later if required.

angelozerr:
> @angelozerr that looks really nice... unfortunately you implemented it for the "wrong" IDE :-)

Indeed I'm an Eclipse user, but I love discover anything like vscode and IJ. I wanted just to share with you my feature that I think it could be very nice on LSP4E.

> I don't have the resources to match this in Eclipse atm. I still believe that the language server view offered by this PR is a worthwhile addition and can be extended later if required.

yes sure!

rubenporras:
@sebthom , I have not followed this PR, but it would be great if we could merge before the new release is done. Is it in an state where could merge it?

mickaelistria (changes requested):
I'd like to release LSP4E today or tomorrow, and I'm not comfortable with the idea of merging it big chunks of lowly-tested code into the codebase just before a release.

I also have concerns with the code that is relatively intrusive (introducing interfaces that core APIs must implement rather than using adapters) while the view itself would rather be totally standalone, and maybe even worth being part of a seperate bundle.
I will try to give a more precise review after the release.

Additionally to those technical concerns, I have a "strategical" one: if we give possibility to workaround possible issues with LS by giving control over the process, doesn't it bring a risk that people just rely more and more on workarounds rather than actually fixing the causes?...

org.eclipse.lsp4e/src/org/eclipse/lsp4e/ui/LanguageServerProcessesView.java
```diff
@@ -0,0 +1,270 @@
...
+	}
+
+	@Override
+	public void setFocus() {
+		viewer.getControl().setFocus();
+	}
+
+	private void updateViewerInput() {
+		final var currentElements = (Object[]) viewer.getInput();
+		final var newElements = LanguageServiceAccessor.startedServers.stream().filter(LanguageServerWrapper::isActive)
```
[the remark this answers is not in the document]
sebthom: I don't see how. Each `LanguageServers` instance is for a specific document or project and obtained via `LanguageServers#forDocument` or `LanguageServers#forProject`. I need a list of all running LS no matter for which project or document.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/ui/LanguageServerProcessesView.java
```diff
@@ -0,0 +1,270 @@
...
+import org.eclipse.swt.events.SelectionAdapter;
+import org.eclipse.swt.events.SelectionEvent;
+import org.eclipse.swt.widgets.Composite;
+import org.eclipse.swt.widgets.TableItem;
+import org.eclipse.swt.widgets.ToolBar;
+import org.eclipse.swt.widgets.ToolItem;
+import org.eclipse.swt.widgets.Widget;
+import org.eclipse.ui.part.ViewPart;
+
+public class LanguageServerProcessesView extends ViewPart {
```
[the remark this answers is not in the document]
sebthom: Done.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/server/ProcessInfo.java
```diff
@@ -0,0 +1,23 @@
...
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *  Sebastian Thomschke (Vegard IT GmbH) - initial implementation
+ *******************************************************************************/
+package org.eclipse.lsp4e.server;
+
+import org.eclipse.jdt.annotation.Nullable;
+
+public interface ProcessInfo {
```
mickaelistria, line 16: I don't think we need such an interface, using a adapter to ProcessHandle would reduce coupling and code to maintain while augmenting the capability of the feature.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/server/ProcessStreamConnectionProvider.java
```diff
@@ -101,7 +113,7 @@ public void setCommands(List<String> commands) {
 		this.commands = commands;
 	}
 
-	protected @Nullable String getWorkingDirectory() {
+	public @Nullable String getWorkingDirectory() {
```
mickaelistria, line 116: This becomes useless if we manage to use ProcessHandle.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServerWrapper.java
```diff
@@ -369,6 +370,27 @@ private URI getRootURI() {
...
+	public String getCommandLine() {
+		if (lspStreamProvider instanceof ProcessInfo pi) {
+			return pi.getCommandLine();
+		}
+		return null;
+	}
+
+	@Nullable
+	public IProject getInitialProject() {
+		return initialProject;
```
mickaelistria, line 383: What about just making the final `initialProject` field public?

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServerWrapper.java
```diff
@@ -369,6 +370,27 @@ private URI getRootURI() {
 		return null;
 	}
 
+	@Nullable
+	public String getCommandLine() {
+		if (lspStreamProvider instanceof ProcessInfo pi) {
+			return pi.getCommandLine();
+		}
+		return null;
+	}
```
mickaelistria, line 379: Those utility method are not best in this class; have something that returns (or better, some adapter) to get the ProcessInfo (or better ProcessHandle) is enough. Then consumers can dereference it as needed.

mickaelistria (changes requested):
Additionally to the inlined comments, here are some suggestions:
* Use ProcessHandle that does everything well
* Use some adapterFactories to adapt the StreamProvider and LanguageServerWrapper to ProcessHandle; and consumers just call `Adapters.adapt(wrapper, ProcessHandle.class)`. This prevents API from extra interface hierarchy
* I don't think such view is useful for the average user. I would rather see it part of a new dedicated bundle (eg org.eclipse.lsp4e.sdk) that is dedicated to LS management, monitoring, debugging... in the IDE.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServerWrapper.java
sebthom: ok, I am addressing this separately with #755

mickaelistria:
Thanks!

rubenporras:
Thanks a lot, I am looking forward for this feature.

Issue #512, opened 2023-03-02
Multiple server instances where started for a document when singleton=false


ghentschke:
Description: 
Several language server instances gets started for a file occasionally. 

Steps to reproduce:
- Setting the `singleton` flag in the server definition of the `org.eclipse.lsp4e.languageServer` extension point to `false`
- Open a project file which triggers a LS  instance to gets started, then several LS instances gets started for this file.  

It looks like a race condition to me, because the (deprecated) `org.eclipse.lsp4e.LanguageServerWrapper.getInitializedServer()` method gets called from several threads an is not thread safe. 
This could be fixed by adding `synchronized` to the `getInitializedServer()`method.

All other methods calling `org.eclipse.lsp4e.LanguageServerWrapper.start()` are tagged  `synchronized`. The `start()` method itself is synchronized as well, but in this case this seems to be not enough.

This screenshot illustrates the problem (start() has been called twice, added a hit count = 2 to breakpoint in line 251): 
![image](https://user-images.githubusercontent.com/123444711/222447569-031efc9d-79ea-4fd7-ae5b-532c59fb7263.png)

ghentschke:
Unfortunately a single `synchronized` won't fix the issues I face when using multiple serer instances

ahmedneilhussain:
Hi @ghentschke I'm tied up with other stuff at the moment but would be keen to help get this resolved. The `start()` logic in `LanguageServerWrapper` is very complicated, which makes it very hard to reason about. I'd like to refactor and simplify this code so the lifecycle is clearer (I recently split out a load of straight-line stuff about the client supported features, but that was the easy bit!)

See https://github.com/eclipse/lsp4e/issues/415

ghentschke:
Yes, I've seen that there is work in progress. I'll take a closer look at the start() mechanism.

mickaelistria:
@ghentschke Do you think you could write an automated case that reproduces this issue?

ahmedneilhussain:
> Yes, I've seen that there is work in progress. I'll take a closer look at the start() mechanism.

Hi @ghentschke I've refactored the start() mechanism in #520  - mostly just rearranged the existing code to make the statefulness easier to reason about, but with a few small tweaks to help militate against race conditions.

[...]

PR #520, opened 2023-03-06
Maintenance: Tidy up server lifecycle

ahmedneilhussain:
Further tweaks to server lifecycle code to make it a little simpler to follow.

Avoid mutating initializeFuture from within blocks chained onto initializeFuture itself. There is no need to e.g. call initializeFuture.completeExceptionally from within an exceptionallyBlock chained onto initializeFuture! If that block is hit the future is already in a 'completed, exceptionally' state.

### DIFF
--- a/org.eclipse.lsp4e.tests.mock/src/org/eclipse/lsp4e/tests/mock/MockLanguageServer.java
+++ b/org.eclipse.lsp4e.tests.mock/src/org/eclipse/lsp4e/tests/mock/MockLanguageServer.java
@@ -118,8 +118,10 @@ public final class MockLanguageServer implements LanguageServer {
 		inFlight.forEach(future -> {
 			try {
 				future.join();
-			} catch (CancellationException | CompletionException e) {
-				System.err.println("Error waiting for in flight requests prior to teardown: " + e.getMessage());
+			} catch (CompletionException e) {
+				System.err.println("Error waiting for in flight requests prior to teardown: " + e);
+			} catch (CancellationException e) {
+				// Not a test error!
 			}
 		});
 	}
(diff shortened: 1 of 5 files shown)


ahmedneilhussain:
@rubenporras @mickaelistria any general thoughts? Need to be very careful making changes to this logic, so I definitely don't think we should be in a hurry to merge any of this. On the other hand, the existing code has evolved quite a lot to the point where it is very hard to reason about the lifecycle, so I think sooner or later it would be good to bite the bullet and try and rationalise things.

Would appreciate thoughts on this as a general approach.

ahmedneilhussain:
Just fixed an issue with workspace folders. Of interest because it is related to the discussion about lifecycle and `LanguageServerWrapper` reuse (or not) in https://github.com/eclipse/lsp4e/issues/512

Two observations come to mind: we seem to be interpreting the 'workspace folders' capability slightly differently to the spec. AIUI, workspace folders are supported at basic level either as a yes or a no, and that doesn't change dynamically. The feature that can be added dynamically is subscription to the `workspaceFolders/didChange` event. That's not the same thing as pure `workspaceFolders` support - the latter is just the server's ability to send a request to the client asking for a list of workspace folders.

Secondly - what are we gaining from keeping wrappers around when the server is shut down?

mickaelistria:
> we seem to be interpreting the 'workspace folders' capability slightly differently to the spec.

Can you please open a dedicated issue for that?

> Secondly - what are we gaining from keeping wrappers around when the server is shut down?

In case a LS instance or connection dies, then the wrapper is still supposed to be partly usable and to revive the LS. Usually, the LS keeps the same capabilities and so on, so we can keep using the wrapper to select the language servers that we will query for an operation or another even if this LS has died in the meantime.

mickaelistria:

PR #531, opened 2023-03-07
Quick Fixes might arrive as empty list

BoykoAlex:
Quick fixes for a problem marker shown fine if there are quick fixes available. However, if there are no quick fixes available an empty list of code actions is received on the lsp4e side. However, this result is ignored (this the chunk of code is changed by this PR) and therefore **Computing...** is shown indefnitely under quick fixes for problem that don't have a quick fix.

### DIFF
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServers.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServers.java
@@ -156,7 +156,7 @@ public abstract class LanguageServers<E extends LanguageServers<E>> {
 		CompletableFuture.allOf(
 				executeOnServers(fn)
 				.map(cf -> cf.thenApply(t -> {
-					if (!isEmpty(t)) { // TODO: Does this need to be a supplied function to handle all cases?
+					if (t != null) { // TODO: Does this need to be a supplied function to handle all cases?
 						result.complete(Optional.of(t));
 					}
 					return t;


rubenporras:
@ahmedneilhussain ,  what do you think of this?

BoykoAlex:
I see there is a unit test failure... looks like this was meant to be work for empty lists in the case of a hover. Anyway, for the case of quick fixes this doesn't work. I can suggest `computeFirstNonNull` perhaps...

rubenporras:
@BoykoAlex, what about following the TODO on the line that you have modified 

> TODO: Does this need to be a supplied function to handle all cases?

and adding a parameter `Function<T, Boolean> isValidResult`?

Consumers could do pass `t != null && t instanceof List l && l.isEmpty()` to keep the current logic or t != null in the case of quickfixes.

Would you like to do that or shall I?

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServers.java
```diff
@@ -156,7 +156,7 @@ public <T> CompletableFuture<Optional<T>> computeFirst(BiFunction<? super Langua
 		CompletableFuture.allOf(
 				executeOnServers(fn)
 				.map(cf -> cf.thenApply(t -> {
-					if (!isEmpty(t)) { // TODO: Does this need to be a supplied function to handle all cases?
```
rubenporras, line 159: isEmpty is now not used, so we could remove it from the class

mickaelistria:
> and adding a parameter Function<T, Boolean> isValidResult?

If we can avoid extra parameters, it would be great. Everything is possible with new parameters, new methods and so on, but in the end they make the API more and more complicated. We should aim at getting a minimal API that makes the best choice.
Here, I believe the issue is the semantics of `computeFirst` which does some filtering. In case the expected future returns is a List, the default value in case nothing is returned should probably be an emptyList list.

Something like
```java
		AtomicBoolean metEmptyList = new AtomicBoolean();
		CompletableFuture.allOf(
				executeOnServers(fn)
				.map(cf -> cf.thenApply(t -> {
					if (t instanceof List<?> && t.isEmpty()) {
						metEmptyList.set(true);
					}
					if (!isEmpty(t)) {
						result.complete(Optional.of(t)); // return immediately non-empty result
					}
					return t;
				})).toArray(CompletableFuture[]::new)
				).whenComplete((v, t) -> completeEmptyOrWithException(result, t));

		return result.thenApplyAsync(result -> (result == null && metEmptyList) ? List.of() : result);
```

rubenporras:
@mickaelistria , is your proposal not the same as the one done by @BoykoAlex? Both cases return the empty list or am I missing something. If that is the goal, I think the proposal done by alex is preferable.

In general it is a bit strange in my opinion that the class has specific code for a List, but not for a Map, a Set, and it is I think not really something that can be achieved in general, because we cannot have filter functions for a type which can be anything. So I think to not have expectations that the method cannot really fulfill, we should just check for != null as in the proposed PR or give a filter function.

One could also offer a filter function and create a method that defaults to just != null, thus having a simple API for most cases but allowing custom filtering.

ahmedneilhussain:
> @mickaelistria , is your proposal not the same as the one done by @BoykoAlex? Both cases return the empty list or am I missing something. If that is the goal, I think the proposal done by alex is preferable.
> 
> In general it is a bit strange in my opinion that the class has specific code for a List, but not for a Map, a Set, and it is I think not really something that can be achieved in general, because we cannot have filter functions for a type which can be anything. So I think to not have expectations that the method cannot really fulfill, we should just check for != null as in the proposed PR or give a filter function.
> 
> One could also offer a filter function and create a method that defaults to just != null, thus having a simple API for most cases but allowing custom filtering.

Hi @rubenporras the reason for the test for empty list is because of the way computeFirst() works (or is intended to work). This method is for when we have multiple language servers but it's only sensible to use one response (e.g. an edit). However we want the first LS that produces a *result* not the first LS to *respond*, otherwise a LS that doesn't have anything 'interesting' to contribute will win out just because it responds first, and the user will see nothing when they actually could have had a useful result from a different LS.

The bug looks to be with the fallback code: if *none* of the LS respond with anything 'interesting' then the future should still terminate. I think this should be happening correctly (there is a test case `LanguageServersTest.testComputeFirstReturnsEmptyOptionalIfNoResult`). But in this case the future will return with an empty optional.

I think the problem is not in `LanguageServers.computeFirst()` in this case, I think it is in the consuming domain code:

see https://github.com/eclipse/lsp4e/blob/da05b55a9f15d48669ee05dc531589d88cbb7b53/org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/codeactions/LSPCodeActionMarkerResolution.java#L135

I think the bug is that this block should have a branch for optional.ifEmpty() that replaces the 'computing' placeholder with nothing. @BoykoAlex what do you think?

@rubenporras the reason we have a (slightly hacky) `instanceof` test for `List` and `empty` is that some LS methods return null objects when they have nothing to report, and some return an empty List. I don't think that maps and sets are really primitive LSP4j types that we have to worry about. I think checking just for a null object or an empty list already handles the likely cases. I did wonder about making the predicate injectable as you see from the comments but I think it isn't likely to be a problem in practice, and it isn't really the problem in this case.

Does that make sense, everyone?

ahmedneilhussain:
I'd also add that in this case is there any reason not to use `collectAll()` rather than `computeFirst()`? There's no harm in showing the user fixes from multiple language servers rather than just those from the first useful responder? I suppose `computeFirst()` might respond a bit quicker...

rubenporras:
@ahmedneilhussain , so you propose a fix like https://github.com/eclipse/lsp4e/pull/535?

mickaelistria:
> Both cases return the empty list or am I missing something. If that is the goal, I think the proposal done by alex is preferable

In the current state
  ** Pro: if the first LS returns and empty list/nothing interesting and second return some content in a 2nd list, then the 2nd list is returned
  ** Con: if all LS return an empty list, null is returned
In @BoykoAlex 's propsal
  ** Pro: if all LS return an empty list, an empty list is returned
  ** Con: if empty list/nothing interesting is returned by first LS, then the empty list is returned and other LS are not queried
In my proposal: Both pros above, none of the cons.

ahmedneilhussain:
> @ahmedneilhussain , so you propose a fix like #535?

Yes @rubenporras , exactly that. @mickaelistria something like your proposal would undoubtedly work, but would it really be useful? Is it at all likely that calling code would need to know the distinction between a null result and an empty list? Either one or more servers returned some content, in which case the calling code can process it, or they didn't, in which case the caller either has nothing to do or (as in this case) just has to update a 'pending' placeholder with 'nothing'. `Optional` captures that duality well, which is why I went for it: is there a real need to distinguish different flavours of 'nothing'?

PR #534, opened 2023-03-08
Avoid Thread Access Exception

rubenporras:
@mickaelistria , apparently by moving the code the execution time was shifted and we run into a thread access exception. Do you think this fix is fine?

PD: It works in the test I did.


rubenporras:
This is for sure better than the current code, so I will merge it. @mickaelistria if you think using an static block like this is not going to work, I will rework it.

rubenporras:
I actually wonder if having the background color is just not worth the trouble, why not just having the crossout and leaving the background color as is? That would be dead simple :)

If I do not here anything back, I will try to follow your suggest and will come back with more questions or with a PR.

mickaelistria:
> why not just having the crossout and leaving the background color as is? That would be dead simple :)

Let's go for it ;) Colors are usually not the best way to emphasize on some semantics, a crossout would be much clearer.

PR #535, opened 2023-03-08
Handle empty quickfix list

rubenporras:
Quick fixes for a problem marker shown fine if there are quick fixes available. However, if there are no quick fixes available an empty list of code actions is received on the lsp4e side. However, this result is ignored (this the chunk of code is changed by this PR) and therefore Computing... is shown indefnitely under quick fixes for problem that don't have a quick fix.


ahmedneilhussain:
Looks good to me, @rubenporras 👍

mickaelistria:
@BoykoAlex Does it fix issue #531 ?

rubenporras:
Okay, I will wait for his feedback before merging.

PR #539, opened 2023-03-08
Handle all collections and not only lists

rubenporras:


rubenporras:
@ahmedneilhussain , no harm in handling other collections than lists, right?

ahmedneilhussain:
> @ahmedneilhussain , no harm in handling other collections than lists, right?

@rubenporras none (except to me when I have to backport it to Java 8 when I update Cocotec's private fork with the latest changes from lps4e-public ;-) )

rubenporras:
Thanks

PR #540, opened 2023-03-08
New API: Document statefulness of executors and prevent reuse

ahmedneilhussain:
@rubenporras this is the small change to make statefulness of the executors clear and force people to create new executors rather than reuse them, as we discussed in https://github.com/eclipse/lsp4e/discussions/523#discussioncomment-5232768

I haven't gone through the code to check we aren't doing this yet. It would have only arisen from the porting work that you and I did, and I don't think in the end we ended up using executors as e.g. fields rather than as transient objects created for the lifetime of a single method call. However I don't think our unit tests are extensive enough to catch such things, so I should probably just look through the code quickly even if the tests pass.

### DIFF
--- a/org.eclipse.lsp4e.test/src/org/eclipse/lsp4e/test/LanguageServersTest.java
+++ b/org.eclipse.lsp4e.test/src/org/eclipse/lsp4e/test/LanguageServersTest.java
@@ -116,6 +116,49 @@ public class LanguageServersTest {
 		assertTrue(hovers.contains("HoverContent1"));
 		assertTrue(hovers.contains("HoverContent2"));
 	}
+	
+	@Test(expected=IllegalStateException.class)
+	public void testExecutorNotReusable() throws Exception {
+		final AtomicInteger hoverCount = new AtomicInteger();
+		MockLanguageServer.INSTANCE.setTextDocumentService(new MockTextDocumentService(MockLanguageServer.INSTANCE::buildMaybeDelayedFuture) {
+			@Override
+			public synchronized void didChange(DidChangeTextDocumentParams params) {
+				super.didChange(params);
+			}
+
+			@Override
+			public synchronized CompletableFuture<Hover> hover(HoverParams position) {
+				Hover hoverResponse = new Hover(Collections.singletonList(Either.forLeft("HoverContent" + hoverCount.incrementAndGet())), new Range(new Position(0,  0), new Position(0, 10)));
+				return CompletableFuture.completedFuture(hoverResponse);
+			}
+		});
+
+		IFile testFile = TestUtils.createUniqueTestFileMultiLS(project, "Here is some content");
+		IEditorPart editor = TestUtils.openEditor(testFile);
+		ITextViewer viewer = LSPEclipseUtils.getTextViewer(editor);
+		final IDocument document = viewer.getDocument();
+
+		final HoverParams params = new HoverParams();
+		final Position position = new Position();
+		position.setCharacter(10);
+		position.setLine(0);
+		params.setPosition(position);
+		
+		final LanguageServerDocumentExecutor ex = LanguageServers.forDocument(document)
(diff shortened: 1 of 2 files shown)


rubenporras:
The following classes keep an executor in a field:

- LSPCodeActionsMenu.java
- LSSearchQuery.java
- LSPSymbolInWorkspaceDialog.java

I am not sure if it is reused or only saved for only one execution, but probably we should rewrite them to be sure

rubenporras:
See https://github.com/eclipse/lsp4e/pull/541

mickaelistria:
I'm not found of the idea of making executors not reusable and mandating strong usage constraints ("do not keep references") on consumer 
Couldn't we allow user to keep a reference but allowing the executor to refresh internally when necessary, or even to just re-create itself from the filter/document if it is "dirty"? Isn't it "just" a matter of dropping the document version after 1 call, or of refreshing the document version on next call?

rubenporras:
I agree that if we can make the executors be reusable, that is preferable

ahmedneilhussain:
Well, when I did the original implementation I had `reset()` and `clone()` methods to mitigate this: maybe both is overkill but one of the two would be OK? Reset is certainly a trivial implementation and clone only a few lines more.

rubenporras:
But `reset()` or `clone()` still would need to call them manually. That is error prone, and if you know how and when to do that, you can as well create a new executor, which is simple. I think the ideal situation is when it would work seamlessly, I understand that is what Mickael meant with "refresh internally when necessary, or even to just re-create itself".

mickaelistria:
It seems to me the transient state of the executor only matters in the case `toVersionedEdits` is used, isn't it? If so, could we just put the guard only in toVersionedEdits for the moment?

ahmedneilhussain:
> The problem is rather `LanguageServerDocumentExecutor.getStartVersion()`, which is used by `toVersionedEdits` and also by VersionedSemanticTokens.toVersionedSemantikTokens(LanguageServerDocumentExecutor, SemanticTokens, SemanticTokensLegend). I agree, putting the guard in `getStartVersion` would be sufficient.

OK so the user can reuse executors if they're not using the versioning support?

ahmedneilhussain:
> Another option (which would be my preferred option is that the executor does not keep the state), the only thing that we need to call to get the state is `DocumentUtil.getDocumentModificationStamp(document)`, which is easy to call just before `compute/collect` on the executor and then can be used inside the lambdas in the `CompletableFuture`;
> 
> I can do a draft PR if it helps the discussion to compare both approaches

Yeah, I guess this would amount to moving this responsibility out into a different `Versioner` class, a bit like cancellation support. I quite like the built-in versioning support in the executor API but, given you need to store the executor in a variable to be able to invoke `toVersionedEdits()` having a separate object wouldn't be any more verbose.

So I guess it would like like:

```
var v = new Versioner(document);
LanguageServers.forDocument(document).withFilter(someFilter)
  .computeFirst(ls ->ls.getTextDocumentService().format(params).thenApply(v::toVersionedEdits)) ...
```

rather than
```
var ex = LanguageServers.forDocument(document).withFilter(someFilter);
ex.computeFirst(ls ->ls.getTextDocumentService().format(params).thenApply(v::toVersionedEdits)) ...
```

so not a lot in it.

rubenporras:
@mickaelistria , following your suggestion, I have updated https://github.com/eclipse/lsp4e/pull/542 with a builder for versioned elements, so that the apply can be written as `.thenApply(builder::build)`. Is https://github.com/eclipse/lsp4e/pull/542 what you would expect?

rubenporras:
@ahmedneilhussain , I think we can close this now for the time being.

ahmedneilhussain:
> @ahmedneilhussain , I think we can close this now for the time being.

Yes, sorry @rubenporras this is superfluous now you've moved versioning out of the executors completely. Apologies for not responding sooner - was on leave on Friday.

Closing this as superseded.

PR #542, opened 2023-03-09
Remove state from language server document executor

rubenporras:


org.eclipse.lsp4e/src/org/eclipse/lsp4e/VersionedEdits.java
```diff
@@ -30,9 +29,9 @@ public class VersionedEdits extends Versioned<List<? extends TextEdit>> {
 
 	private final IDocument document;
 
-	public VersionedEdits(long version, List<? extends TextEdit> data, IDocument document) {
-		super(version, data);
-		this.document = document;
+	public VersionedEdits(Pair<IDocument, Long> data, List<? extends TextEdit> textEdits) {
```
mickaelistria, line 32: Why a Pair here? I don't think it makes API cleaner.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/format/LSPFormatter.java
```diff
@@ -54,17 +57,17 @@ public CompletableFuture<Optional<VersionedEdits>> requestFormatting(@NonNull ID
 		// TODO: Could refine this algorithm: at present this grabs the first non-null response but the most functional
 		// implementation (if a text selection is present) would try all the servers in turn to see if they supported
 		// range formatting, falling back to a full format if unavailable
+		var builder = new VersionedBuilder<List<? extends TextEdit>, @NonNull VersionedEdits>(document, VersionedEdits::new);
```
mickaelistria, line 60: What a builder and not directly a VersionedEdit ?

mickaelistria (commented):
I think it's more or less the same idea; I just have some questions inline to better understand this proposal.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/VersionedEdits.java
rubenporras: Yes, it is does not really captures the semantic. I used a Pair as a quick replacement of a TriFunction, since the apply method needs three input arguments (document, timestamp, <request>). I have not replaced it by a Versioned<Document> with conveys the meaning better.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/format/LSPFormatter.java
rubenporras: I think it fits nicely our problem:

- We need to record the document state right before sending the request
- That recorded value should be added to the returned data to create a Versioned<T> than then you can use

Having the builder clearly makes this visible in the code. We could of course create already a Versioned<T> that we then later complete, and then gets usable, but it does not enforce that the pattern is followed.

rubenporras:
@mickaelistria , this passes now the tests. I am happy with the new API as it removes effectively the state in `LanguageServers` and the consumers that actually check versions are not harder to write/understand than before, what do you think?

If we do not want to have the lines `var builder = new VersionedBuilder<List<? extends TextEdit>, @NonNull VersionedEdits>(document, VersionedEdits::new);` and `	var builder = new VersionedBuilder<Pair<SemanticTokens, SemanticTokensLegend>, @NonNull VersionedSemanticTokens>(document, VersionedSemanticTokens::new);` we could create static factory methods in `VersionedBuilder`  but I think it is not worth to have a single entry point, as we probably use those builders at exactly one place.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/format/LSPFormatter.java
mickaelistria: > We need to record the document state right before sending the request

That can be achieved by directly instantiating a VersionedEdits here (Versioned should probably contain the document as field by the way)

> That recorded value should be added to the returned data to create a Versioned than then you can use

Yes, but something like `versionedEdits.apply` could do it (`apply` being `public void apply(List<TextEdit>>)` which would check the doc version before performing the actual change.

So code would look like
```java
VersionedEdits edits = new VersionedEdits(document);
return executor.computeFirst((w, ls) -> {
		final ServerCapabilities capabilities = w.getServerCapabilities();
		if (isDocumentRangeFormattingSupported(capabilities)
				&& !(isDocumentFormattingSupported(capabilities)
						&& textSelection.getLength() == 0)) {
			return ls.getTextDocumentService().rangeFormatting(rangeParams).thenApply(builder::build);
		}
		return ls.getTextDocumentService().formatting(params).thenApply(edits::apply);
});
```
This has the advantage of avoid intermediary object, IMO making the code clearer.

mickaelistria: > But what if you would use computeAll?

Indeed, this case would require multiple VersionedEdits I guess. However, do we already have this case yet? I'd like to avoid introducing one solution before the problem exists ;)

> `VersionedEdits edits = new VersionedEdits(document);` would also be an extra object as in the builder.

Yes, but I think most people are more comfortable with `new` over `Builder` unless it's proven a Builder makes things simpler. 
```java
VersionedEdits edits = new VersionedEdits(document);
edits.apply(textEdits);
// or
edits.addEdits(textEdits);
edits.apply();
```
seems to cover the current need (assuming we don't need it with computeAll at the moment, which may be a wrong assumption), with just 1 method addition that is relatively straightforward to guess with completion.

rubenporras: Agree, if the builder does not solve a problem, and most people are indeed more confortable with `new` over `Builder`, that is fine for me, it was indeed my proposal in the first commit https://github.com/eclipse/lsp4e/pull/542/commits/2f4924315972e64db99e8df4be623ca8e24fd255
Just 
`		long modificationStamp = DocumentUtil.getDocumentModificationStamp(document);`
to record the timestamp and then
`edits -> new VersionedEdits(modificationStamp, edits, document)`
inside the lamba. No new pattern and it is straghforward. You need to pass three arguments instead of one.

Shall we go for that one?

Note that as I tried to explain above, even if computeFirst returns only the first result, it does not guarantee that only one lambda is executed, so
```
VersionedEdits edits = new VersionedEdits(document);
return executor.computeFirst((w, ls) -> {
                ...
		return ls.getTextDocumentService().formatting(params).thenApply(edits::apply);
});
```
needs `edits::apply` to be thread safe, and even then the versioned edit could contain edits from more than one server.

rubenporras:
I have rebased the original PR and submitted it as https://github.com/eclipse/lsp4e/pull/546. The approach does not contain the builder, but it still solves the race conditions for `computeFirst`/`computeAll`.... It looks quite trivial so I think we can go for it.

mickaelistria:
So should we close this on now #546 is merged?

PR #544, opened 2023-03-10
Some factorization for Versioned

mickaelistria:


org.eclipse.lsp4e/src/org/eclipse/lsp4e/VersionedEdits.java
```diff
@@ -44,14 +41,19 @@ public VersionedEdits(long version, List<? extends TextEdit> data, IDocument doc
...
-		if (getVersion() != DocumentUtil.getDocumentModificationStamp(this.document)) {
+		if (this.sourceDocumentVersion != DocumentUtil.getDocumentModificationStamp(this.document)) {
 			throw new ConcurrentModificationException();
 		} else {
-			LSPEclipseUtils.applyEdits(this.document, get());
+			LSPEclipseUtils.applyEdits(this.document, data);
 		}
 	}
 
 	public static @NonNull VersionedEdits toVersionedEdits(@NonNull LanguageServerDocumentExecutor executor, List<? extends TextEdit> data) {
```
rubenporras, line 51: we could now remove this factory method based on the executor, right?

org.eclipse.lsp4e/src/org/eclipse/lsp4e/Versioned.java
```diff
@@ -12,32 +12,35 @@
...
+	/**
+	 * Constructs a new Versioned for the given document, and specify source version
+	 */
+	public Versioned(final IDocument document, long sourceDocumentVersion, final T data) {
+		this.document = document;
+		this.sourceDocumentVersion = sourceDocumentVersion;
 		this.data = data;
 	}
 
-	public T get() {
```
rubenporras, line 31: I would rather keep the getters instead and make them protected than making the fields protected and removing the methods, but since the fields are final, it does not matter much in the end.

rubenporras (commented):
Yes, moving the document into the `Versioned` class makes sense. I have only two minor comments

org.eclipse.lsp4e/src/org/eclipse/lsp4e/VersionedEdits.java
mickaelistria: Yes, done.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/Versioned.java
mickaelistria: I think "Effective Java" recommends usage of public (or protected here is enough) final fields for non-API objects. Adding getters and setters can make sense when things become more complex, but usually directly accessing final fields is pretty comfortable to uses and allows the JVM to run much more optimization.

rubenporras (approved):

org.eclipse.lsp4e/src/org/eclipse/lsp4e/Versioned.java
rubenporras: Thanks, I will check that. I checked the record classes in Java 17 in case we could use them, but we cannot because they are final.

PR #546, opened 2023-03-12
Remove state from LanguageServerDocumentExecutor

rubenporras:


mickaelistria (approved):
I believe there is still some room for concurrently issues to jump in (eg between the storage of documentVersion and actual request), but the cases would be relatively rare and more importantly they were actually still possible with previous API (document could change between computeVersion and request actually sent); so I see this as a good simplification worth being merged.

rubenporras:
> I believe there is still some room for concurrently issues to jump in (eg between the storage of documentVersion and actual request)

Yes, you are right. I did not think about that.

ahmedneilhussain:
> > I believe there is still some room for concurrently issues to jump in (eg between the storage of documentVersion and actual request)
> 
> Yes, you are right. I did not think about that.

@rubenporras @mickaelistria technically, yes, but that was just as true when the document version was read within the `computeAll()` etc methods, and in practice it can't be avoided without locking the document in some way I think.

The thing that makes it OK is that the executors will ordinarily be called on the SWT thread which is where document mutation should occur. So a call to `LSPFormatter.requestFormatting()` might be made on the SWT thread with some keystroke enqueued that might then change the document, the handler for that keystroke would not get processed until `requestFormatting()` had passed control back to the SWT event loop.

Sorry for not responding sooner on this stuff - I was on leave on Friday and was a bit tied up with other stuff today.

rubenporras:
Yes, for the formatter it is probably fine, but I am do not think it is fine for the semantic highlight reconciler. Probably, as you say, we need to lock the document. It is something I plan to look at, and I would like to fix it before the newer LSP4E release if feasible

rubenporras:
I have checked and `DocumentUtil.getDocumentModificationStamp` will lock the document if it is a `IDocumentExtension4` and if it is `SynchronizableDocument` which can be locked, it will be locked, so I think we are fine.

PR #551, opened 2023-03-15
Enable LSPSymbolInWorkspaceHandler only if LSP4E bundle is active

### LINKED ISSUE #38: Conflicting keybinding for Open Type

oleosterhagen:
Fixes #38

`LSPSymbolInWorkspaceHandler.setEnabled(Object)` for updating the enable status will only be called by the framework when the bundle org.eclipse.lsp4e is active.

Without the now introduced check for `bundleState=ACTIVE` the conflicting key binding will not be filtered out by method `KeyBindingDispatcher.getExecutableMatches(KeySequence, IEclipseContext)` in bundle `org.eclipse.e4.ui.bindings`.

Before Eclipse Release 2022-12 the conflicting key binding for `Shift+Ctrl+T` didn't show up, because the bundle `org.eclipse.lsp4e` was always activated by a startup extension in `plugin.xml` and that is why conflict resolution in the `KeyBindingDispatcher` has worked. The use of this extension has been removed with a0dd75aeb405986e187c9a2b1325144e2afa9273. (Bundle activation status can easily checked with `ss lsp4e` in the _Host OSGi Console_.)

Note: This fix will not prevent a key binding conflict for `Shift+Ctrl+T` when LSPSymbolInWorkspaceHandler is enabled (e.g. in the Generic Editor for JavaScript). The conflict with `Shift+Alt+R` also needs a different solution.

### DIFF
--- a/org.eclipse.lsp4e/plugin.xml
+++ b/org.eclipse.lsp4e/plugin.xml
@@ -131,6 +131,11 @@
       <handler
             class="org.eclipse.lsp4e.operations.symbols.LSPSymbolInWorkspaceHandler"
             commandId="org.eclipse.lsp4e.symbolinworkspace">
+         <activeWhen>
+            <reference
+                  definitionId="org.eclipse.lsp4e.textSelectionHasLanguageServer">
+            </reference>
+         </activeWhen>
       </handler>
       <handler
             commandId="org.eclipse.lsp4e.togglelinkwitheditor">


mickaelistria (commented):
Have you verified if #38 is still present with latest snapshots? I'm not sure it is.
The bundle state is not a good enough reason to decide whether a bundle is enabled or not; if LSP4E is active then the handler shouldn't be active unless editor content is LSP-enabled. The enablement of the handler is already properly controlled in the handler implementation class.

oleosterhagen:
Unfortunately https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/476 narrows only the applicability of the handler from JDT when actually in this case the handler from LSP4E have to be filtered out.

The filtering in [KeyBindingDispatcher](https://git.eclipse.org/r/plugins/gitiles/platform/eclipse.platform.ui/+/refs/heads/master/bundles/org.eclipse.e4.ui.bindings/src/org/eclipse/e4/ui/bindings/keys/KeyBindingDispatcher.java#548) does not work anymore because since Eclipse 2022-12 the LSP4E bundle is not always loaded on startup (see https://github.com/eclipse/lsp4e/commit/a0dd75aeb405986e187c9a2b1325144e2afa9273). 

But without activating the bundle, the programmatically defined enablement check in `LSPSymbolInWorkspaceHandler` cannot be invoked and the Eclipse framework decides the handler is _active_. This is a performance optimization to prevent loading all bundles eagerly when filling menus etc.

Note: While testing/debugging one has to be careful not to open the `org.eclipse.lsp4e.test` project, because this activates the bundle `org.eclipse.lsp4e` and it seems as everything works already without any patches.

mickaelistria:
OK, thanks for the details. I was wrongly assuming that `canExecute` would return false is the handler comes from a non-active bundle; from there most of my reasoning got wrong.
What about using the same activation/enablement condition as other handlers, ie
```xml
         <activeWhen>
            <reference definitionId="org.eclipse.lsp4e.textSelectionHasLanguageServer" />
         </activeWhen>
```
?
That would indeed load LSP4E on startup, but it would make the activation here more correct.

mickaelistria (approved):

oleosterhagen:
Thanks, I have adjusted and tested the change according to your suggestion.

mickaelistria:
Thank you!

de-jcup:
@oleosterhagen - Thank you for fixing this. :rocket:

PR #556, opened 2023-03-17
Remove most of the deprecated for removal methods. Some are still used

rubenporras:
in tests and will be dealt with in a follow-up.

### DIFF
--- a/org.eclipse.lsp4e.test/src/org/eclipse/lsp4e/test/LanguageServiceAccessorTest.java
+++ b/org.eclipse.lsp4e.test/src/org/eclipse/lsp4e/test/LanguageServiceAccessorTest.java
@@ -17,7 +17,6 @@ import static org.eclipse.lsp4e.LSPEclipseUtils.getDocument;
 import static org.eclipse.lsp4e.LSPEclipseUtils.getTextViewer;
 import static org.eclipse.lsp4e.LanguageServiceAccessor.getActiveLanguageServers;
 import static org.eclipse.lsp4e.LanguageServiceAccessor.getInitializedLanguageServers;
-import static org.eclipse.lsp4e.LanguageServiceAccessor.getLSPDocumentInfosFor;
 import static org.eclipse.lsp4e.LanguageServiceAccessor.getLSWrapper;
 import static org.eclipse.lsp4e.LanguageServiceAccessor.getLSWrappers;
 import static org.eclipse.lsp4e.LanguageServiceAccessor.getLanguageServers;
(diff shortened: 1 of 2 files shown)


mickaelistria:
I'm all for removing deprecated stuff; but I'm aware of some integrations here and there that may need a heads up: @sebthom @martinlippert @BoykoAlex @sbouchet ... FYI.

sebthom:
Thanks for the heads up. I am going to check if I face any issues in haxe4 or dart4e and report back.

sebthom:
I would appreciate if we could keep the `LSPDocumentInfos` class and the `LanguageServerAccessor.getLSPDocumentInfosFor` method a bit longer. Otherwise I would instantly have to drop support for all older Eclipse versions and I would like to avoid having to maintain two separate development branches.

mickaelistria:
> I would have to instantly drop support for all older Eclipse versions

Wouldn't it be a good thing? The more maintainers do care about supporting older versions; the less users are likely to upgrade to newer versions; and the less they're connected with the current state of the project.
But apart of this general reasoning, if it's profitable enough to you (and only if it is ;), I think it's fair to keep those a bit longer to facilitate how much you can profit from LSP4E.

sebthom:
I am mainly concerned about regression bugs in the Eclipse platform itself. For example in Eclipse 2022-12 the revision compare editor was completely broken if you tried to synchronize changes. It is better in 2023-03 but I am still see strange rendering/refresh issues. I would like to give users the ability to go back one or two Eclipse versions until such bugs (hopefully) get fixed. Since the there are no more patches/hot fixes released, this means waiting for the next quarterly release.

rubenporras:
I have no problem into keeping `LSPDocumentInfos` around, but since we have code in the latest released LPS4E which requires the SWT in 2023-03, I am not sure what you have in mind is actually possible. Is it?

sebthom:
Our plugins do not depend on the latest version of LSP4E but are compatible with it.

So user can use the latest version of our plugin with Eclipse 2022-09, 2022-12 and 2023-03 and the LSP4E version from the respective release train.

BoykoAlex:
Anything you can recommend to use instead of `LSPDocumentInfo#getVersion()`?

rubenporras:
> Anything you can recommend to use instead of `LSPDocumentInfo#getVersion()`?

Hi @BoykoAlex , The usual answer would be to use `LanguageServers`, but I would need to take a look at your code to tell if that is appropriate or if we missed something. Do you have a link to codebase or is it closed source?

BoykoAlex:
@rubenporras I noticed that removal of `LSPDocumentInfo` class is not part of this PR anymore. Thus we might be off the hook for now.

 The class `LSPDocumentInfo` has `getVersion()` method that i call from here: https://github.com/spring-projects/sts4/blob/main/eclipse-language-servers/org.springframework.tooling.ls.eclipse.commons/src/org/springframework/tooling/ls/eclipse/commons/STS4LanguageClientImpl.java#L212
The call `Utils.isProperDocumentIdFor(doc, highlightParams.getDoc())` checks if version of the current doc is the same as cached highlight/codelenses doc id.
Everythime viewer is refreshed to update visual annotations or codelenses the `updateSourceViewer(IEditorPart editor)` method is called.

rubenporras:
@mickaelistria , okay if I resolve the conflicts and merge this now? The PR has been here for a while and we did also a release since we set these methods for removal, so for me, now it would be good time to merge.

mickaelistria:
@rubenporras OK with me.

rubenporras:
I will start first with https://github.com/eclipse/lsp4e/pull/582, it remove less methods.

rubenporras:
I never got around finishing the PR, I think most of the old API has been removed, as what remains is not really a maintenance problem. I will close it to reduce the work that we do on each PR that is merged.

Issue #555, opened 2023-03-17
Add hyperlinked documents to origin LS


ghentschke:
When opening a linked external file/document, this file should be connected to the language server of the origin document.

This ensures that convenient language server(s) will be used for the linked file. This becomes important when using one LS per project in the Eclipse workspace. 

Background: Applicable when using a non singleton language server which does not support multi projects / workspace folder capability.

In the current implementation a new wrapper/server will be started for each, via hyperlink opened, external file. This lead to the following problems:
- The LS cannot find a compilation database for external files
- Excessive usage of LS when many external files are opened.

I can provide a PR for this issue.

mickaelistria:
> non singleton language server which does not support multi projects / workspace folder capability.

So how do we know such language server is able to process random files? I'm not sure we can reliably generalize such decision to use the same instance for opened files; this is deeply specific to 1 LS or the other.

ghentschke:
I don't think it's for random files. The adding is related only to external files with the same content type as the LS supports.  The use case is for C/C++ very common: to include standard library header files form a C/C++ source or header. 

The problem is how to tell the LS which compilation database it should use for external source code files. The usage depends on the origin file. The compiler settings used to compile the source file should be used for the external file as well. That's why this change is so important.

ghentschke:
> Most language server fix that by either being singleton or enabling workspace folders support

I'll check that. Thank you.

One more question. What is the desired behavior for hyperlinked files of LSP4E when the LS is singleton and the LS does not support workspace folders regarding LS/wrapper? In the current implementation a new LS/wrapper will be started for each opened file.

mickaelistria:
> What is the desired behavior for hyperlinked files of LSP4E when the LS is singleton and the LS does not support workspace folders regarding LS/wrapper?

Initially, LSP4E was not using workspace folders, and the expected behavior was that only one instance of a wrapper/or a LanguageServer gets started. So under this expectation, the behavior you describe seems more like a bug.

ghentschke:
@mickaelistria Sorry, I forget the NOT: 
> What is the desired behavior for hyperlinked files of LSP4E when the LS is NOT a singleton and the LS does not support workspace folders regarding LS/wrapper? 

Currently a new LS will be started for each external file when the server is NOT a singleton. As a user I cannot change this behavior.

>> This case seems relatively specific to your particular LS integration; I suspect it would better fit in the plugin that provides the LS integration than into LSP4E commons. 

[...]

PR #558, opened 2023-03-17
Add hyperlink documents to origin LS (#555)

ghentschke:
This is first draft.


ghentschke:
I'll try to fix it in the LS

PR #559, opened 2023-03-18
New Rename command to resolve a key binding conflict for Shift+Alt+R

oleosterhagen:
From the discussion of: eclipse-jdt/eclipse.jdt.ui/pull/443

The key binding `Shift+Alt+R` should not be added to the existing command `org.eclipse.ui.edit.rename` (with existing key binding `F2`), because this leads to an ambiguity when using `Shift+Alt+R` in the **Package Explorer** from JDT UI. `Shift+Alt+R` would be associated with two commands:

  * `org.eclipse.ui.edit.rename` (Rename)
  * `org.eclipse.jdt.ui.edit.text.java.rename.element` (Rename - Refactoring)

Introducing an own command ID for rename in LSP4E and adding the key binding `Shift+Alt+R` only to this new command resolves this ambiguity. The new command is only available in the LSP4E Editor because its handler already has an appropriate activeWhen condition.

### DIFF
--- a/org.eclipse.lsp4e.test/src/org/eclipse/lsp4e/test/TestUtils.java
+++ b/org.eclipse.lsp4e.test/src/org/eclipse/lsp4e/test/TestUtils.java
@@ -79,7 +79,7 @@ public class TestUtils {
 		IEditorInput input = new FileEditorInput(file);
 
 		IEditorPart part = page.openEditor(input, "org.eclipse.ui.genericeditor.GenericEditor", false);
-		part.setFocus();
+		page.activate(part);
 		return part;
 	}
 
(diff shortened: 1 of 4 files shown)


mickaelistria:
If JDT wants to enforce 1 particular command to be used in its partcular context (of Package Explorer here), then JDT should consider defining the command in this context to prevent external plugins such as LSP4E fro participating.
Multiplying commands in the IDE is an anti-pattern (it's actually the root cause here that there are multiple rename commands when there could be only one); I don't think LSP4E should participate to increase the problem by defining yet-another-command.

oleosterhagen:
@mickaelistria you are completely right.

For me, a very good **first time user experience** is important. The current situation with the popup window is not that nice. (In my daily work I got used to using `F2` in the package explorer for renaming classes, which resolves to a single command.)

I hope we could then resolve the problem in JDT UI.

oleosterhagen:
Limiting the applicability of the key binding `Shift+Alt+R` in LSP4E would also solve the conflict with JDT UI. (The `LSPRenameHandler` is already only enabled for the text editor.)  I have created a new PR: #564

PR #568, opened 2023-03-23
Replace usage of rule by cancellation

mickaelistria:
Usage of resource rule seems to cause unnecessary delays; attempt to improve this by cancelling outdated jobs instead of using a rule.


mickaelistria:
The new change simply removes the workspace rule.

PR #575, opened 2023-03-28
Prevent code actions on read only sources.

andrewL-avlq:
Code actions cannot be performed on read only sources, so there is no need to ask the language server for them.

### DIFF
--- a/org.eclipse.lsp4e.test/src/org/eclipse/lsp4e/test/codeactions/CodeActionTests.java
+++ b/org.eclipse.lsp4e.test/src/org/eclipse/lsp4e/test/codeactions/CodeActionTests.java
@@ -26,6 +26,7 @@ import org.eclipse.core.resources.IFile;
 import org.eclipse.core.resources.IMarker;
 import org.eclipse.core.resources.IProject;
 import org.eclipse.core.resources.IResource;
+import org.eclipse.core.resources.ResourceAttributes;
 import org.eclipse.core.runtime.CoreException;
 import org.eclipse.core.runtime.NullProgressMonitor;
 import org.eclipse.lsp4e.LanguageServerPlugin;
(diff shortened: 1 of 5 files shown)


mickaelistria (changes requested):
The LSP spec mentions that a codeAction can return a WorkspaceEdit or a command, and those can affect other files or just affect the environment/context/configuration... We should still query the LS for those.
However, what would be good from a UX POV would be to filter out codeActions that return a WorkspaceEdit that affects read-only files.

rubenporras:
@andrewL-avlq, I have seen similar errors happening on another PRs after this PR has been merged. Could you take a look to be sure that the test failures where unrelated?

mickaelistria:
I'm pretty sure they're unrelated to this particular change and am currently looking at how to avoid them in tests. I suspect just reorganizing the close/delete blocks will do the work.

mickaelistria:
However, it doesn't look like this test cover the addition of filtering changes that affect read-only content

andrewL-avlq:
The failing tests seem to be the result of errors logged while performing AllCleanRule.finished - given that it is the  NoErrorLoggedRule test rule that is setting the test as failed, a simple way to ignore such exceptions would be to order the rules (add (order=x) to the annotations) so that the NoErrorLoggedRule is 'inner'.

The test added in this PR tests the filtering of marker resolutions on markers, and fails if you set the filter to always return true.

I didn't see any other tests for the behavior of code actions - do you think more should be added?

mickaelistria:
> The failing tests seem to be the result of errors logged

I attempted to fix it in #585 ; let's hope it works reliably.

>  do you think more should be added?

If you don't feel the need for more at the moment, then it's fine to stick to the current state. If need be, they will be improved later.

PR #576, opened 2023-03-28
Match diagnostics with empty ranges

mickaelistria:
In Eclipse, we set a minimal range of 1 character, even for diagnostics that have an empty range (just a position); so this fixes ensure such markers and diagnostics are still matched together upon update, to avoid re-creating diagnostics.

### DIFF
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/diagnostics/LSPDiagnosticsToMarkers.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/diagnostics/LSPDiagnosticsToMarkers.java
@@ -207,7 +207,7 @@ public class LSPDiagnosticsToMarkers implements Consumer<PublishDiagnosticsParam
 		for (IMarker marker : remainingMarkers) {
 			try {
 				if (LSPEclipseUtils.toOffset(diagnostic.getRange().getStart(), document) == MarkerUtilities.getCharStart(marker)
-						&& LSPEclipseUtils.toOffset(diagnostic.getRange().getEnd(), document) == MarkerUtilities.getCharEnd(marker)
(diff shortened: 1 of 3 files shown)


org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/diagnostics/LSPDiagnosticsToMarkers.java
```diff
@@ -207,7 +207,7 @@ private IMarker getExistingMarkerFor(IDocument document, Diagnostic diagnostic,
 		for (IMarker marker : remainingMarkers) {
 			try {
 				if (LSPEclipseUtils.toOffset(diagnostic.getRange().getStart(), document) == MarkerUtilities.getCharStart(marker)
-						&& LSPEclipseUtils.toOffset(diagnostic.getRange().getEnd(), document) == MarkerUtilities.getCharEnd(marker)
+						&& (LSPEclipseUtils.toOffset(diagnostic.getRange().getEnd(), document) == MarkerUtilities.getCharEnd(marker) || Objects.equals(diagnostic.getRange().getStart(), diagnostic.getRange().getEnd()))
```
rubenporras, line 210: Since when we create the marker, we already set the end position to be one char forward, would the existing condition not match already? Why do we need a new one?

mickaelistria: > Since when we create the marker, we already set the end position to be one char forward

Only in case of empty range

> would the existing condition not match already? 

No, it does not match this case.

mickaelistria:
The test failure is actually caused by the MockLanguageServer being incorrect in some aspects and sending diagnostics from the wrong server. I'm working on improving this in a dedicated commit.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/diagnostics/LSPDiagnosticsToMarkers.java
rubenporras: thanks

PR #580, opened 2023-03-29
Type hierarchy

mickaelistria:


org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/typeHierarchy/TypeHierarchyContentProvider.java
```diff
@@ -0,0 +1,77 @@
...
+			}
+		}
+		return new Object[0];
+	}
+
+	@Override
+	public Object[] getChildren(Object parentElement) {
+		if (parentElement instanceof TypeHierarchyItem parentItem) {
+			try {
+				LanguageServers.forDocument(document).withPreferredServer(lsDefinition)
```
rubenporras, line 56: I think you must be sure to target the LS that answered the prepareTypeHierarchy, so in the method above you should remember the wrapper, and use `wrapper.execute(ls -> ls....)` in here. You should find a couple of example in the code base, for example for prepare rename & rename.

org.eclipse.lsp4e/plugin.xml
```diff
@@ -108,6 +108,11 @@
             id="org.eclipse.lsp4e.selectionRange.down"
             name="%command.selection.range.down.name">
       </command>
+      <command
+            categoryId="org.eclipse.lsp4e.category"
+            id="org.eclipse.lsp4e.typeHierarchy"
```
ghentschke, line 113: I could not call the `Show Type Hierarchy` from context menu on text selection until i added:
```xml
         <command
               commandId="org.eclipse.lsp4e.typeHierarchy">
         </command>
```
after line 331 below `org.eclipse.lsp4e.openCallHierarchy`  under the menuContribution `popup:#TextEditorContext?before=org.eclipse.ui.genericeditor.findReferences`

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/typeHierarchy/TypeHierarchyContentProvider.java
```diff
@@ -0,0 +1,77 @@
...
+	private IDocument document;
+
+	public TypeHierarchyContentProvider(LanguageServerDefinition lsDefinition, IDocument document) {
+		this.lsDefinition = lsDefinition;
+		this.document = document;
+	}
+
+	@Override
+	public Object[] getElements(Object inputElement) {
+		if (inputElement instanceof ITextSelection textSelection) {
```
ghentschke, line 37: inputElement is a `java.lang.Object` only. So nothing will be done here.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/typeHierarchy/TypeHierarchyDialog.java
```diff
@@ -0,0 +1,94 @@
...
+
+	public TypeHierarchyDialog(@NonNull Shell parentShell, @NonNull IDocument document, @NonNull LanguageServerDefinition ls) {
+		super(parentShell, PopupDialog.INFOPOPUPRESIZE_SHELLSTYLE, true, true, true, false, false, null, null);
+		this.lsDefinition = ls;
+		this.document = document;
+		create();
+	}
+
+	@Override
+	protected Control createDialogArea(Composite parent) {
```
ghentschke, line 56: Will be called but no dialog will appear.
Nothing happens after this is pressed:
image

ghentschke:
When I add the  `Show Type Hierarchy` menu entry to the context menu in the plugin.xml, a type hierarchy window pops up 1 time:
image
When I close the pop up and try it again, the window won't be opened. I have to restart the IDE.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/typeHierarchy/TypeHierarchyHandler.java
```diff
@@ -0,0 +1,41 @@
...
+import org.eclipse.jface.text.IDocument;
+import org.eclipse.jface.text.ITextSelection;
+import org.eclipse.lsp4e.LSPEclipseUtils;
+import org.eclipse.lsp4e.LanguageServers;
+import org.eclipse.lsp4j.ServerCapabilities;
+import org.eclipse.ui.IEditorPart;
+import org.eclipse.ui.handlers.HandlerUtil;
+import org.eclipse.ui.texteditor.ITextEditor;
+
+public class TypeHierarchyHandler extends AbstractHandler implements IHandler {
```
rubenporras, line 26: should this not extend `LSPDocumentAbstractHandler` and implement the `setEnabled` method as all other handlers?

mickaelistria:
I think this current PR is good enough for a merge. Any objection?

org.eclipse.lsp4e/plugin.properties
```diff
@@ -33,4 +33,5 @@ command.selection.range.up.name=Enclosing Element
 command.selection.range.up.description=Expand Selection To Enclosing Element
 command.selection.range.down.name=Restore To Last Selection
 command.selection.range.down.description=Expand Selection To Restore To Last Selection
-markers.name = Language Servers
\ No newline at end of file
+markers.name = Language Servers
+typeHierarchy.command.name = Show Type Hierarchy
```
ghentschke, line 37: Wouldn't it be better to call it `Open Type Hierarchy` instead of `Show Type Hierarchy` for consistency (see `Open Call Hiererachy`)

ghentschke:
Sorry, I'am a little confused but I cannot figure out how to open the type hierarchy (it's not in the context menu of a text selection)

mickaelistria:
> Sorry, I'am a little confused but I cannot figure out how to open the type hierarchy (it's not in the context menu of a text selection)

Ctrl+3 > Show type hierarchy, or F4. At this stage, the context menu gets messy, I'd rather group some search/navigate operations (such as call hierarchy, type hierarchy...) together into a parent item. But I'd prefer considering it in a separate PR.

ghentschke:
> Ctrl+3 > Show type hierarchy, or F4. At this stage, the context menu gets messy, I'd rather group some search/navigate operations (such as call hierarchy, type hierarchy...) together into a parent item. But I'd prefer considering it in a separate PR.

Okay, I've still the issue that it works only 1 time. 

Is it planed to to add a `Open Type Hierarchy`  in a new PR which would open a view instead of opening a pop up only? I think in a view context would `Open`a better naming. 
I think the current Implantation can be compared to the functionality `Quick Type Hierarchy` in Eclipse.

ghentschke:
>> Okay, I've still the issue that it works only 1 time.

> I'll look at it before merging.

Not it works. Thank you!

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/typeHierarchy/TypeHierarchyContentProvider.java
rubenporras: Hi @mickaelistria , do you plan to follow on this comment? The code might fail like this if there is more than one server that has the required capability.

mickaelistria: I don't think I'll work on it until I face that issue in one of the integrations I'm dealing with. So if you feel interested in implementing it, that would be very welcome!

PR #583, opened 2023-03-31
Add icon for NULL SymbolKind

gerdleon:
Modified the "class.svg" icon.


mickaelistria:
I'm not so sure about the icon to represent null. What about something like Ø ?

gerdleon:
> I'm not so sure about the icon to represent null. What about something like Ø ?

Yes, let's do that.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/ui/LSPImages.java
```diff
@@ -34,219 +34,222 @@
 
 public final class LSPImages {
 
-	private LSPImages() {
```
rubenporras, line 37: @gerdleon , it looks like the line tabs are now spaces, could you convert them back to tabs?

gerdleon: > I'm also not sure deriving null icon from the class one makes a lot of sense: null is not about defining something, but rather an undefined reference, or a reference to nothing, IMO, the simplest would be to put the symbol on a transparent background, similarly to number or string or boolean icons which are about "native" concepts.

Makes sense.

gerdleon:
I will look into a better icon.

mickaelistria:
thank you!

Issue #586, opened 2023-04-03
An NPE occured in `SemanticTokensDataStreamProcessor.getStyleRanges`


vrubezhny:
The NPE is repeatedly occurring in WWD JUnit tests: https://ci.eclipse.org/wildwebdeveloper/job/Wildwebdeveloper/job/PR-1158/1/consoleText

The reason is probably the fact that an editor is closed (and the reconciled is uninstalled) during the reconciling job is being executed.

```
Exception in thread "org.eclipse.jface.text.reconciler.Reconciler" java.lang.NullPointerException: Cannot invoke "org.eclipse.lsp4e.operations.semanticTokens.SemanticTokensDataStreamProcessor.getStyleRanges(java.util.List, org.eclipse.lsp4j.SemanticTokensLegend)" because "this.semanticTokensDataStreamProcessor" is null
	at org.eclipse.lsp4e.operations.semanticTokens.SemanticHighlightReconcilerStrategy.saveStyle(SemanticHighlightReconcilerStrategy.java:183)
	at org.eclipse.lsp4e.operations.semanticTokens.VersionedSemanticTokens.apply(VersionedSemanticTokens.java:40)
	at org.eclipse.lsp4e.operations.semanticTokens.SemanticHighlightReconcilerStrategy.lambda$5(SemanticHighlightReconcilerStrategy.java:260)
	at java.base/java.util.Optional.ifPresent(Optional.java:178)
	at org.eclipse.lsp4e.operations.semanticTokens.SemanticHighlightReconcilerStrategy.fullReconcile(SemanticHighlightReconcilerStrategy.java:259)
	at org.eclipse.lsp4e.operations.semanticTokens.SemanticHighlightReconcilerStrategy.reconcile(SemanticHighlightReconcilerStrategy.java:280)
	at org.eclipse.ui.internal.genericeditor.CompositeReconcilerStrategy.reconcile(CompositeReconcilerStrategy.java:63)
	at org.eclipse.jface.text.reconciler.Reconciler.process(Reconciler.java:149)
	at org.eclipse.jface.text.reconciler.AbstractReconciler$BackgroundThread.run(AbstractReconciler.java:207)
```

vrubezhny:
And yet another one after the first one is fixed:
```
Exception in thread "org.eclipse.jface.text.reconciler.Reconciler" java.lang.NullPointerException: Cannot invoke "org.eclipse.jface.text.ITextViewer.getTextWidget()" because "this.viewer" is null
[...]

PR #591, opened 2023-04-04
Fix blocking on SemanticHighlightReconcilerStrategy.cancelSemanticTok…

vrubezhny:
…ensFull

### DIFF
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/semanticTokens/SemanticHighlightReconcilerStrategy.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/semanticTokens/SemanticHighlightReconcilerStrategy.java
@@ -13,6 +13,8 @@ import java.util.List;
 import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
 import java.util.function.Function;
 
 import org.eclipse.core.runtime.IProgressMonitor;
@@ -261,11 +263,11 @@ public class SemanticHighlightReconcilerStrategy
 									Pair.of(semanticTokens, getSemanticTokensLegend(w)), theDocument)));
 
 			try {
-				semanticTokensFullFuture.get() // background thread with cancellation support, no timeout needed
+				semanticTokensFullFuture.get(15, TimeUnit.SECONDS) // background thread with cancellation support, no timeout needed
 						.ifPresent(versionedSemanticTokens -> {
 							versionedSemanticTokens.apply(this::saveStyle, this::invalidateTextPresentation);
 						});
-			} catch (ResponseErrorException | ExecutionException e) {
+			} catch (TimeoutException | ResponseErrorException | ExecutionException e) {
 				if (!CancellationUtil.isRequestCancelledException(e)) {
 					LanguageServerPlugin.logError(e);
 				}


rubenporras:
Hi @vrubezhny , according to the latest discussion with @mickaelistria , we do not need  timeout, and it is commented as such (`// background thread with cancellation support, no timeout needed`). Why do you need the timeout?

vrubezhny:
@rubenporras  WWD Tests get blocked on `SemanticHighlightReconcilerStrategy.cancelSemanticTokensFull` that is invoked from a JUnit test teardown method.
It looks like a `CancellationException` is thrown but never caught which breaks the test being teardown and prevents the following tests from being executed: (For sure the tests get blocked at this point, I'm not sure but it looks like `get()` is never interrupted)  

https://github.com/eclipse/wildwebdeveloper/actions/runs/4602320282/jobs/8131146654?pr=1160

In this case the `cancelSemanticTokenFull` is calling `CompletableFuture.cancel()` from another thread which may be a reason why the test teardown gets blocked.

mickaelistria:
> we do not need timeout

In an ideal world yes, but it seems like Victor has a case where the cancellation is not working as well as expected; so a timeout is still useful. In general, `.get()` without a timeout is really dangerous, so even when cancellation and so on are supposed to be in place, it's still safer to set a timeout just in case.

rubenporras:
Should we not rather investigate why the cancel did not worked? It looks like it was working before, so maybe https://github.com/eclipse/lsp4e/commit/4ba7f961067417b01c7ebd76db6b27575564f337 broke cancellation?

rubenporras:
In addition to my last comment, where I comment that we should rather understand why now suddenly we get cancellation exceptions (and also https://github.com/eclipse/lsp4e/commit/90c7b073ebc56d164aee9358205ab3b0d77b41ac), 5 seconds is not enough.

I did the test because we have our own DSLs, and also big files, and I know it takes also more than 5 seconds for some of the sources our customers work with.

To have a comparison, I tested the JDT as well, with huge java files, and it takes more than five seconds to do semantic highlight (as expected), still it works with huge files, even if the highlighting takes 15 seconds to do the job, it will still do it, so I do not think we should bail out that early. That would be bad user experience. Of course one can always tell the user to write smaller files, but if he does not do it, the editor should still do the job.

vrubezhny:
@rubenporras We can set 15 seconds timeout or tune it later to have even bigger value - it's not the main problem I'm trying to solve.

Problem is that it looks like we're getting blocked on `CompletableFuture.cancel()` - where, in theory, we shouldn't get blocked. The blocking may happen on infinite waiting (`get()`) or just on JUnit test crashing/infinite wait at some point down the stacktrace given which should not happen. 

So this PR is more a try to exclude the possible infinite wait on `get()`, as a normal user, IMHO, will not be able to reproduce the same circumstances in 99% - the reconciling should be happening at the same time a user forces the editor to be closed  (and I'm not sure if normal 'Close Editor' action repeats the steps taken by `PlatformUI.getWorkbench().getActiveWorkbenchWindow().getActivePage().closeAllEditors(false);`

vrubezhny:
@rubenporras Yes, this issue as well as https://github.com/eclipse/lsp4e/issues/586 probably have  started to  appeare after https://github.com/eclipse/lsp4e/commit/90c7b073ebc56d164aee9358205ab3b0d77b41ac as WWD currently uses LSP4E snapshots. 

My assumption is only in that `SemanticHighlightReconcilerStrategy.uninstall` method (and probably some others) now is either invoked at different point (or probably from a different thread) which made it to clear the  `SemanticHighlightReconcilerStrategy` members before all the CompletableFutures are resolved, which is not a bad thing, but the future cancellation was also affected

rubenporras:
I think we should rather investigate why `SemanticHighlightReconcilerStrategy.uninstall` is throwing an exception with the new reconciler registration introduced in https://github.com/eclipse/lsp4e/commit/4ba7f961067417b01c7ebd76db6b27575564f337 and was not doing it before. @angelozerr , could you look into this exception thrown in https://github.com/eclipse/wildwebdeveloper/actions/runs/4602320282/jobs/8131146654?pr=1160?
```

Exception in thread "org.eclipse.jface.text.reconciler.Reconciler" java.util.concurrent.CancellationException
	at java.base/java.util.concurrent.CompletableFuture.cancel(CompletableFuture.java:2478)
	at org.eclipse.lsp4e.operations.semanticTokens.SemanticHighlightReconcilerStrategy.cancelSemanticTokensFull(SemanticHighlightReconcilerStrategy.java:244)
	at org.eclipse.lsp4e.operations.semanticTokens.SemanticHighlightReconcilerStrategy.uninstall(SemanticHighlightReconcilerStrategy.java:145)
```

If uninstalling the reconciler throw exceptions when cancelling the completableFuture I think we need to look at that and not create a timeout. Otherwise it is very likely that exceptions will occur also for users each time they close and editor, and that is not a good state. In the same way, I do not think we want logs of unit tests full of exceptions ;)

About the 15sec timeout: that looks better to me, but it feels arbitrary and side-stepping the problem without understanding the root cause.

vrubezhny:
Looking at https://github.com/eclipse/lsp4e/commit/4ba7f961067417b01c7ebd76db6b27575564f337 I see that `uninstall` was never invoked on the reconcilers before this change (or was intentionally invoked when needed) because `install` and `uninstall` methods were not implementing/overriding the `ITextViewerLifecycle` interface methods. So, they were never invoked during the Text Viewer Lifecycle - this probably is the reason why `SemanticHighlightReconciler.uninstall` (before the https://github.com/eclipse/lsp4e/commit/4ba7f961067417b01c7ebd76db6b27575564f337) was never invoked from `PlatformUI.getWorkbench().getActiveWorkbenchWindow().getActivePage().closeAllEditors(false);`

After the mentioned change the `uninstall` method is started to be invoked always when a Text Viewer needs it (on editor close - is for sure).

But still a question for me is wh/wherey the `CancellationException` is thrown? Yes it is created in `CompletableFuture.cancel()`,  but then it is not thrown, but  is set as `AltResult` value.  There is some post-processing on that result, but still I can't find where it's been thrown in the code

vrubezhny:
So, I believe this change worth to be merged, however it doesn't solve the initial problem of WWD JUnit tests blocking/hanging up when running on Windows (See: https://github.com/eclipse/wildwebdeveloper/actions/runs/4602320282/jobs/8131146654?pr=1160). 

I was able to reproduce the problem, running WWD JUnit tests locally on Windows and the tests still hang up the same way as they do when running WindowsPR on GitHub even when using LSP4E with this PR included.

rubenporras:
I will try to take a look at the problem this week/next week, I would like to understand why if `cancel()` the threads are blocked, I think that should not happen, so I am not convinced that setting a timeout is a good solution.

vrubezhny:
Reading the `jstack` log I see that the "main" thread is just suspended on Reconciler `uninstall` without any chance to be resumed:
 
```
"main" #1 prio=6 os_prio=0 cpu=113328.13ms elapsed=549.93s allocated=1852M defined_classes=9594 tid=0x000001fea4880ec0 nid=0x1b0c in Object.wait()  [0x0000005fbd5f9000]
   java.lang.Thread.State: WAITING (on object monitor)
	at java.lang.Object.wait(java.base@17.0.6/Native Method)
	- waiting on <no object reference available>
	at java.lang.Object.wait(java.base@17.0.6/Object.java:338)
	at org.eclipse.jface.text.reconciler.AbstractReconciler$BackgroundThread.suspendCallerWhileDirty(AbstractReconciler.java:123)
	- locked <0x000000074d353b48> (a org.eclipse.jface.text.reconciler.DirtyRegionQueue)
	at org.eclipse.jface.text.reconciler.AbstractReconciler$Listener.inputDocumentAboutToBeChanged(AbstractReconciler.java:271)
	at org.eclipse.jface.text.reconciler.AbstractReconciler.uninstall(AbstractReconciler.java:479)
	at org.eclipse.jface.text.reconciler.Reconciler.uninstall(Reconciler.java:192)
	at org.eclipse.ui.internal.genericeditor.CompositeReconciler.uninstall(CompositeReconciler.java:63)
	at org.eclipse.jface.text.source.SourceViewer.unconfigure(SourceViewer.java:739)
	at org.eclipse.jface.text.source.SourceViewer.handleDispose(SourceViewer.java:800)
	at org.eclipse.jface.text.source.projection.ProjectionViewer.handleDispose(ProjectionViewer.java:1303)
	at org.eclipse.jface.text.TextViewer.lambda$1(TextViewer.java:1741)
	at org.eclipse.jface.text.TextViewer$$Lambda$1042/0x0000000801265e30.widgetDisposed(Unknown Source)
	at org.eclipse.swt.widgets.TypedListener.handleEvent(TypedListener.java:127)
	at org.eclipse.swt.widgets.EventTable.sendEvent(EventTable.java:89)
```

rubenporras:
> So, I believe this change worth to be merged, however it doesn't solve the initial problem of WWD JUnit tests blocking/hanging up when running on Windows (See: https://github.com/eclipse/wildwebdeveloper/actions/runs/4602320282/jobs/8131146654?pr=1160).
> 
> I was able to reproduce the problem, running WWD JUnit tests locally on Windows and the tests still hang up the same way as they do when running WindowsPR on GitHub even when using LSP4E with this PR included.

thanks for looking into the problem as well. I am sorry that I missed you post yesterday in between other tasks. If it does not solve the problem, in my opinion the timeout would worsen the user experience, there is no reason to cancel a request just because the server is slow, if we can wait without blocking the UI and we cancel it whenever it is not needed (as in the case of closing the editor or changing the document).

rubenporras:
> But still a question for me is wh/wherey the `CancellationException` is thrown? Yes it is created in `CompletableFuture.cancel()`, but then it is not thrown, but is set as `AltResult` value. There is some post-processing on that result, but still I can't find where it's been thrown in the code

That is the expected behaviour. `CompletableFuture.cancel()` will set the result of the future, the thread waiting for the future will be woken up and receive the exception. I think we should catch the exception and not log it, as it is an expected situation. I do not think the user is interested to see it in the log. I will create a PR later for that.

I think that we need to look into why `AbstractReconciler$BackgroundThread.suspendCallerWhileDirty` is not returning. I will look into that whenever I have time.

rubenporras:
@vrubezhny , I have seen that at

> 	org.eclipse.jface.text.reconciler.AbstractReconciler$BackgroundThread.suspendCallerWhileDirty(AbstractReconciler.java:123)

the code is waiting on `Object#wait` on the object `dirtyRegionQueue`.

The comment on the method says "Suspends the caller of this method until this background thread has emptied the dirty region queue." I have the impression that by handling the cancellation in https://github.com/eclipse/lsp4e/pull/603 this might work properly now.

Can you check your if your jobs run successfully with a newer LSP4E snapshot?

vrubezhny:
@rubenporras Thanks for looking at `suspendCallerWhileDirty` - Your fix (#603)  is actually already tested on today's PR WindowsPR builds on GitHub and I'll double check it on my local Windows.

PR #603, opened 2023-04-06
Handle client cancellation without reporting an error

rubenporras:


rubenporras:
@vrubezhny I think this will not prevent the hang in your tests, but will prevent the error log which should not be there.

PR #604, opened 2023-04-06
Improve log when shutting down the server times out

rubenporras:


org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServerWrapper.java
```diff
@@ -440,11 +440,12 @@ public synchronized void stop() {
 
 		Runnable shutdownKillAndStopFutureAndProvider = () -> {
 			if (languageServerInstance != null) {
-				CompletableFuture<Object> shutdown = languageServerInstance.shutdown();
 				try {
-					shutdown.get(5, TimeUnit.SECONDS);
+					languageServerInstance.shutdown().get(5, TimeUnit.SECONDS);
```
rubenporras, line 444: since this runs async and will not block, and we do not handle timeout exceptions other than to continue, I wonder if we should call just `get()`, the [spec](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#shutdown) cleatly states that 

> Clients should also wait with sending the exit notification until they have received a response from the shutdown request

which we are clearly not doing here in the case the shutdown takes more than five secs.

rubenporras:
I never got around finishing the PR, I will close it to reduce the work that we do on each PR that is merged.

PR #605, opened 2023-04-06
Test case demoing eclipseide-jdtls#46

akurtakov:
Test case showing the issue faced in https://github.com/redhat-developer/eclipseide-jdtls/issues/46

### DIFF
--- a/org.eclipse.lsp4e.test/src/org/eclipse/lsp4e/test/completion/CompleteCompletionTest.java
+++ b/org.eclipse.lsp4e.test/src/org/eclipse/lsp4e/test/completion/CompleteCompletionTest.java
@@ -364,6 +364,19 @@ public class CompleteCompletionTest extends AbstractCompletionTest {
 		assertEquals(" and foo", viewer.getDocument().get());
 		// TODO check link edit groups
 	}
+	
+	@Test
+	public void testBasicSnippetReplace() throws PartInitException, CoreException {
+		CompletionItem completionItem = createCompletionItem("System.out.println(${0});", CompletionItemKind.Class, new Range(new Position(0, 4), new Position(0, 4)));
+		completionItem.setInsertTextFormat(InsertTextFormat.Snippet);
+		MockLanguageServer.INSTANCE.setCompletionList(new CompletionList(false, Collections.singletonList(completionItem)));
+		ITextViewer viewer = TestUtils.openTextViewer(TestUtils.createUniqueTestFile(project,"syso"));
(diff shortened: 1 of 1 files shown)


org.eclipse.lsp4e.test/src/org/eclipse/lsp4e/test/completion/CompleteCompletionTest.java
```diff
@@ -364,6 +364,19 @@ public void testBasicSnippet() throws PartInitException, CoreException {
 		assertEquals(" and foo", viewer.getDocument().get());
 		// TODO check link edit groups
 	}
+	
+	@Test
+	public void testBasicSnippetReplace() throws PartInitException, CoreException {
+		CompletionItem completionItem = createCompletionItem("System.out.println(${0});", CompletionItemKind.Class, new Range(new Position(0, 0), new Position(0, 1)));
```
mickaelistria, line 370: This range in CompletionItem exactly tells to replace the 1st character by the new text, so the current behavior seems correct.

akurtakov: I've corrected the range to be 0,4-0,4 as this is exactly the type of range JDT.ls returns.
```
LANGUAGE_SERVER_TO_LSP4E com.redhat.eclipseide.jdtlsclient.inProcessServer:
{"jsonrpc":"2.0","id":"69","result":{"label":"sysout","kind":15,"detail":"print to standard out","documentation":{"kind":"markdown","value":"```java\nSystem.out.println();\n```"},"insertText":"System.out.println(${0});","insertTextFormat":2,"textEdit":{"range":{"start":{"line":5,"character":5},"end":{"line":5,"character":5}},"newText":"System.out.println(${0});"},"command":{"title":"","command":"java.completion.onDidSelect","arguments":["17","0"]}}}
```

mickaelistria: Still, this really tells the client to insert the whole proposal at offset 4, see https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textEdit about TextEdit. So I think LSP4E is correct. What makes things work in VSCode-Java is something else, either VSCode-Java receives another item or adds non-standard customization afterwards.

akurtakov:
Issue is in the Java LS not in LSP4E https://github.com/eclipse/eclipse.jdt.ls/issues/2584

PR #606, opened 2023-04-06
Fix SemanticHighlightReconcilerStrategy uninstallation

vrubezhny:
This PR adds a better processing during the uninstallation of SemanticHighlightReconcilerStrategy:
- Removes the repeated invokation of  `SemanticHighlightReconcilerStrategy.cancelSemanticTokensFull()` after `uninstall` is invoked (it could be invoked once again from `fullReconcile')
- Skips as much as it possible any processing after `uninstall` is invoked by disabeling it in `fullReconcile` if uninstall is already invoked (so, actually no any processing is needed after that)


rubenporras:
thanks

PR #607, opened 2023-04-11
Intern the validation code and source of the diagnostic to save memory,

rubenporras:
as this is kept as part of the marker attribute, which is kept in memory until the marker is addressed, that is potentially as long as the IDE is running.

Since this strings will be the same for many markers, it makes sense to keep only a canonical representation. Note that the message should not be interned, as it can contain specific information to a marker instance.


rubenporras:
The JVM does not optimize duplicated immutable objects. One could argue that LS P4J should do it, but the problem is not creating the diagnostic when it is deserialized, rather that we are keeping all the diagnostics in memory by adding them to the markers view. Thus I think it is fine to intern the string in the place where it is "persisted".

All that being said, if you still think LS P4J is better, just mention that again and I will prepare the PR.

Issue #610, opened 2023-04-13
Language Server sometimes get into start/stop loop


BoykoAlex:
One of Spring Boot LS plugin users sometimes observes Boot LS process being started/stopped very frequently over a short period of time.
Here is the eclipse log. [metadata.zip](https://github.com/eclipse/lsp4e/files/11226192/metadata.zip )

I have noticed the chunk of the stacktrace below in the log:
```
	at java.base/java.util.concurrent.CompletableFuture.cancel(CompletableFuture.java:2478)
	at org.eclipse.lsp4e.LanguageServerWrapper.stop(LanguageServerWrapper.java:429)
	at org.eclipse.lsp4e.LanguageServerWrapper.start(LanguageServerWrapper.java:250)
	at org.eclipse.lsp4e.LanguageServerWrapper.getInitializedServer(LanguageServerWrapper.java:698)
	at org.eclipse.lsp4e.LanguageServers$LanguageServerDocumentExecutor.filter(LanguageServers.java:263)
```
My only guess therefore is that `LanguageServerWrapper.start()` is called once and then again before `isActive()` starts returning `true`, i.e. before `launcherFuture` is initialized. This triggers `stop()` (from the body of the `start()` method before LS is initialized and connected to the IDE client. I'd check `isActive() || (this.initializeFuture != null && !this.initializeFuture.isDone() && !this.initializeFuture.isCancelled())` rather then just `isActive()`.

See https://github.com/spring-projects/sts4/issues/925

BoykoAlex:
Should be fixed

PR #611, opened 2023-04-13
Prevent infinite start/stop loop - initialize launcherFuture early

BoykoAlex:
Naive and simple fix for #610


BoykoAlex:
In case `isActive()` needs to be left alone as is, another option could be check `isActive() || initializedFuture != null` perhaps or perhaps brand new boolean flag `isStarted` which would become true when `start()` called and false when `stop()` is called as well as initially... Feedback is welcomed

mickaelistria:
I don't see `isActive` modified in this patch. But I think it'd be OK to modify it if it's missing 1 condition.

BoykoAlex:
`isActive()` looks at `launcherFuture` value... `isActive()` will start returning `true` sooner than it used to (which is a good thing i think) before lsp4j jsonrpc `Launcher` is started (ls to client connection established?). This is what i meant by the change to `isActive()` - implementation not changed but behaviour is changed.

PR #584, opened 2023-03-31
Run open in UIJob (#579)

ghentschke:
Fix for IllegalArgumentException when applying fix #579

PR #612, opened 2023-04-17
Enforce UI thread for open(...) call

BoykoAlex:


mickaelistria:
See also #584 ; @ghentschke which patch do you think is better?

ghentschke:
Hi, I prefer this change here by @BoykoAlex. My change has still two unit test errors. Haven't had the the time to check why.

mickaelistria:
Thanks for the patch and the review.

PR #614, opened 2023-04-20
Resolve code action only if edit and command are null (vscode behavior).

angelozerr:
Resolve code action only if edit and command are null (vscode behavior).


mickaelistria:
I don't think this is correct according to the spec. The relevant part in the spec says "In general, a client should offer data support if it offers resolve support. It should also be noted that servers shouldn’t alter existing attributes of a code action in a codeAction/resolve request.", but in case one of the attributes is null, it can be changed in resolved.
So I think we can skip resolve when `edit != null && command != null` as we already have the final value for actionable attributes; but in case one isn't set, we may still need resolve to fill on of them.

mickaelistria:

PR #623, opened 2023-05-01
Project relative path in Open Declaration popup where applicable

BoykoAlex:


mickaelistria:
Wouldn't it be better to make it relative to current resource?

martinlippert:
> Wouldn't it be better to make it relative to current resource?

I like that idea, since it might make the path even shorter (which is good I think), but on the other hand I am worried about the case where the popup shows multiple resources that are in a different area of the project, so that the path would show up like `../../some/other/application.yml` and `../../../../bar/application.properties`, which might make it quite difficult to understand each of those locations.

mickaelistria:
So maybe showing resource-relative path for siblings and their children; and maybe even single parent (only one `../`), and a project relative path for other cases?

martinlippert:
Maybe it makes even sense to show the name of the file as a first element of the path, then a visual separator (e.g. ` - `), followed by the path from the project root or a resource-relative path for siblings and their children.

My assumption is that the file name is probably the most important part to visualize here, so showing this front and center might make sense, I think.

BoykoAlex:
I wouldn't want to pass the document URI to calculate the path relative to the currently opened file path... It is highly likely that file opened in the editor belongs to the workspace. Therefore showing **<locationType> - <file name> - <workspace relative file path>** should be okay. Don't think project relative path would be too long. VSCode seem to show workspace relative paths as well. If declaration file is not in the workspace then full absolute file path is okay to show i think...  we could in this case show the file name in the middle as well... Don't think

mickaelistria (approved):
I think the current proposal is good enough to be merged.

Issue #635, opened 2023-05-11
IllegalArgumentException on Apply Resolution


rubenporras:
I got today this exception testing LSP4E:

Message: An internal error occurred during: "apply resolution - ....".

java.lang.IllegalArgumentException
	at org.eclipse.ui.ide.IDE.openEditor(IDE.java:572)
	at org.eclipse.lsp4e.LSPEclipseUtils.openEditor(LSPEclipseUtils.java:750)
	at org.eclipse.lsp4e.LSPEclipseUtils.openFileLocationInEditor(LSPEclipseUtils.java:709)
	at org.eclipse.lsp4e.LSPEclipseUtils.open(LSPEclipseUtils.java:625)
	at org.eclipse.lsp4e.LSPEclipseUtils.open(LSPEclipseUtils.java:611)
	at org.eclipse.lsp4e.LSPEclipseUtils.open(LSPEclipseUtils.java:607)
	at org.eclipse.lsp4e.LSPEclipseUtils.lambda$1(LSPEclipseUtils.java:867)
	at java.base/java.util.Optional.ifPresent(Optional.java:178)
	at org.eclipse.lsp4e.LSPEclipseUtils.applyWorkspaceEdit(LSPEclipseUtils.java:863)
	at org.eclipse.lsp4e.operations.codeactions.CodeActionMarkerResolution.run(CodeActionMarkerResolution.java:66)
	at org.eclipse.ui.internal.genericeditor.markers.MarkerInformationControl$1$1.run(MarkerInformationControl.java:141)
	at org.eclipse.core.internal.jobs.Worker.run(Worker.java:63)

PR #636, opened 2023-05-11
Illegal argument exception

rubenporras:
for #635


mickaelistria:
Can you please squash those commits into a single one?

rubenporras:
done

PR #639, opened 2023-05-15
[#638] Fix outline view for external files

### LINKED ISSUE #638: Outline view does not work for files outside of the workspace
When a file is opened from outside of the workspace, the outline view does not work (see https://github.com/eclipse-cdt/cdt-lsp/issues/63#issue-1707639564)

The problem is that for external files, the fields in `documentFile` and `documentURI` of the  `org.eclipse.lsp4e.outline.LSSymbolsContentProvider.OutlineViewerInput.OutlineViewerInput`  are null.

This leads to problems when `org.eclipse.lsp4e.outline.LSSymbolsContentProvider.refreshTreeContentFromLS()` gets called, since the `documentURI`  is null.

I can provide a PR on monday.

ghentschke:
Fixes #638
Since there is no IFile object for a file outside the workspace, replaced
IFile by IDocument in DocumentSymbolWithFile. Maybe rename DocumentSymbolWithFile into DocumentSymbolWithDocument?


mickaelistria:
The test failure on CI is unrelated, let's merge.
Thanks a lot!

PR #640, opened 2023-05-15
[#615] fix missing space in outline view text

### LINKED ISSUE #615: Outline view should provide more visual separation displaying symbols gathered from properties files
## PROBLEM

Using Eclipse 4.27 with Quarkus Tools (using LSP4MP 0.6)  I see this for application.properties in the Outline view 

image

Note how the key and value are collapsed together without a space.

On the other hand, looking at a Java file in Eclipse, we see  `<element name> : <type>` with the space and colon.

image

## SUGGESTION

I think simply adding a space, combined with the different font, would be enough of a separation (we don't need the ':' IMHO).

## OTHER INFO

Note the [comment](https://github.com/eclipse/lsp4mp/issues/386#issuecomment-1516851206) showing the JSON-RPC payload returned by LSP4MP in a case similar to my screenshot.   (I first opened the issue against LSP4MP before getting directed here).

ghentschke:
fixes #615


mickaelistria:
Thanks!

PR #642, opened 2023-05-15
Use URI instead of IDocument for DocumentSymbolWithFile

ghentschke:
Since URI is more generic than IDocument this type should be preferred. Its not possible to get the URI or IFile of an IDocument by using the static helper methods in LSPEclipseUtils when the IDocument is not opened in an editor (buffer is null then). There are use-cases where it makes sense to use a DocumentSymbolWithFile object when the document is not opened in an editor (e.g. in an content provider for the project explorer view).


ghentschke:
@mickaelistria This would be very helpful for us in https://github.com/eclipse-cdt/cdt-lsp/issues/36

mickaelistria:
Thanks!

PR #647, opened 2023-05-18
Use LSP4J 0.21

mickaelistria:


target-platforms/target-platform-latest/target-platform-latest.target
```diff
@@ -15,7 +15,7 @@
 			<unit id="org.eclipse.lsp4j.jsonrpc.debug" version="0.0.0"/>
 			<unit id="com.google.gson" version="0.0.0"/>
 			<unit id="com.google.guava" version="0.0.0"/>
-			<repository location="https://download.eclipse.org/lsp4j/updates/releases/0.20.1/"/>
+			<repository location="https://download.eclipse.org/lsp4j/updates/releases/0.21.0/"/>
```
cdietrich, line 18: depending what you do in your codebase, you can remove xbase (and guava) now

mickaelistria: Awesone, thanks for the heads up!

cdietrich: looks like guava is still used in a few places in lsp4e codebase

mickaelistria:
The dash license issue can be ignored now; as we're removing deps and not adding any.

PR #648, opened 2023-05-19
Make Outline view works for documents with custom URI

akurtakov:
JDT.ls uses custom protocol for sources coming from jvm/jar files (e.g. jdt://contents/java.base/java.lang/Appendable.class?=proba/%5C/usr%5C/lib%5C/jvm%5C/java-17-openjdk-17.0.7.0.7-5.fc38.x86_64%5C/lib%5C/jrt-fs.jar%60java.base=/javadoc_location=/https:%5C/%5C/docs.oracle.com%5C/en%5C/java%5C/javase%5C/17%5C/docs%5C/api%5C/=/%3Cjava.lang(Appendable.class ) which is retrievable from ITextFileBuffer if/when custom EFS for the protocol is provided.
Extra change was required in LSPEclipseUtils.getDocument(URI) which wrongly assumed that every IDocument has to be backed by file.

### DIFF
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LSPEclipseUtils.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LSPEclipseUtils.java
@@ -559,11 +559,7 @@ public final class LSPEclipseUtils {
 		if (resource != null) {
 			return getDocument(resource);
 		}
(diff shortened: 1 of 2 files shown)


akurtakov:
Fixes https://github.com/redhat-developer/eclipseide-jdtls/issues/52

mickaelistria:
Thanks!

Issue #645, opened 2023-05-16
Override markerType in IMarkerAttributeComputer


ChristophKaser:
I am trying to have some "special" diagnostics appear different to the "default" diagnostics in the Editor, e.g. to indicate diagnostics differently depending on the source. The easiest way to do that seems to be to specify a different markerType and configure the annotationType per markerType (or maybe I got lost in the Marker vs Annotation distinction?).

Unfortunately, while I can configure marker attributes per diagnostic in IMarkerAttributeComputer, there is no way to override the markerType.

I suggest adding a constant IMarkerAttributeComputer.MARKER_TYPE that can be used in addMarkerAttributesForDiagnostic to set the markerType by setting it in attributes:

```java
public void addMarkerAttributesForDiagnostic(Diagnostic diagnostic, IDocument document, IResource resource, Map<String, Object> attributes) {
  attributes.put(IMarkerAttributeComputer.MARKER_TYPE , "de.dummy.marker_type");
}
```

Alternatively, a new method in IMarkerAttributeComputer like 
`String getMarkerTypeForDiagnost(Diagnostic diagnostic, IDocument document, IResource resource)` 
(with a default implementation returning null) might be a clearer design.

The returned marker type would need to be used in `LSPDiagnosticsToMarkers.updateMarkers`, which would require to determine `toDeleteMarkers`  without looking for a specific marker type. I don't think that is a problem, however, since findMarkers seems to accept null as type, and the LANGUAGE_SERVER_ID attribute of the marker is checked anyway.

What do you think about this proposal? Am I missing something?
If you think that this proposal makes sense, I can try to provide a PR.

rubenporras:
Hi @ChristophKaser ,

you can configure it using the makerType attribute in the extension
         point="org.eclipse.lsp4e.languageServer", but that would only work for you if you have different servers for the different source types. Is that enough for you or are you using a singleton server?

@mickaelistria , if the solution can be replicated, then fair enough, we can go other ways, but removing and creating the marker again will have a performance impact (as far as I remember it will lock the workspace during the operation). Also the user might see flickering if he has the markers view open. So it would not be the same. I think updating the markers attributes in a listener will have a similar problem. Since the current API is very small and well defined, and does not cause any harm, I would rather keep it, or am I mistaken about the implications.

ChristophKaser:
Hi @rubenporras 

> you can configure it using the makerType attribute in the extension
[...]

PR #649, opened 2023-05-19
LSPDiagnosticsToMarker: Handle subtypes

ChristophKaser:
Changes LSPDiagnosticsToMarker to consider subtypes of the specified markerType so it is possible to customize the marker type based on diagnostic attributes (e.g. in a workspace change listener).

See issue #645.


rubenporras (approved):

mickaelistria:
I see on https://accounts.eclipse.org/users/christophkaser that you have approved the ECA. So I imagine the issue is more about the email address you're using that is not the primary one in the Eclipse Foundation DB or something like that. But the ECA check on your account is good enough to me and I'm merging.
Thank you!
Just a quick ping by the way to @waynebeaton as this small glitch in the ECA check may be interesting to him.

PR #651, opened 2023-05-19
[#650] Handle HTML without HEAD

### LINKED ISSUE #650: Issue with HTML Rendering in Content Assist Completion Proposal
Screenshot 2023-05-18 at 22 30 07

There appears to be an issue with the content assist completion proposal display, where raw HTML is shown to the user. This seems to stem from the method `org.eclipse.lsp4e.operations.hover.FocusableBrowserInformationControl.styleHtml()`, which inserts styles at the `-1` position, thereby disrupting pre-existing elements in the content. The link to the specific line in the source code is [here](https://github.com/eclipse/lsp4e/blob/1e5244520850d07c449ca1d5007c2e05196c1cad/org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/hover/FocusableBrowserInformationControl.java#L165).

The issue is further exacerbated by the  `LSCompletionProposal.getAdditionalProposalInfo()` method. This method doesn't seem to allow for a valid `<head>` element in completion proposal details, as it wraps entirety of `org.eclipse.lsp4j.CompletionItem.getDetail()` return value within a `<p>` (paragraph), as shown [here](https://github.com/eclipse/lsp4e/blob/1e5244520850d07c449ca1d5007c2e05196c1cad/org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/completion/LSCompletionProposal.java#L348). As per HTML standards, the `<head>` element cannot be contained within a `<p>` element.

The problem was originally discovered in https://github.com/vgcpge/eclipse.copilot/issues/6

vgcpge:
Fixes #650

FocusableBrowserInformationControl expects to receive HTML with a HEAD element. However, this isn't always the case - in fact, the input could be plain text with minimal decorations from LSCompletionProposal.getAdditionalProposalInfo().

When no HEAD element is provided, FocusableBrowserInformationControl displays raw HTML in the completion proposal detail pane (see #650).

This pull request addresses the situation where the HTML input doesn't contain a HEAD element by prepending a new HEAD to the input.

I'm uncertain if an automated test is applicable in this case. One possibility could be to extract the styling into a separate class for testing purposes. Should I proceed with this?

### DIFF
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/hover/FocusableBrowserInformationControl.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/hover/FocusableBrowserInformationControl.java
@@ -39,10 +39,13 @@ import org.eclipse.swt.widgets.Shell;
 import org.eclipse.ui.editors.text.EditorsUI;
 import org.eclipse.ui.texteditor.AbstractDecoratedTextEditorPreferenceConstants;
 
+import com.google.common.base.Strings;
+
 @SuppressWarnings("restriction")
 public class FocusableBrowserInformationControl extends BrowserInformationControl {
 
 	private static final String HEAD = "<head>"; //$NON-NLS-1$
+	private static final String HTML_TEMPLATE = "<head>%s</head>%s"; //$NON-NLS-1$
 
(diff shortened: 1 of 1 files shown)


vgcpge:
This account does not have ECA

PR #654, opened 2023-05-23
re-add DSPLaunchDelegate.createDebugTarget(...) as deprecated method

sebthom:
The signature of the public API method `DSPLaunchDelegate.createDebugTarget` was changed by 1e5244520850d07c449ca1d5007c2e05196c1cad resulting in compile / linkage errors in our projects where we need to override it.

This PR re-adds the method with the old signature as deprecated to allow graceful API transition phase.

### DIFF
--- a/org.eclipse.lsp4e.debug/src/org/eclipse/lsp4e/debug/launcher/DSPLaunchDelegate.java
+++ b/org.eclipse.lsp4e.debug/src/org/eclipse/lsp4e/debug/launcher/DSPLaunchDelegate.java
@@ -384,8 +384,26 @@ public class DSPLaunchDelegate implements ILaunchConfigurationDelegate {
 	 * of {@link DSPDebugTarget}, but does not have to be. The arguments to this
 	 * method are normally just passed to {@link DSPDebugTarget} constructor.
 	 */
-	protected IDebugTarget createDebugTarget(SubMonitor subMonitor, Supplier<TransportStreams> streamsSupplier, ILaunch launch, Map<String, Object> dspParameters) throws CoreException {
-		final var target = new DSPDebugTarget(launch, streamsSupplier, dspParameters);
+	protected IDebugTarget createDebugTarget(SubMonitor subMonitor, Supplier<TransportStreams> streamsSupplier,
+			ILaunch launch, Map<String, Object> dspParameters) throws CoreException {
+		/*
+		 * TODO use the following code when removing createDebugTarget deprecated
+		 * final var target = new DSPDebugTarget(launch, streamsSupplier,
+		 * dspParameters); target.initialize(subMonitor.split(80)); return target;
+		 * return target;
+		 */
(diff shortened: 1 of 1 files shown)


org.eclipse.lsp4e.debug/src/org/eclipse/lsp4e/debug/launcher/DSPLaunchDelegate.java
```diff
@@ -384,8 +384,26 @@ public void close() {
...
+	protected IDebugTarget createDebugTarget(SubMonitor subMonitor, Supplier<TransportStreams> streamsSupplier,
+			ILaunch launch, Map<String, Object> dspParameters) throws CoreException {
+		/*
+		 * TODO use the following code when removing createDebugTarget deprecated
+		 * final var target = new DSPDebugTarget(launch, streamsSupplier,
+		 * dspParameters); target.initialize(subMonitor.split(80)); return target;
+		 * return target;
+		 */
+		final var supplier = streamsSupplier.get();
+		return createDebugTarget(subMonitor, supplier::close, supplier.in, supplier.out, launch, dspParameters);
```
mickaelistria, line 396: The newer/recommended method should not reference the deprecated one, it would better be the other way round.

sebthom: That does not work since the new `createDebugTarget` method is invoked by `public void launch(DSPLaunchDelegateLaunchBuilder builderSrc)` of the same class. I would then have to revert all your changes to this class. 

If I make the deprecated method delegate to the new method it will effectively never be called, thus overriding it will be without any effect.

mickaelistria:
The code is not functionally equivalent, while with the current master https://github.com/eclipse/wildwebdeveloper/pull/1174 does work decently (I can debug with vscode-js-debug), this patch leads to a failure on server-side. I suspect one issue is that the Debug target is then not using the correct streams (ie it's trying to keep using the same main stream while a new socket connection to the given port must be made).

sebthom:
I am out of ideas how to solve this in a backwards compatible way then.

mickaelistria:
Unfortunately I don't see an easy way either, the only viable one would be that you copy DSPLaunchDelegate in your code... I think overall, the simpler and most profitable path for everyone (including you) is to just drop backward compatibility and set narrow version ranges to avoid runtime errors. Would it help if we set the bundle version to 0.16.0 instead of 0.15.2 (I don't have high hopes it makes things really easier, but if can facilitate things, it's also an easy change for LSP4E)

sebthom:
Right now it makes no difference for us if it will be a 0.16.0 or 0.15.2.

In the future, the best would be of course if the project follows semver. But I think it would already help if the project commits to always increment the minor version for bundles that introduce backward incompatible changes and only increments the patch version if binary compatibility is guaranteed.

PR #655, opened 2023-05-23
Refactorings->Rename doesn't work in case of more than one LS is defined

vrubezhny:
If more than one LS defined for a document provide a prepareRename capabilities need to aqcuire all LS prepareRename received reults and not just fail on first null-result received.

See: https://github.com/eclipse/wildwebdeveloper/issues/1209


mickaelistria:
Why do we need to ensure all `prepareRename` are sent if in the end we only call rename on a single server? What do you mean by "fail" in "not just fail on first null-result received" ?

vrubezhny:
> Why do we need to ensure all `prepareRename` are sent if in the end we only call rename on a single server? 

Ideally for an identifier there should be only one LS providing a rename operation to make NO conflicts between the different LSs.

I really don't know how to deal with a situation when more than one LS returns non-null `prepareRename` result because for such a result received we'll need to show a new name selection dialogue on client side - not good if we'll have to show several dialogues (one for each LS returned non-null result).

Originally we use the only first `prepareRename` result to initiate the `rename` operation. 
The problem was that eventually we were getting null-result from Angular LS and gave up by that. This acquires all LSs, then takes the first non-null `prepareRename` result (if any) and initiates `rename` using that result.

> What do you mean by "fail" in "not just fail on first null-result received" ?

Currently from a list of LSs the only one (first) LS that is responded to `prepareRename` is used. In case of "Algular LS" and "TypeScript LS" the angular one wins almost all because it returns null-result. This ends the Rename operation with error `The Rename request is not valid at the given position`

vrubezhny:
On WWD side the issue can be tested using [this PR](https://github.com/eclipse/wildwebdeveloper/pull/1210)

vrubezhny:
> > In case of "Algular LS" and "TypeScript LS" the angular one wins almost all because it returns null-resu
> 
> So isn't it a bug in Angular LS that it returns a prepareRename while it's not preparing a rename?

I've opened https://github.com/angular/vscode-ng-language-service/issues/1893 for Angular LS, but it was refused to fix. 
Probably it's not an issue on Angular LS side, as it can provide `rename` capability for some angular-related stuff which cannot be processed by TypeScript LS, while returning null-`prepareRename` result if applied to an identifier of a variable or a function.

mickaelistria:
> I've opened https://github.com/angular/vscode-ng-language-service/issues/1893 for Angular LS, but it was refused to fix.

As I read it, the issue is more that the description didn't make very clear what is the bug in Angular LS; and it's not really clear what the bug in LSP4E would be either.
I see that Angular LS does respond to prepareRename, why couldn't it respond to rename? What is the request sent by LSP4E and the response from Angular LS? If the request looks OK but response is wrong, then we can tell it's a bug in Angular LS and reopen the issue.

vrubezhny:
In https://github.com/eclipse/wildwebdeveloper/issues/1209 a user tries  to rename a variable or a function, so two `prepareRename` requests are sent to Angular LS and TypeSript LS. 
Angular LS respond with `null` result:
```
[2023-05-23T19:32:53.975981187+02:00] LSP4E_TO_LANGUAGE_SERVER org.eclipse.wildwebdeveloper.angular:
{"jsonrpc":"2.0","id":"7","method":"textDocument/prepareRename","params":{"textDocument":{"uri":"file:///home/jeremy/projects/runtime-wwd-0.15/eslint/folding-test.ts"},"position":{"line":1,"character":11}}}

[2023-05-23T19:32:53.977718100+02:00] LANGUAGE_SERVER_TO_LSP4E org.eclipse.wildwebdeveloper.angular:
{"jsonrpc":"2.0","id":"7","result":null}
```
While TypeScript LS returns some regions:
```
[2023-05-23T19:32:53.975862626+02:00] LSP4E_TO_LANGUAGE_SERVER org.eclipse.wildwebdeveloper.jsts:
{"jsonrpc":"2.0","id":"14","method":"textDocument/prepareRename","params":{"textDocument":{"uri":"file:///home/jeremy/projects/runtime-wwd-0.15/eslint/folding-test.ts"},"position":{"line":1,"character":11}}}

[2023-05-23T19:32:53.993264853+02:00] LANGUAGE_SERVER_TO_LSP4E org.eclipse.wildwebdeveloper.jsts:
{"jsonrpc":"2.0","id":"14","result":{"start":{"line":1,"character":9},"end":{"line":1,"character":13}}}
```
Next happens the lsp4e problem - it takes the result only from Angular (`null`-result) and decides not to continue with `textDocument/rename` requests (even having a good non-`null` response from TypeScript LS)

vrubezhny:
Asked in Angular LS issue if `prepareRename`  capability is correctly reported by Angular LS:  https://github.com/angular/vscode-ng-language-service/issues/1893#issuecomment-1559891345

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/rename/LSPRenameProcessor.java
```diff
@@ -106,11 +107,14 @@ public RefactoringStatus checkInitialConditions(IProgressMonitor pm)
...
-				= LanguageServers.forDocument(document).withFilter(LSPRenameProcessor::isPrepareRenameProvider)
-				.computeFirst((w, ls) -> ls.getTextDocumentService().prepareRename(params).thenApply(result -> new Pair<>(w, result)))
-				.get(1000, TimeUnit.MILLISECONDS);
+			@SuppressWarnings("null")
+			Optional<Pair<LanguageServerWrapper, Either3<Range, PrepareRenameResult, PrepareRenameDefaultBehavior>>> tmp =
+					LanguageServers.forDocument(document).withFilter(LSPRenameProcessor::isPrepareRenameProvider)
+					.collectAll((w, ls) -> ls.getTextDocumentService().prepareRename(params)
+							.thenApply(result -> new Pair<>(w, result)))
+					.get(1000, TimeUnit.MILLISECONDS).stream()
+					.filter(Objects::nonNull).filter(t-> t.getSecond() != null)
```
mickaelistria, line 116: This filter is actually invalid: there are "good" LS that would return a Range or a PrepareRenameDefaultBehavior and for which `getSecond()` would be null, and now filtered out here, while the `rename` request can work.

mickaelistria (changes requested):

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/rename/LSPRenameProcessor.java
vrubezhny: Lost the answer... So, briefly repeating here...

Actually it's not. Whatever is returned by a LS is stored in `tmp.getSecond()` so, independently of  type of the returned object, the rename operation will be continued if it's not `null`.

vrubezhny:
> Asked in Angular LS issue if `prepareRename` capability is correctly reported by Angular LS: [angular/vscode-ng-language-service#1893 (comment)](https://github.com/angular/vscode-ng-language-service/issues/1893#issuecomment-1559891345)

@mickaelistria It looks like Angular LS is providing the `prepareRename`  and `rename` capability and operations
by configuring `prepareRenameProvider` and `renameProvider` capabilities if  `ServerHost provides native OS functionality`. 
The `prepareRename`  (https://github.com/angular/vscode-ng-language-service/blob/main/server/src/session.ts#L943) and `rename` (https://github.com/angular/vscode-ng-language-service/blob/main/server/src/session.ts#L906) handlers are also defined and used.

So, I suppose, yes - Angular LS is to provide rename support and together with TypeScript LS (in  https://github.com/eclipse/wildwebdeveloper/issues/1209) are two LSs we need to count on.

This PR allows to run a rename operation for the first LS that returns non-`null` result of `prepareRename` and in my opinion is worth to be merged and this will fix https://github.com/eclipse/wildwebdeveloper/issues/1209.

But then we need to decide on what should we do in case both LSs return non-`null` result and must participate in a rename operation.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/rename/LSPRenameProcessor.java
```diff
@@ -106,13 +108,17 @@ public RefactoringStatus checkInitialConditions(IProgressMonitor pm)
...
+					.collectAll((w, ls) -> ls.getTextDocumentService().prepareRename(params)
+							.thenApply(result -> new Pair<>(w, result)))
+					.get(1000, TimeUnit.MILLISECONDS);
 
-			Optional<Pair<LanguageServerWrapper, Either3<Range, PrepareRenameResult, PrepareRenameDefaultBehavior>>> tmp
-				= LanguageServers.forDocument(document).withFilter(LSPRenameProcessor::isPrepareRenameProvider)
-				.computeFirst((w, ls) -> ls.getTextDocumentService().prepareRename(params).thenApply(result -> new Pair<>(w, result)))
-				.get(1000, TimeUnit.MILLISECONDS);
+			Optional<Pair<LanguageServerWrapper, Either3<Range, PrepareRenameResult, PrepareRenameDefaultBehavior>>> tmp = list
+					.stream().filter(Objects::nonNull).filter(t -> t.getSecond() != null).findFirst();
```
mickaelistria, line 119: OK, I see the `getSecond` just match the condition that used to be below, after `tmp.isEmpty()`, so this code is indeed an improvement over current one.
I'm still unsure about only checking `getSecond`, but it's a different issue I guess.

mickaelistria (approved):

Issue #644, opened 2023-05-16
Opening/closing editors leave a lot of Stop Language Server threads


rubenporras:
Opening/closing editors leave a lot of Stop Language Server threads, see the screenshot with lsp4e configured with             lastDocumentDisconnectedTimeout="360000"

![image](https://github.com/eclipse/lsp4e/assets/43636626/4e2e1e1f-5ad9-4b1a-a8a9-707851426f23)

There are several pages to scroll in the debug window.

PR #656, opened 2023-05-23
LS start/stop improvement (based on LSP4IJ)

angelozerr:
I have not tested this draft PR but I tried to set all update that I did for the LSP4IJ (LSP for IJ) which is a port of LSP4E.

I created this PR to start some discussion.

### DIFF
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageClientImpl.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageClientImpl.java
@@ -51,6 +51,7 @@ public class LanguageClientImpl implements LanguageClient {
 
 	private LanguageServer server;
 	private LanguageServerWrapper wrapper;
+	private boolean disposed;
 
 	public final void connect(LanguageServer server, LanguageServerWrapper wrapper) {
 		this.server = server;
@@ -156,4 +157,22 @@ public class LanguageClientImpl implements LanguageClient {
 			return new ShowDocumentResult(true);
 		});
 	}
+
+	/**
+	 * Dispose language client.
+	 */
+	public void dispose() {
+		progressManager.dispose();
+		disposed = true;
+	}
+
+	/**
+	 * Returns true if the client is disposed and false otherwise.
+	 *
+	 * @return true if the client is disposed and false otherwise.
+	 */
+	public boolean isDisposed() {
+		return disposed;
+	}
+
 }
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServerWrapper.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServerWrapper.java
@@ -426,6 +426,9 @@ public class LanguageServerWrapper {
 			return;
 		}
 		removeStopTimer();
+		if (this.languageClient != null) {
+            this.languageClient.dispose();
+        }
 		if (this.initializeFuture != null) {
 			this.initializeFuture.cancel(true);
 			this.initializeFuture = null;
@@ -451,12 +454,12 @@ public class LanguageServerWrapper {
 				}
 			}
 
-			if (serverFuture != null) {
-				serverFuture.cancel(true);
-			}
-
 			if (languageServerInstance != null) {
 				languageServerInstance.exit();
+			}
+
+			if (serverFuture != null) {
+				serverFuture.cancel(true);
 			}
 
 			if (provider != null) {
@@ -471,7 +474,7 @@ public class LanguageServerWrapper {
 		this.lspStreamProvider = null;
 
 		while (!this.connectedDocuments.isEmpty()) {
-			disconnect(this.connectedDocuments.keySet().iterator().next());
+			disconnect(this.connectedDocuments.keySet().iterator().next(), true);
 		}
 		this.languageServer = null;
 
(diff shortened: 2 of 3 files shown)


org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServerWrapper.java
```diff
@@ -426,6 +426,9 @@ public synchronized void stop() {
 			return;
 		}
 		removeStopTimer();
+		if (this.languageClient != null) {
+            this.languageClient.dispose();
```
angelozerr, line 430: In IJ usecase the LanguageClient implementation should add listener and we need a dispose to remove it.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServerWrapper.java
```diff
@@ -451,14 +454,14 @@ public synchronized void stop() {
...
 
-			if (serverFuture != null) {
-				serverFuture.cancel(true);
-			}
-
 			if (languageServerInstance != null) {
 				languageServerInstance.exit();
 			}
 
+			if (serverFuture != null) {
```
angelozerr, line 461: I move this serverFuture cancel to avoid having Piped stream closing error when exit() is called.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServerWrapper.java
```diff
@@ -471,7 +474,7 @@ public synchronized void stop() {
 		this.lspStreamProvider = null;
 
 		while (!this.connectedDocuments.isEmpty()) {
-			disconnect(this.connectedDocuments.keySet().iterator().next());
+			disconnect(this.connectedDocuments.keySet().iterator().next(), true);
```
angelozerr, line 477: On the stop all file must be disconnected and should never recall stop or restartStopTimer.

@rubenporras perhaps this disconnect method with stopping parameter could fix your issue https://github.com/eclipse/lsp4e/issues/644 ?

angelozerr:
@mickaelistria @rubenporras I have stressed my LSP4IJ language server (by stopping ,starting several ls, etc) and sometimes I have an error with Piped stream which is closed. When I'm in this context, my LanguageServerWrapper starts to be crazy it start and stop just after everytime.

I suspect the problem comes that launcherFuture is canceled in async mode https://github.com/eclipse/lsp4e/blob/8d85108be1500172fc909fadb8ccd6c8c4243f0f/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServerWrapper.java#L443

The server start its is not active, the stop is called and launcherFuture is recreate, BUT the runAsynch cancel this launcherFuture again.

So my question is why shutdown,launcherFuture , etc are done in runAsync and not without runAsync?

sebthom:
LGTM This could also fix an issue I am sometimes seeing with the Dart Language Server where endlessly running progress bars in the progress view pile up.

sebthom:
@angelozerr the build fails with:

```
[ERROR] Failed to execute goal org.eclipse.tycho:tycho-compiler-plugin:3.0.4:compile (default-compile) on project org.eclipse.lsp4e: Compilation failure: Compilation failure: 
[ERROR] /home/jenkins/agent/workspace/lsp4e-github_PR-656/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageClientImpl.java:[164] 
[ERROR] 	progressManager.dispose();
[ERROR] 	                ^^^^^^^
[ERROR] The method dispose() is undefined for the type LSPProgressManager
[ERROR] 1 problem (1 error)
```

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageClientImpl.java
```diff
@@ -156,4 +156,11 @@ public CompletableFuture<ShowDocumentResult> showDocument(ShowDocumentParams par
 			return new ShowDocumentResult(true);
 		});
 	}
+
+	/**
+	 * Dispose language client.
+	 */
+	public void dispose() {
+		progressManager.dispose();
```
sebthom, line 164: Nice find! This could also fix an issue I am sometimes seeing with the Dart Language Server where endlessly running progress bars in the progress view pile up. Would be nice if this could make it into the upcoming release.

angelozerr:
@sebthom please note that I have not spend time to test my code, because it is hard to test with LSP4E since there is not LSP console which shows the language server process (started, stopped) since it is not possible to start / stop language server with an action menu contextual like I have in LSP4IJ.

As I have the same code thanLSP4E I created the PR, but an update of LanguageServerWrapper requires to be very carefull.

The thing that I don't understand is the call of shutdown and launcherFure cancel in an runAsync which could provide some trouble with started/stopped, started stopped lifecycle of language server and/or Piped stream closed

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageClientImpl.java
angelozerr: To be honnest with you, I have not tested andI did that just to find a usecase with dispose in LSP4E. In my case, my language server which implements LanguageClientImpl (in IJ) override dispose method to remove some project listeners added in the constructor.

angelozerr:
> @angelozerr the build fails with:

Ooops sorry I missed to push LSPProgressManager, I pushed it but I repeat me, I have not tested

mickaelistria (changes requested):
In general, the fact that the IJ clone of LSP4E has some issue doesn't imply that LSP4E has some issues. Both already have and will diverge, we won't backport "blindly" changes from LSP4IJ into LSP4E, we need to get insights about the actual issue it fixes for LSP4E.
So please refine the commit message to explicit what issue is fixed and describe how this patch intends to fix it. Also, I see some non-backward compatible change (LanguageServerWrapper#disconnect() has a different signature), which I don't think is profitable.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServerWrapper.java
rubenporras: I have tested the change and I see the same exception patterns with and without when running the tests.

basilevs: @rubenporras see [my bug report](https://github.com/eclipse/lsp4e/issues/681) - on Eclipse shutdown this code is not executed, but results in same error, making reproduction attempts confusing.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageClientImpl.java
rubenporras: I have taken over this part of the PR in https://github.com/eclipse/lsp4e/pull/705 and plan to work on it.

angelozerr: Thanks @rubenporras !

rubenporras:
I most of them but not all. I think we were unsure about a minor part of the PR. I agree that it is probably best to close the PR, as we are unsure but most likely we have the most important bits merged.

akurtakov:
Closing in this case. Please note that every open PR gets rebuild after every push to master thus the number of open PRs really matter for the speed of feedback.

PR #661, opened 2023-05-25
Fix "mark occurrences" not working because of NPE

sebthom:
I realized that "mark occurrences" was not working (at least with the Haxe and Dart Language Servers).

The reason are silently swallowed NPEs in the HighlightReconcilingStrategy. Adding two null guards restored the functionality.


org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/highlight/HighlightReconcilingStrategy.java
```diff
@@ -285,7 +289,10 @@ void removeOccurrenceAnnotations() {
 		}
 	}
 
-	private String kindToAnnotationType(DocumentHighlightKind kind) {
+	private String kindToAnnotationType(@Nullable DocumentHighlightKind kind) {
+		if (kind == null)
```
rubenporras, line 293: is this specified by the protocol?

sebthom: Yes.

image

PR #670, opened 2023-05-31
Perform updates to Stop Language Server Timer synchronously.

andrewL-avlq:
Multiple Timer threads can be left running if
LanguageServerWrapper.startStopTimer() runs in parallel with LanguageServerWrapper.removeStopTimer() or another startStopTimer(). We prevent this issue by reducing the Timer threads created down to one, and then only scheduling or cancelling TimerTasks on that Timer thread synchronously.

This should fix issue #644.


rubenporras (approved):
Thanks

PR #671, opened 2023-06-01
fix: adhere to lsp spec when calculating offsets

### LINKED ISSUE #669: applyEdit(s) is not conformant with the LSP spec
The LSP spec states for Ranges:

> A range in a text document expressed as (zero-based) start and end positions. A range is comparable to a selection in an editor. Therefore, the end position is exclusive. If you want to specify a range that contains a line including the line ending character(s) then use an end position denoting the start of the next line.

LSPEclipseUtils calls `toOffset` like this:
```
int length = LSPEclipseUtils.toOffset(textEdit.getRange().getEnd(), document) - offset;
```

In LSPEclipseUtils.toOffset:
```
public static int toOffset(Position position, IDocument document) throws BadLocationException {
  return document.getLineInformation(position.getLine()).getOffset() + position.getCharacter();
}
```

When trying to include the last line completely in a given range, if the number of lines in the document is 100, 101 should be accepted according to protocol. But this, with the current implementation lets JFace ListLineTracker throw a BadLocationException.

To fix this, document.getLineInformation() should give `Math.min(position.getLine(), document.length` (pseudo-code). One could also check if the end position char is 0, but the result most likely would be the same.

bastiandoetsch:
Fixes #669

### DIFF
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageClientImpl.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageClientImpl.java
@@ -115,7 +115,9 @@ public class LanguageClientImpl implements LanguageClient {
 			final var job = new Job(Messages.serverEdit) {
 				@Override
 				public IStatus run(IProgressMonitor monitor) {
-					LSPEclipseUtils.applyWorkspaceEdit(params.getEdit(), params.getLabel());
+					UI.getDisplay().syncExec(() -> {
+						LSPEclipseUtils.applyWorkspaceEdit(params.getEdit(), params.getLabel());
+					});
 					return Status.OK_STATUS;
 				}
 			};
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageClientImpl.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageClientImpl.java
@@ -115,9 +115,7 @@ public class LanguageClientImpl implements LanguageClient {
 			final var job = new Job(Messages.serverEdit) {
 				@Override
 				public IStatus run(IProgressMonitor monitor) {
-					UI.getDisplay().syncExec(() -> {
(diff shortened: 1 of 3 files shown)


mickaelistria:
Thanks. Can you please try to add a test case covering this aspect of the spec?

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageClientImpl.java
```diff
@@ -115,7 +115,9 @@ public final CompletableFuture<ApplyWorkspaceEditResponse> applyEdit(ApplyWorksp
 			final var job = new Job(Messages.serverEdit) {
 				@Override
 				public IStatus run(IProgressMonitor monitor) {
-					LSPEclipseUtils.applyWorkspaceEdit(params.getEdit(), params.getLabel());
+					UI.getDisplay().syncExec(() -> {
+						LSPEclipseUtils.applyWorkspaceEdit(params.getEdit(), params.getLabel());
+					});
```
mickaelistria, line 120: Does this have to be part of this patch?

bastiandoetsch: Yeah, else you get InvalidThreadAccess when `workspace/applyEdit` is sent.

bastiandoetsch: The alternative would be to change `org.eclipse.lsp4e.LSPEclipseUtils.applyWorkspaceEditIfSingleOpenFile(WorkspaceEdit)`

bastiandoetsch: That said, it could be part of another PR/commit, but the bug we were running into is only fixed if it's done, too.

bastiandoetsch: I've reverted that change

mickaelistria:
Thanks!

PR #672, opened 2023-06-01
feat: implement codelens refresh message from server

bastiandoetsch:
Implements https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#codeLens_refresh


mickaelistria:
This is marked as draft. Is there further work to be done before we can review/merge?

bastiandoetsch:
Haven't written a test yet ;). But else it's mergeable :)

bastiandoetsch:
Codelenses in general are problematic in LSP4e. They don't appear for already open editors when starting a language server (I suppose it's related to the check in the codeminer in the plugin.xml), and also aren't refreshed on save. But this PR is not about fixing that :).

mickaelistria:
OK, let's wait for a test if you plan to add one then.

ahmedneilhussain:
> Codelenses in general are problematic in LSP4e. They don't appear for already open editors when starting a language server (I suppose it's related to the check in the codeminer in the plugin.xml), and also aren't refreshed on save. But this PR is not about fixing that :).

Is that so? I thought so too (except I observed it as code lenses missing when eclipse started up with documents already open because of workbench state) but after trying to reproduce it I discovered it not to be the case: they appear *eventually* after the language server has initialised and scanned everything. In our case I had one or two massive projects open and so it took the backend quite a long time to plough through them all and eventually send code lenses to the front end.

It might depend on the back end implementation, though?

bastiandoetsch:
Interesting, I'll look into that @ahmedneilhussain . @mickaelistria , looking at the source again, a test would be either extremely complicated or almost completely mocking. As it's in the end only one call (after getting the sourceviewer), I think should just merge it as is.

angelozerr:
@bastiandoetsch you need to fill CodeLensWorkspaceCapabilities#refreshSupport to true to notify to the language server that LSP4E supports this new feature.

In Qute LS (language server) we use this refreshSupport to know if we can use it to refresh codelens.

angelozerr:
> @angelozerr done.

Thanks so much!

I wonder if you could do the same thing for InlayHint since CodeLens and InlayHint use CodeMining.

angelozerr:
I mean supporting https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#workspace_inlayHint_refresh

mickaelistria:
Thanks for the patch!

PR #679, opened 2023-06-02
[nobug] Bump o.e.lsp4e verion to 0.17.0 due to API-breaking changes

vrubezhny:


rubenporras (approved):
Thanks, this makes sense.

PR #680, opened 2023-06-02
Removal of all APIs that expose the language server object.

rubenporras:
Removal of all APIs that expose the language server object.

The test LanguageServiceAccessorTest.testStatusHandlerLSAsRunConfiguration
has been deleted as it cannot be reliably tested, it depends on creation
of the stream provider in LanguageServerWrapper#start, which happens in
a background thread.

The method LanguageServerWrapper#getVersion has been renamed and made public as a
replacement of the marked for removal class `LSPDocumentInfo`, which can be used to
indirectly get access to the until now private method.

### DIFF
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServerWrapper.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServerWrapper.java
@@ -975,7 +975,10 @@ public class LanguageServerWrapper {
 		}
 	}
 
-	int getVersion(URI uri) {
+	/**
+	 * return the TextDocument version, suitable to build a TextDocumentIndentifier
+	 */
+	public int getTextDocumentVersion(URI uri) {
 		DocumentContentSynchronizer documentContentSynchronizer = connectedDocuments.get(uri);
 		if (documentContentSynchronizer != null) {
 			return documentContentSynchronizer.getVersion();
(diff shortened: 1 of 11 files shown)


rubenporras:
Hi @BoykoAlex , thanks for the hint, I can do that. I will prepare a new version of the PR tomorrow if there are no other opinions.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServerWrapper.java
```diff
@@ -975,7 +975,7 @@ void unregisterCommands(List<String> cmds) {
 		}
 	}
 
-	int getVersion(URI uri) {
+	public int getVersion(URI uri) {
```
mickaelistria, line 978: Before making it a public API, I think we need clear documentation, and/or clear name, about which version we're mentioning here. Is it the one for LSP's TextDocument object, or the one for the underlying Eclipse IDocument? I wold suggest
```java
/**
 * return the TextDocument version, suitable to build a TextDocumentIndentifier
 */
public int getTextDocumentVersion(URI uri) {
  ...
}
```,
Or maybe returning a `VersionedTextDocumentIdentifier` here, to clarify what it's meant to be used for.

rubenporras:
I have addressed the comments. If we still want to change the name of the new method, I am happy to it in a follow up.

PR #685, opened 2023-06-06
Error reporting for failed quick fix

BoykoAlex:
Show error dialog if quick fix failed on the server side.


org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/codeactions/CodeActionMarkerResolution.java
```diff
@@ -92,8 +95,16 @@ public void run(IMarker marker) {
...
 					if (provider != null && provider.getCommands().contains(command.getCommand())) {
+						final String serverId = wrapper.serverDefinition.id;
 						wrapper.execute(ls -> ls.getWorkspaceService()
-								.executeCommand(new ExecuteCommandParams(command.getCommand(), command.getArguments())));
+								.executeCommand(new ExecuteCommandParams(command.getCommand(), command.getArguments()))).exceptionally(t -> {
+									String errorTitle = "Error Executing Quick Fix"; //$NON-NLS-1$
+									String errorMessage = "Failed to fetch quick fix edit for '" + codeAction.getTitle() //$NON-NLS-1$
+											+ "'. See Language Server '" + serverId + "' log for more details."; //$NON-NLS-1$ //$NON-NLS-2$
+									Display display = PlatformUI.getWorkbench().getDisplay();
+									display.asyncExec(() -> MessageDialog.openError(display.getActiveShell(), errorTitle, errorMessage));
```
rubenporras, line 105: I think we should rather use ServerMessageHandler.showMessage, as in other places in the codebase. Could you do that?

rubenporras (changes requested):

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/codeactions/CodeActionMarkerResolution.java
mickaelistria, line 105: If you want to have fun with newer constructs:
```java
  .execute(ls -> ls.getWorkspaceService()
     .executeCommand(...)
     .exceptionally(this::constructErrorMessages()) // could be a Pair<String, String> or an Entry<String, String>
     .thenAcceptAsync(messages -> MessageDialog.openError(Display.getCurrent(), messages.getKey(), messages.getValue()), PlatformUI.getWorkbench().getDisplay())); // Display is now an Executor that can be used to "host" future chains
```

BoykoAlex:
@rubenporras @mickaelistria I have update the PR to address your comment. Thanks for the review :-) (Couldn't quite get the nice syntax for the future)

mickaelistria:
Thanks!

Issue #703, opened 2023-06-21
LSSymbolsContentProvider populating the outline view freezes UI for minutes for large files.


sebthom:
Opening large json files results in UI freezes sometimes lasting for minutes caused by the `org.eclipse.lsp4e.outline.LSSymbolsContentProvider`. Maybe the number populated to the outline should limited.

```java
"main" #1 prio=6 os_prio=0 cpu=77390.62ms elapsed=1488.10s tid=0x000002b98cb78d70 nid=0x817c runnable  [0x000000ce59afc000]
   java.lang.Thread.State: RUNNABLE
        at org.eclipse.swt.internal.win32.OS.CallWindowProc(Native Method)
        at org.eclipse.swt.widgets.Tree.callWindowProc(Tree.java:1576)
        at org.eclipse.swt.widgets.Control.windowProc(Control.java:4844)
        at org.eclipse.swt.widgets.Tree.windowProc(Tree.java:6142)
        at org.eclipse.swt.widgets.Display.windowProc(Display.java:5022)
        at org.eclipse.swt.internal.win32.OS.SendMessage(Native Method)
        at org.eclipse.swt.widgets.Tree.createItem(Tree.java:2129)
        at org.eclipse.swt.widgets.TreeItem.<init>(TreeItem.java:180)
        at org.eclipse.swt.widgets.TreeItem.<init>(TreeItem.java:86)
        at org.eclipse.jface.viewers.TreeViewer.createNewRowPart(TreeViewer.java:767)
        at org.eclipse.jface.viewers.TreeViewer.newItem(TreeViewer.java:279)
        at org.eclipse.jface.viewers.AbstractTreeViewer.createTreeItem(AbstractTreeViewer.java:863)
        at org.eclipse.jface.viewers.AbstractTreeViewer.updateChildren(AbstractTreeViewer.java:2813)
        at org.eclipse.jface.viewers.AbstractTreeViewer.internalRefreshStruct(AbstractTreeViewer.java:1971)
        at org.eclipse.jface.viewers.TreeViewer.internalRefreshStruct(TreeViewer.java:684)
        at org.eclipse.jface.viewers.AbstractTreeViewer.internalRefresh(AbstractTreeViewer.java:1947)
        at org.eclipse.jface.viewers.AbstractTreeViewer.internalRefresh(AbstractTreeViewer.java:1904)
        at org.eclipse.ui.navigator.CommonViewer.internalRefresh(CommonViewer.java:532)
[...]

PR #704, opened 2023-06-21
Support limiting the number of outline symbols to avoid UI freezes

sebthom:
This PR addresses #703 by limiting the maximum number of elements rendered in the UI to 5000 elements.

On my workstation Eclipse can add/render around 1.000 elements per second to the outline. Since adding elements is done in the UI thread, more elements mean longer freeze on initial population and each update.

### DIFF
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/outline/LSSymbolsContentProvider.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/outline/LSSymbolsContentProvider.java
@@ -63,6 +63,7 @@ import org.eclipse.ui.navigator.ICommonContentExtensionSite;
 import org.eclipse.ui.navigator.ICommonContentProvider;
 import org.eclipse.ui.progress.PendingUpdateAdapter;
 import org.eclipse.ui.texteditor.ITextEditor;
+import org.eclipse.ui.views.WorkbenchViewerSetup;
 
 public class LSSymbolsContentProvider implements ICommonContentProvider, ITreeContentProvider {
 
@@ -256,6 +257,10 @@ public class LSSymbolsContentProvider implements ICommonContentProvider, ITreeCo
 		}
 
 		this.viewer = (TreeViewer) viewer;
+		
+		// this enables limiting the number of outline entries to mitigate UI freezes
+		WorkbenchViewerSetup.setupViewer(this.viewer);
+
 		isQuickOutline = Boolean.TRUE.equals(viewer.getData(VIEWER_PROPERTY_IS_QUICK_OUTLINE));
 
 		outlineViewerInput = (OutlineViewerInput) newInput;


mickaelistria (changes requested):
Let's just wait for https://github.com/eclipse-platform/eclipse.platform.ui/pull/810

sebthom:
@mickaelistria another issue I realized is that the Language Server also takes a considerable amount of time to provide the list of outline symbols for large files. is there a way to tell the LS to limit the number of elements returned in that request?

https://github.com/eclipse/lsp4e/blob/9ad25d4828c1041014993695fa5d1d2684510fed/org.eclipse.lsp4e/src/org/eclipse/lsp4e/outline/LSSymbolsContentProvider.java#L316-L317

I can see that DocumentSymbolParams extends WorkDoneProgressAndPartialResultParams but it is not clear if the decision to return partial results can only be made by the server or also requested by the client.

mickaelistria:
I don't see anything that fits immediately in https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/ . I suggest you lookup in the issues reported against LSP, IIRC there was some discussion about "paginated" responses which could fit here.

angelozerr:
In LemMinx we have exactly this kind of problem (when we integrate LemMinx in vscode). It is the reason why we provide a custom limit for symbol.

> I don't see anything that fits immediately in https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/ .

Indeed but folding specifies rangeLimit, see 
https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_foldingRange

So it can be a good idea to have too a rangeLimit kind for symbols.

Please note that symbol is used in breadcrumb in vscode, so I think symbols request should perhaps send too an information about the cursor location(just to give symbols hierachy for the cursor location).

mickaelistria:
Overall, LSP and more particularly JSON-RPC may be missing some capacity for sending responses in (ordered) chunks so that clients such as LSP4J could get the output as a stream or returning potentially long sequences in a BlockingQueue... But this problem is deeply rooted in JSon-RPC/LSP and should rather be addressed there.

akurtakov:
Is this still needed after platform.ui change?

sebthom:
A code change is still necessary because the platform.ui change is not enabled by default for all tables/tree viewers.

However this change does only address one part of the UI freeze. With newer lsp4j releases the following issue popped up eclipse-lsp4j/lsp4j#815

ghentschke:
I ran into performance issues as well. I think this PR is a good starting point. Can this be merged @mickaelistria ?

rubenporras:
Hi @ghentschke,  but would the problem not also greatly improved (also increasing performance in all cases) if we would fix https://github.com/eclipse/lsp4e/issues/907 without using experimental APIs? 

PD: If you agree to merge this PR, that is fine for me as well, regardless of what is the answer to my question.

sebthom:
> without using experimental APIs?

While the javadoc says it is experimental, it is actively used in eclipse.platform.ui and eclipse.jdt.ui so I doubt it is so experimental.

ghentschke:
IMO a better solution would  be lazy loading. Fetch the data in the background and update the treeview when the user scrolls. But this PR is a good starting to point to keep eclipse responsive.

sebthom:
> IMO a better solution would be lazy loading. Fetch the data in the background and update the treeview when the user scrolls. But this PR is a good starting to point to keep eclipse responsive.

I also suggested that in the platform issue but they somehow found the current solution better.

mickaelistria (approved):

ghentschke:
The next step would be to refactor the `org.eclipse.lsp4e.outline.LSSymbolsContentProvider` to support partial results as described [here](https://github.com/eclipse-lsp4j/lsp4j/issues/815)

PR #705, opened 2023-06-30
Close monitors tied to a LS notifications when the LS stops

rubenporras:


org.eclipse.lsp4e/src/org/eclipse/lsp4e/progress/LSPProgressManager.java
```diff
@@ -123,6 +125,7 @@ private void createJob(final LinkedBlockingDeque<ProgressParams> queue, final St
 			}
 		});
 		job.schedule();
+		jobs.add(job);
```
mickaelistria, line 128: The new job should have a listener to remove itself from the "jobs" when it's done.

rubenporras: Thanks, I have done the change.

PR #706, opened 2023-06-30
fix erratic ConcurrentModificationException during unit tests

sebthom:
This PR fixes an erratic  ConcurrentModificationException that results in random test failures.

```java
java.util.ConcurrentModificationException
  at java.base/java.util.WeakHashMap.forEach(WeakHashMap.java:1031)
  at org.eclipse.lsp4e.ConnectDocumentToLanguageServerSetupParticipant.waitForAll(ConnectDocumentToLanguageServerSetupParticipant.java:67)
  at org.eclipse.lsp4e.test.AllCleanRule.clear(AllCleanRule.java:66)
  at org.eclipse.lsp4e.test.AllCleanRule.finished(AllCleanRule.java:58)
  at org.junit.rules.TestWatcher.finishedQuietly(TestWatcher.java:122)
```

### DIFF
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/ConnectDocumentToLanguageServerSetupParticipant.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/ConnectDocumentToLanguageServerSetupParticipant.java
@@ -11,6 +11,8 @@
  *******************************************************************************/
 package org.eclipse.lsp4e;
 
+import java.util.Collections;
+import java.util.Map;
 import java.util.WeakHashMap;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ExecutionException;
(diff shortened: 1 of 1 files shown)


org.eclipse.lsp4e/src/org/eclipse/lsp4e/ConnectDocumentToLanguageServerSetupParticipant.java
```diff
@@ -34,7 +36,7 @@
  *
  */
 public class ConnectDocumentToLanguageServerSetupParticipant implements IDocumentSetupParticipant, IDocumentSetupParticipantExtension {
-	private static final WeakHashMap<CompletableFuture<?>, Void> PENDING_CONNECTIONS = new WeakHashMap<>();
+	private static final Cache<CompletableFuture<?>, Void> PENDING_CONNECTIONS = CacheBuilder.newBuilder().weakKeys().build();
```
rubenporras, line 39: Would it not be more standard to wrap it as in `Collections.synchronizedMap(new WeakHashMap<>())`?

I have checked the CacheBuilder and I think in the end it will something very similar, but it is harder to understand, as one needs to know Guava, which is an extra API on top of standard Java.

rubenporras (commented):
Thanks for the PR. This failed indeed in my last PR test execution and I was wondering as well what it could be.

rubenporras (approved):
Thanks

PR #722, opened 2023-07-05
Do not propagate exceptions to the UI as outline elements

rubenporras:


rubenporras:
Fixes https://github.com/eclipse/lsp4e/issues/278

mickaelistria:
Why not #722 (and cancellationexception only)? It seems important that user can make the difference between an empty outline and a bug. Maybe instead of the details exception, we could add a note "An error occured, see log for details" in such case.

rubenporras:
Okay, I will add that in case of an empty outline. If the outline is not empty, I think it is better to show the elements which are present, instead of showing an exception, that would flickering for spurious errors

rubenporras:
I am merging to mark it as done in my todo list,  if there is some more comments, I would address them as a followup.

PR #723, opened 2023-07-05
Remove map which is always empty

rubenporras:


sebthom (approved):

sebthom:
I stumbled on this myself and wondered if there is just some code missing and the map is actually useful for something. But as the saying in conventional medicine goes "If in doubt, cut it out" :)

mickaelistria (changes requested):
If I read it correctly, this issue highlights that `getParent` is never returning the right value, is it?
I would qualify it as a bug, worth being tracked in an issue first, as `getParent` is necessary for some things like expand or other operations on the viewer to work properly. Instead of getting rid of the map, we should make getParent work.

mickaelistria:
Please open an issue regarding `getParent` not working.

rubenporras:
I have created https://github.com/eclipse/lsp4e/issues/724, I can revert the commit if you like but I would not know how to fix the code (if so I would have done it myself already) and reverting will not fix anything. If you know what the code should do, could you provide the fix?

sebthom:
I think as long as we don't have a broken use/test case it makes no sense to revert the commit and reintroduce dead code.

mickaelistria:
I agree reverting wouldn't help. What's interesting is keeping track of a bug from a bad code smell; keeping the bad smell one we have the bug tracked doesn't bring anything good.

PR #731, opened 2023-07-14
Add LS console name

### LINKED ISSUE #730: Name Language Server Console
Currently the LS console name is the plugin id. IMO it should get a proper name that user can identify, since the console log is configurable in the prefs:
Old: 
image

Proposal: 
image

ghentschke:
fixes #730

### DIFF
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LoggingStreamConnectionProviderProxy.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LoggingStreamConnectionProviderProxy.java
@@ -28,6 +28,7 @@ import org.eclipse.core.runtime.IPath;
 import org.eclipse.jdt.annotation.Nullable;
 import org.eclipse.jface.preference.IPreferenceStore;
 import org.eclipse.lsp4e.server.StreamConnectionProvider;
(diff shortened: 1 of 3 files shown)


org.eclipse.lsp4e/src/org/eclipse/lsp4e/LoggingStreamConnectionProviderProxy.java
```diff
@@ -52,6 +52,7 @@ public static File getLogDirectory() {
 
 	private static final String FILE_KEY = "file.logging.enabled"; //$NON-NLS-1$
 	private static final String STDERR_KEY = "stderr.logging.enabled"; //$NON-NLS-1$
+	private static final String LS_CONSOLE = "LS Console"; //$NON-NLS-1$
```
mickaelistria, line 55: It is is a user-visible label, it should be internationalized, using Messages/messages.properties.

mickaelistria: Also, I think naming it "Language Servers" is more explicit to anyone, and that we don't need much to repeat console if we're already in a Console tab.

mickaelistria:
Thanks!

PR #733, opened 2023-07-15
Add basic type hierarchy view

### LINKED ISSUE #732: Add Type Hierarchy View
As a programmer I want to use the Type Hierarchy view when working with a LSB based editor. 

At the end it could similar to the Java Type Hierarchy in Eclipse: 
image

ghentschke:
This is a basic implementation of a LSP based type hierarchy view with the same info as in the existing type hierarchy dialog. In the next step the view could be expanded with further filters and a member info box.

Using the default Eclipse keys: F4 for the Type Hierarchy view and Ctrl+T or Command+T (Mac) for the Quick Type Hierarchy

Currently it looks like this: 
image
image

fixes #732


ghentschke:
Maybe we should consider to put all views stuff (type and call) in the `org.eclipse.lsp4e.ui.views` package?

ghentschke:
IMO we should support the default Eclipse keys for Type Hierarchy: F4 for the view and Ctrl+T for the "Quick Type Hierarchy" which is currently mapped to F4.

WDYT?

mickaelistria:
> F4 for the view and Ctrl+T for the "Quick Type Hierarchy" which is currently mapped to F4.
 
OK.

org.eclipse.lsp4e/plugin.xml
```diff
@@ -397,6 +430,12 @@
             schemeId="org.eclipse.ui.defaultAcceleratorConfiguration"
             sequence="F4">
       </key>
+      <key
+            commandId="org.eclipse.lsp4e.openTypeHierarchy"
+            contextId="org.eclipse.ui.genericeditor.genericEditorContext"
+            schemeId="org.eclipse.ui.defaultAcceleratorConfiguration"
+            sequence="Ctrl+Alt+T">
```
mickaelistria, line 437: Needs to be something like `M1+M2+T` to work with mac as well

org.eclipse.lsp4e/plugin.xml
```diff
@@ -195,6 +202,10 @@
                   definitionId="org.eclipse.lsp4e.textSelectionHasLanguageServer">
             </reference>
          </activeWhen>
+      </handler>
+      <handler
```
mickaelistria, line 206: This hanndler should be enabled only when current selection or editor can be assisted by language server.

org.eclipse.lsp4e/plugin.xml
```diff
@@ -329,6 +340,52 @@
             </visibleWhen>
          </command>
       </menuContribution>
+      <menuContribution
+            locationURI="popup:#TextEditorContext?before=org.eclipse.ui.genericeditor.findReferences">
+         <command
+               commandId="org.eclipse.lsp4e.typeHierarchy">
+            <visibleWhen
```
mickaelistria, line 347: If you properly configure `enabledWhen` on the handler, then it can be replaced with `visibleWhen checkEnabled="true"`

org.eclipse.lsp4e/plugin.xml
```diff
@@ -329,6 +340,52 @@
...
+                  </with>
+               </and>
+            </visibleWhen>
+         </command>
+      </menuContribution>
+      <menuContribution
+            locationURI="popup:#TextEditorContext?before=org.eclipse.ui.genericeditor.findReferences">
+         <command
+               commandId="org.eclipse.lsp4e.openTypeHierarchy">
+            <visibleWhen
```
mickaelistria, line 370: same

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LSPEclipseUtils.java
```diff
@@ -333,7 +334,12 @@ public static CallHierarchyPrepareParams toCallHierarchyPrepareParams(int offset
 		Position position =  LSPEclipseUtils.toPosition(offset, document);
 		TextDocumentIdentifier documentIdentifier = toTextDocumentIdentifier(document);
 		return new CallHierarchyPrepareParams(documentIdentifier, position);
+	}
 
+	public static TypeHierarchyPrepareParams toTypeHierarchyPrepareParams(int offset, final @NonNull IDocument document) throws BadLocationException {
```
mickaelistria, line 339: If this is only used in TypeHierarchyContentProvider, let's just move it there. LSPEclipseUtils is more meant for reusable bits that several operations require.

mickaelistria:
Seems all good now. Thanks!

PR #740, opened 2023-07-31
[#735] Add member info to Type Hierarchy view

### LINKED ISSUE #735: Add member info to Type Hierarchy view
As a user I want to experience similar editing comfort when working with a LSP based text editor. That's why the LSP based Type Hierarchy view should provide similar info as the existing Type Hierarchy View in Eclipse for Java and other languages:

In the first step there should be a TableView containing the members/symbols of the selected class:
image

The symbols should be updated when the associated document has been modifed. A double click on the symbol will open its declaration in the corresponding document.

I am working on a PR

ghentschke:
- show member of selected type in separate table viewer
- table viewer will be updated when the associated document has been modified
- Doubleclick on a member opens its declaration in the editor

fixes #735


org.eclipse.lsp4e/src/org/eclipse/lsp4e/FileBufferListenerAdapter.java
```diff
@@ -15,7 +15,7 @@
 import org.eclipse.core.filebuffers.IFileBufferListener;
 import org.eclipse.core.runtime.IPath;
 
-class FileBufferListenerAdapter implements IFileBufferListener {
+public class FileBufferListenerAdapter implements IFileBufferListener {
```
sebthom, line 18: Maybe instead of now exposing FileBufferListenerAdapter as part of the plugin's API we should move it to the internal package.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/typeHierarchy/TypeHierarchyViewContentProvider.java
```diff
@@ -42,18 +44,15 @@
 public class TypeHierarchyViewContentProvider implements ITreeContentProvider {
 	private TreeViewer treeViewer;
 	private LanguageServerWrapper languageServerWrapper;
-	private List<TypeHierarchyItem> hierarchyItems;
+	private List<TypeHierarchyItem> hierarchyItems = new ArrayList<>();
```
sebthom, line 47: I think this can be replaced with `= Collections.emptyList();`

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/typeHierarchy/TypeHierarchyViewContentProvider.java
```diff
@@ -90,14 +89,12 @@ public boolean hasChildren(Object element) {
 	public void inputChanged(final Viewer viewer, final Object oldInput, final Object newInput) {
 		ITreeContentProvider.super.inputChanged(viewer, oldInput, newInput);
 
-		treeViewer = (TreeViewer) viewer;
 		if (newInput instanceof HierarchyViewInput viewInput) {
-			hierarchyItems = null;
```
sebthom, line 95: shouldn't this be  `hierarchyItems = Collections.emptyList();` now?

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/typeHierarchy/TypeHierarchyViewContentProvider.java
```diff
@@ -42,18 +44,15 @@
...
 	public boolean showSuperTypes = true;
 
 	@Override
 	public Object[] getElements(Object inputElement) {
-		if (hierarchyItems != null) {
-			if (hierarchyItems.isEmpty()) {
-				return new Object[] { Messages.TH_no_type_hierarchy };
-			}
-			return hierarchyItems.toArray();
+		if (hierarchyItems.isEmpty()) {
```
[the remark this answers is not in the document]
ghentschke: Hm, strange. I can't find a null assignment in line 122 in commit 591ff89. In this commit there is no null assignment to `hierarchyItems` anymore.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/typeHierarchy/TypeHierarchyViewContentProvider.java
ghentschke: I removed this line, because the shown type hierarchy and its members should not be deleted if there is no Type Hierarchy for the current selection (i.e. select an integer variable and press F4 -> Type Hierarchy view does not change (Behavior of the old Type Hierarchy view))

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/typeHierarchy/TypeHierarchyViewContentProvider.java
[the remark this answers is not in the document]
sebthom: In line 128 you are testing for nullability of hierarchyItems: `if (hierarchyItems != null && !hierarchyItems.isEmpty()) {` which tells me you expect that hierarchyItems can be null, therefore my question.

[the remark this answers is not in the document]
ghentschke: Sorry, now I get it ;-) 
> Or maybe the hierarchyItems != null check is redundant and can be removed.

That's right!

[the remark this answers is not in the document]
ghentschke: I'll remove the whole if clause, because I already check the items to be present in line 120: 
`if (!p.getSecond().isEmpty()) {`

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/typeHierarchy/TypeHierarchyViewContentProvider.java
```diff
@@ -125,10 +124,8 @@ private void initialise(final @NonNull IDocument document, final int offset, Tre
 							if (treeViewer != null) {
 								treeViewer.refresh();
 								treeViewer.expandToLevel(2);
-								if (hierarchyItems != null && !hierarchyItems.isEmpty()) {
```
sebthom: you now also removed the `!hierarchyItems.isEmpty()` check. Was that on purpose? `hierarchyItems.get(0)` will fail if the list is empty

ghentschke: > Was that on purpose? 

Yes, because `hierarchyItems.get(0)` can't fail since  `hierarchyItems` will be assigned to `p.getSecond()`which will be checked  to be non empty in the if condition (see line 120 and following):
```java
if (!p.getSecond().isEmpty()) {
  hierarchyItems = p.getSecond();
  treeViewer = viewer;
  PlatformUI.getWorkbench().getDisplay().asyncExec(() -> {
	  if (treeViewer != null) {
		  treeViewer.refresh();
		  treeViewer.expandToLevel(2);
		  treeViewer.getControl().setEnabled(true);
		  treeViewer.setSelection(new StructuredSelection(hierarchyItems.get(0)));
	  }
   });
}
```
`hierarchyItems` will be assinged to `Collections.emptyList()` only during init or in `handleRootError`.

ghentschke: I added the check again

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/typeHierarchy/TypeHierarchyView.java
```diff
@@ -11,32 +11,343 @@
...
+		pagebook.showPage(splitter);
+	}
+
+	private void onHierarchySelectionChanged(SelectionChangedEvent event) {
+		var selection = event.getSelection();
+		if (selection instanceof TreeSelection && !selection.isEmpty()) {
+			var element = ((TreeSelection) selection).getFirstElement();
+			if (element instanceof TypeHierarchyItem item) {
+				typeName = item.getName();
+				IFile file = null;
```
rubenporras, line 283: could we move down the declaration of file to line 290? Just before the block where it is asigned?

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/typeHierarchy/TypeHierarchyView.java
```diff
@@ -74,21 +385,99 @@ private void updateHierarchyModeItem(ToolItem hierchyModeItem, boolean showSuper
 				hierchyModeItem.setToolTipText(showSuperTypes ? Messages.typeHierarchy_show_subtypes : Messages.typeHierarchy_show_supertypes);
 			}
 		};
-		treeViewer = filteredTree.getViewer();
-		treeViewer.setContentProvider(contentProvider);
+	}
 
-		treeViewer.setLabelProvider(new TypeHierarchyItemLabelProvider());
+	private SymbolsContainer getSymbolsContainer(IFile file) {
+		SymbolsContainer symbolsContainer = cachedSymbols.get(file.getLocationURI());
```
rubenporras, line 391: instead of using  `cachedSymbols.get` and  `cachedSymbols.put`, could we not use  `cachedSymbols.computeIfAbsent`?

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/typeHierarchy/TypeHierarchyView.java
```diff
@@ -11,32 +11,343 @@
...
+					symbolsContainer.setFile(file);
+					//create new cache element under new URI
+					cachedSymbols.put(file.getLocationURI(), new SymbolsContainer(file));
+				}
+			}
+		}
+
+		private SymbolsContainer getSymbolsContainer(IFileBuffer buffer) {
+			if (buffer != null) {
+				return cachedSymbols.get(LSPEclipseUtils.toUri(buffer));
```
rubenporras, line 175: LSPEclipseUtils.toUri(buffer) could return null, I think we need to check that before accessing the cachedSymbols map.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/typeHierarchy/TypeHierarchyView.java
```diff
@@ -74,21 +385,99 @@ private void updateHierarchyModeItem(ToolItem hierchyModeItem, boolean showSuper
...
+			if (document != null) {
+				final var params = new DocumentSymbolParams(
+						LSPEclipseUtils.toTextDocumentIdentifier(document));
+				CompletableFuture<Optional<LanguageServerWrapper>> languageServer = LanguageServers
+						.forDocument(document)
+						.withFilter(
+								capabilities -> LSPEclipseUtils.hasCapability(capabilities.getDocumentSymbolProvider()))
+						.computeFirst((w, ls) -> CompletableFuture.completedFuture(w));
+				try {
+					symbols = languageServer.get(500, TimeUnit.MILLISECONDS).filter(Objects::nonNull)
```
rubenporras, line 414: if symbols contains already a future, should we not cancel it?

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/typeHierarchy/TypeMemberContentProvider.java
```diff
@@ -0,0 +1,58 @@
...
+import org.eclipse.lsp4e.outline.SymbolsModel.DocumentSymbolWithURI;
+import org.eclipse.lsp4j.DocumentSymbol;
+
+public class TypeMemberContentProvider implements IStructuredContentProvider {
+	private static final Object[] NO_CHILDREN = new Object[0];
+	DocumentSymbolWithURI symbolContainer;
+
+	@Override
+	final public void inputChanged(Viewer viewer, Object oldInput, Object newInput) {
+		if (newInput instanceof DocumentSymbolWithURI) {
```
rubenporras, line 29:
```
if (newInput instanceof DocumentSymbolWithURI symbolWithURI) {
			symbolContainer = symbolWithURI;
}
```

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/typeHierarchy/TypeMemberContentProvider.java
```diff
@@ -0,0 +1,58 @@
...
+
+	@Override
+	public void dispose() {
+		symbolContainer = null;
+	}
+
+	@Override
+	public Object[] getElements(Object inputElement) {
+		return Optional.ofNullable(symbolContainer)
+				.map(container -> container.symbol)
```
rubenporras, line 42: if it can be done, combining several map operations in one, will perform better, since readability is still good, I would propose `.map(container -> toContainer(container.symbol.getChildren()))`

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/typeHierarchy/TypeMemberContentProvider.java
```diff
@@ -0,0 +1,58 @@
...
+				.map(symbol -> symbol.getChildren())
+				.map(this::toContainer).orElse(NO_CHILDREN);
+	}
+
+	private Object[] toContainer(List<DocumentSymbol> symbols) {
+		if (symbols != null) {
+			List<DocumentSymbolWithURI> container = new ArrayList<>();
```
rubenporras, line 49: here for performance as well, we can create an array directly and fill that, without creating a list that must be copied later to an array, without loosing readability.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/typeHierarchy/TypeHierarchyView.java
```diff
@@ -11,32 +11,343 @@
...
+				memberViewer.setInput(symbol);
+				memberLabel.setText(typeName);
+				if (symbol != null ) {
+					memberViewer.getControl().setEnabled(true);
+					memberViewer.setSelection(new StructuredSelection(symbol));
+					memberLabel.setImage(symbolsLabelProvider.getImage(symbol));
+				}
+			} else {
+				memberViewer.setInput(null); // null clears it
+				var text = Messages.TH_cannot_find_file;
```
rubenporras, line 314: should the text not be wrapped in a call to NLS so that it can be translated if desired? For translations to work properly, the file should be a placeholder inside the string that is the message.properties file.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/typeHierarchy/TypeHierarchyView.java
```diff
@@ -11,32 +11,343 @@
...
+		treeViewer.setContentProvider(contentProvider);
+
+		treeViewer.setLabelProvider(new TypeHierarchyItemLabelProvider());
+
+		treeViewer.setUseHashlookup(true);
+		treeViewer.getControl().setEnabled(false);
+		treeViewer.addDoubleClickListener(new IDoubleClickListener() {
+			@Override
+			public void doubleClick(final DoubleClickEvent event) {
+				var selection = ((IStructuredSelection) event.getSelection()).getFirstElement();
```
rubenporras, line 230: could we have a ClassCastException here?

rubenporras (changes requested):

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/typeHierarchy/TypeHierarchyView.java
ghentschke: `cachedSymbols.get` returns `null` if the input argument is `null`

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/typeHierarchy/TypeHierarchyView.java
ghentschke: The calling method `refreshSymbols` returns when the last completions stage has joined (see line  436). `refreshSymbols` will be called by the `synchronized` method `refreshMemberViewer`. That's why I assumed that the future cannot be called again until it has been completed.

ghentschke: Correct me if I am wrong

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/typeHierarchy/TypeHierarchyView.java
rubenporras: Great, my bad then.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/typeHierarchy/TypeHierarchyView.java
ghentschke: I don't think so because `event.getSelection` returns a `TreeSelection` object and `TreeSelection` extends `StructuredSelection` which implements `IStructuredSelection`.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/typeHierarchy/TypeHierarchyView.java
ghentschke: Messages already extends NLS. So I assumed that the String `TH_cannot_find_file` will be translated correctly.

rubenporras: Still, one need to account for different grammar structures in different languages, so one should not make assumptions as to what those structures will be. The current code makes the assumption that the file can be just appended at the end end of the sentence, which might not be true in some languages. That is why the message should be
TH_cannot_find_file = Cannot find file {0}
so that a translator can decide where {0} will go.

ghentschke: Okay, I got it. Thank you!

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/typeHierarchy/TypeHierarchyView.java
```diff
@@ -11,32 +11,343 @@
...
+		private SymbolsContainer getSymbolsContainer(IFileBuffer buffer) {
+			if (buffer != null) {
+				return cachedSymbols.get(LSPEclipseUtils.toUri(buffer));
+			}
+			return null;
+		}
+
+	};
+
+	private IDocumentListener documentListener = new IDocumentListener() {
```
rubenporras, line 182: if we make `fileBufferListener` final, we should do the same here.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/typeHierarchy/TypeHierarchyView.java
```diff
@@ -74,21 +384,95 @@ private void updateHierarchyModeItem(ToolItem hierchyModeItem, boolean showSuper
...
+		final IDocument document = symbolsContainer.getDocument();
+		try {
+			if (document != null) {
+				CompletableFuture<List<Either<SymbolInformation, DocumentSymbol>>> symbols;
+				final var params = new DocumentSymbolParams(
+						LSPEclipseUtils.toTextDocumentIdentifier(document));
+				CompletableFuture<Optional<LanguageServerWrapper>> languageServer = LanguageServers
+						.forDocument(document)
+						.withFilter(
+								capabilities -> LSPEclipseUtils.hasCapability(capabilities.getDocumentSymbolProvider()))
```
rubenporras, line 406: this can just be `.withCapability(ServerCapabilities::getDocumentSymbolProvider)`

rubenporras (commented):
The change looks good to me now but I do not know very good the JFace and SWT APIs, should some else do an additional review?

mickaelistria:
I don't think I'll give a detailed review here. I trust you here @ghentschke and @rubenporras that if you both agree it's a good patch to merge, let's just merge it. In worst case, if there are some issues, we'll fix or revert as we find them.

mickaelistria:
I intended to use "Squash and merge" but "rebased and merged" instead. Sorry for the noise in the history!

Issue #743, opened 2023-08-05
Completion "hangs" with no response 


vrubezhny:
In some cases Completion hardly "hangs" without a response from server logged, like:
```
[2023-07-20T15:04:27.107621992+02:00] LSP4E_TO_LANGUAGE_SERVER org.eclipse.wildwebdeveloper.xml:
{"jsonrpc":"2.0","id":"31","method":"textDocument/completion","params":{"textDocument":{"uri":"file:///home/jeremy/projects/eclipse/source/tycho/demo/osgi-repository/repository/pom.xml"},"position":{"line":36,"character":5}}}
```
For reference issue see: https://github.com/eclipse/lemminx-maven/issues/444

I believe LSP4E has problems in sending a request to the server, - a problem which I'd like to be investigated - so the response cannot be generated, but even if a LS is causing the problem, completion request is never finishes on LSP4E due to "infinite waiting on the completion future by calling `.get()`:
https://github.com/eclipse/lsp4e/blob/55ce45f20dba1147cfa8383412f0105bfe96fc2e/org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/completion/LSContentAssistProcessor.java#L114C11-L114C11
```java
this.completionLanguageServersFuture = LanguageServers.forDocument(document)
	.withFilter(capabilities -> capabilities.getCompletionProvider() != null)
	.collectAll((w, ls) -> ls.getTextDocumentService().completion(param)
		.thenAccept(completion -> proposals.addAll(toProposals(document, offset, completion, w))));
this.completionLanguageServersFuture.get();
```
While it's very difficult to catch the reason of sending request/receiving response failures, the using `CompletableFuture.get()` is to be strongly avoided.

angelozerr:
Using `this.completionLanguageServersFuture.get();` means that you trust the language server, but if language server return a result whish never finish, the this.completionLanguageServersFuture.get() is never free.

If user open several completions, a Thread is taken again from the pool again and after several completion, there is no Thread in the pool and language server becomes unusable not only with completion but with other features like highlighting, validation etc

I suspect that this issue with strange error comes from that there is not Thread in the pool https://github.com/eclipse-wildwebdeveloper/wildwebdeveloper/issues/1251

[...]

PR #744, opened 2023-08-05
Completion "hangs" with no response #743

vrubezhny:
Getting rid of infinite waiting in LSContentAssistProcessor

Issue: #743

### DIFF
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/completion/LSContentAssistProcessor.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/completion/LSContentAssistProcessor.java
@@ -64,6 +64,7 @@ import com.google.common.base.Strings;
 public class LSContentAssistProcessor implements IContentAssistProcessor {
 
 	private static final long TRIGGERS_TIMEOUT = 50;
+	private static final long GET_TIMEOUT = 1000;
 	private static final long CONTEXT_INFORMATION_TIMEOUT = 1000;
 	private IDocument currentDocument;
 	private String errorMessage;
@@ -111,7 +112,7 @@ public class LSContentAssistProcessor implements IContentAssistProcessor {
 					.withFilter(capabilities -> capabilities.getCompletionProvider() != null)
 					.collectAll((w, ls) -> ls.getTextDocumentService().completion(param)
 							.thenAccept(completion -> proposals.addAll(toProposals(document, offset, completion, w))));
-			this.completionLanguageServersFuture.get();
+			this.completionLanguageServersFuture.get(GET_TIMEOUT, TimeUnit.MILLISECONDS);
 		} catch (ResponseErrorException | ExecutionException e) {
 			if (!CancellationUtil.isRequestCancelledException(e)) { // do not report error if the server has cancelled the request
 				LanguageServerPlugin.logError(e);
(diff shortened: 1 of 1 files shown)


vrubezhny:
IMHO, the failing JUnit test cases are not related to the fix - the PR is built 3 times - every time the test failures were different.

vrubezhny:
Unfortunately, the stack trace doesn't allow to determine which exact test generated the TimeoutException:
```
org.eclipse.lsp4e.test.edit.LSPEclipseUtilsTest.testResourceOperations -- Time elapsed: 0.116 s <<< FAILURE!
java.lang.AssertionError: Some errors were logged expected:<[]> but was:<[Status ERROR: org.eclipse.lsp4e code=0 TimeoutException java.util.concurrent.TimeoutException]>
	at org.junit.Assert.fail(Assert.java:89)
	at org.junit.Assert.failNotEquals(Assert.java:835)
	at org.junit.Assert.assertEquals(Assert.java:120)
	at org.eclipse.lsp4e.test.NoErrorLoggedRule.finished(NoErrorLoggedRule.java:61)

```
and
```
SEVERE: Status ERROR: org.eclipse.lsp4e code=0 TimeoutException java.util.concurrent.TimeoutException
java.util.concurrent.TimeoutException
	at java.base/java.util.concurrent.CompletableFuture.timedGet(CompletableFuture.java:1960)
	at java.base/java.util.concurrent.CompletableFuture.get(CompletableFuture.java:2095)
	at org.eclipse.lsp4e.LanguageServerWrapper.lambda$10(LanguageServerWrapper.java:458)
	at java.base/java.lang.Thread.run(Thread.java:833)
```
But as we don't see any tests failed - it looks like actually the `LSPEclipseUtilsTest.testResourceOperations` test cases is succeeded - and timeout exception actually didn't hurt the execution, but `NoErrorLoggedRule` rule has marked it as "failed".
I don't know if it was actually a real "resource operation problem" or it's just that it took too much time for some reason - it still worth investigating, but timeouts may happen and logged and in most real cases are not because of any failure - we just need to be prepared and process them accordingly.

sebthom:
@vrubezhny I created PR #749 which should improve the build stability.

angelozerr:
Please see an another proposal to fix completion hang https://github.com/eclipse/lsp4e/pull/750

rubenporras:
@vrubezhny, I think the approach proposed by @angelozerr is preferable, as it does not need a timeout that might discard useful answers just because they are a bit too late, and I also expect it would be more responsive as it is able to cancel futures that are no longer needed. What do you think?

mickaelistria:
> I already addressed some of them but I feel it is getting worse. maybe something in the jenkins infrastructure changed

I have not followed recent developments as I was on holiday, but often when this has happened, it can be caused by an actual performance regression in LSP4E itself more than an infra issue.

mickaelistria:
Damn, I hit "merge" by mistake.
For completion processors, operations can be very long yet very valuable. For some cases where completion would be a very deep analyzer, it could be OK to wait 15 seconds for the output of a completion request (imagine this one generates a whole software...).
What we want here is indeed @rubenporras mentioned and @angelozerr suggests: support for proper cancellation.
So I'm reverting it.

rubenporras:
I am not sure if in addition to @angelozerr suggestion, we still need this PR or some improved version, but I am happy that we are cautious here. The cancellation support could be merged and it should improve the situation, what do you think, @mickaelistria ?

PR #750, opened 2023-08-08
Completion "hangs" with no response

angelozerr:
After emulating a long process (which can be cancellable), and debugging LSP4E the thing which is the most important is to cancel  the following future

https://github.com/eclipse/lsp4e/blob/55ce45f20dba1147cfa8383412f0105bfe96fc2e/org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/completion/LSContentAssistProcessor.java#L114

when a new completion is triggered. I think we can live with just one future which do a .get(). If this future is cancel it is good. The master problem is that as future was never cancelled,, after some several completion,the Thread pooling was not free.

My PR fixes this problem and more, because it supports cancelllation support for completion. For maven completion for instance it is very interesting to have this feature to cancel ASAP the compute of the completion by using LSP4J cancelchecker https://github.com/eclipse/lemminx-maven/blob/0f3337bcc40244905d1fad9a7b2b522e26e17d26/lemminx-maven/src/main/java/org/eclipse/lemminx/extensions/maven/participants/completion/MavenCompletionParticipant.java#L203


angelozerr:
IMHO I think this CancellationSupport should be hosted by the LanguageServers API.

Today we have ` LanguageServers.forDocument(document)....colectAll(w, ls)`

I think we should have `LanguageServers.forDocument(document)...collectAll(context, cancelChecker)`

 * context could have getter for the LanguageServer, the LanguageServerWrapper and the execute method that I add in the CancellationSupport (it does nothing than add a future in the futures to cancel)
 * cancelChecker check that the main future is not cancelled. When the main future will be cancelled, the call of checkCancelled will throw a CancelllationException

rubenporras:
> IMHO I think this CancellationSupport should be hosted by the LanguageServers API.
> 
> Today we have ` LanguageServers.forDocument(document)....colectAll(w, ls)`
> 
> I think we should have `LanguageServers.forDocument(document)...collectAll(context, cancelChecker)`
> 
> * context could have getter for the LanguageServer, the LanguageServerWrapper and the execute method that I add in the CancellationSupport (it does nothing than add a future in the futures to cancel)
> * cancelChecker check that the main future is not cancelled. When the main future will be cancelled, the call of checkCancelled will throw a CancelllationException

@ahmedneilhussain , what do you think of the proposal of adding the API to collectAll?

ahmedneilhussain:
> 
> won't work because the implementation of `collectAll` is unable to get at the raw completable futures.

Thinking about it a bit more, I suppose we could adopt an aspect-oriented approach and make the executor pass in a proxied wrapper of the language server that could decorate the result of `hover()`. It would be a bit fiddly, but might be possible. However I'm not sure it would play well with custom extensions of the LanguageServer interface.

angelozerr:
After implementing cancellation support in lsp4ij for all operations I think my original idea the PR I did is the better solution which is simple and gives master of which future it should be cancelled.

Sorry to give you some confusions with suggestions with move canncellation support la the API

rubenporras:
I would like to merge this PR before the release, as it does not introduce any public class, we can refactor later if needed, and it clearly improves the current behaviour. What do you think @mickaelistria ?

mickaelistria:
Please wait a bit, I'd review it before the end of this day.

rubenporras:
sure, I will wait with the release until you review it.

mickaelistria:
I think this PR is good. I wish we can make cancellation feel a bit more native later (by having `cancel` on the compound feature) but it can be a next iteration and this PR can serve as a good basis.
@angelozerr do you think it would be doable to write a test-case for it? Cancellation is a non-trivial topic and I'm afraid it will be easy to introduce regressions if we don't have some checks to cover it.

rubenporras:
Since this PR might not be ready, I will proceed with the release without it. Thanks

angelozerr:
> think this PR is good. I wish we can make cancellation feel a bit more native later

I m using this cancellation support in IJ for all opérations and it seems it is working prêtre well.

I tried to do à smarter cancellation support but the code was very complex so I give up.

IHMO I think this PR is very important since completion is opened anywhere and dont free thread. 

I m sorry I will not have time to write tests. I wanted just to share with lsp4e how I implemented cancel support in IJ

mickaelistria:
Yes, sorry, I forgot about it. Thanks all for this good addition!

PR #753, opened 2023-08-10
Cannot open any css file with the generic editor #752

vrubezhny:
Fixes: #752


org.eclipse.lsp4e/src/org/eclipse/lsp4e/outline/LSSymbolsContentProvider.java
```diff
@@ -86,14 +85,14 @@ public static final class OutlineViewerInput {
...
-			TextDocumentIdentifier docIdentifier = LSPEclipseUtils.toTextDocumentIdentifier(document);
 			if (path == null) {
 				documentFile = null;
 				URI uri;
 				// Set the documentURI if valid URI for LSs that use custom protocols e.g. jdt://
 				try {
+					TextDocumentIdentifier docIdentifier = LSPEclipseUtils.toTextDocumentIdentifier(document);
 					uri = new URI(docIdentifier.getUri());
-				} catch (URISyntaxException e) {
+				} catch (Exception e) {
```
akurtakov, line 95: Catching Exception is really bad style. As this PR is about fixing NPE couldn't you keep the original exception and do an extra null check.

vrubezhny: Added a null-check for document URI and  restored catching original `URISyntaxException`

sebthom (approved):

PR #755, opened 2023-08-10
add LanguageServerWrapper#getProcessHandle()

sebthom:
This PR adds the `LanguageServerWrapper#getProcessHandle()` method and enriches the the `LanguageServerWrapper#toString()` info with the pid of the underlying LS process.

The `LanguageServerWrapper#getProcessHandle()` will also be useful when implementing a LanguageServers UI view to access information such as pid, working directory or command of the underlying LS process.


vrubezhny (commented):
@sebthom How about making  `LoggingStreamConnectionProviderProxy` to also implement `IAdaptab;e`? Otherwise its clients will not be able to obtain process handle, will they?

angelozerr:
Same thing for LaunchConfigurationStreamProvider I think?

vrubezhny:
@sebthom Looks good to me.  Having access to  process handle is a good thing, so  `+1`

PR #762, opened 2023-08-16
export format package

### LINKED ISSUE #761: Support LSP based source code formatting on file save
As a user of a LSP based editor I want to format my source code on a file save action. (see this [cdt-lsp issue](https://github.com/eclipse-cdt/cdt-lsp/issues/178))

Therefor I want to use the `org.eclipse.lsp4e.operations.format.LSPFormatter` class. The package `org.eclipse.lsp4e.operations.format` should be exported.

ghentschke:
This package is needed in DocumentProviders of LSP based editors to support the format on file save action. Only needed for servers who do not support willSaveWaitUntil

fixes #761

### DIFF
--- a/org.eclipse.lsp4e/META-INF/MANIFEST.MF
+++ b/org.eclipse.lsp4e/META-INF/MANIFEST.MF
@@ -50,6 +50,7 @@ Export-Package: org.eclipse.lsp4e;x-internal:=true,
  org.eclipse.lsp4e.operations.codeactions;x-internal:=true,
  org.eclipse.lsp4e.command;x-internal:=true,
  org.eclipse.lsp4e.operations.completion;x-internal:=true,
+ org.eclipse.lsp4e.operations.format;x-internal:=true,
  org.eclipse.lsp4e.operations.hover;x-internal:=true,
  org.eclipse.lsp4e.outline;x-internal:=true,
  org.eclipse.lsp4e.server;version="0.1.0"


ghentschke:
I think the Jenkins build job hangs...

rubenporras:
How would you like to use the API? Are you going to have react to a "saved" event? 

Will the format operation not leave the editor dirty? Would the operation save the buffer as well after applying format to avoid that?

How do you plan to deal with concurrent edits that could happen after the user saves but before the format request is processed?

I the reason I am asking is to make your aware of these problems. My reason for using `willSaveWaitUntil` was to avoid these problems.

ghentschke:
> How would you like to use the API? Are you going to have react to a "saved" event?

Yes. I'am using the  `org.eclipse.ui.editors.documentProviders` extension point for a `LspDoucmentProvider` who extends the  `TextFileDocumentProvider` and overwrite the `createSaveOperation` method.

> Will the format operation not leave the editor dirty?

No. 

>  Would the operation save the buffer as well after applying format to avoid that?

Yes., the buffer will be saved after applying the format changes (if this is enabled)

> How do you plan to deal with concurrent edits that could happen after the user saves but before the format request is processed?

Thank you. This is an open  issue. I assume the time window for the changes is very small, since there is a 1s timeout for the server to process. My implementation is similar to the existing format-on-save in the old C/C++ Editor in cdt. Except that the `TextEdit`'s comes from the LSP server now.

Please check my [WIP PR #178](https://github.com/eclipse-cdt/cdt-lsp/pull/179) on cdt-lsp for format-on-save

PR #764, opened 2023-08-20
support PendingUpdateAdapter in SymbolsLabelProvider

### LINKED ISSUE #763: Support PendingUpdateAdapter in SymbolsLabelProvider
In [cdt-lsp](https://github.com/eclipse-cdt/cdt-lsp) project we use the `org.eclipse.lsp4e.outline.SymbolsLabelProvider` to provide symbols label in the Project Explorer view to our Content Provider. 
When the language server has to build a large AST for the file, the calculation of the tree can take some time.
It would be great when the `SymbolsLabelProvider` would support the `PendingUpdateAdapter` object in `org.eclipse.lsp4e.outline.SymbolsLabelProvider.getImage(Object)` and `org.eclipse.lsp4e.outline.SymbolsLabelProvider.getStyledText(Object)` to display 'Computing symbols...':
image

i.e.:
```java
if (element == LSSymbolsContentProvider.COMPUTING || element instanceof PendingUpdateAdapter) {
	return new StyledString(Messages.outline_computingSymbols);
}
```

See also [cdt-lsp #186](https://github.com/eclipse-cdt/cdt-lsp/pull/186)

WDYT?

ghentschke:
fixes #763

### DIFF
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/outline/SymbolsLabelProvider.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/outline/SymbolsLabelProvider.java
@@ -120,7 +120,7 @@ public class SymbolsLabelProvider extends LabelProvider
 		if (element == null){
 			return null;
 		}
-		if (element == LSSymbolsContentProvider.COMPUTING || element instanceof PendingUpdateAdapter) {
+		if (element instanceof PendingUpdateAdapter) {
 			return JFaceResources.getImage(ProgressManager.WAITING_JOB_KEY);
 		}
 		if (element instanceof Throwable) {
@@ -252,7 +252,7 @@ public class SymbolsLabelProvider extends LabelProvider
 	@Override
 	public StyledString getStyledText(Object element) {
 
-		if (element == LSSymbolsContentProvider.COMPUTING || element instanceof PendingUpdateAdapter) {
+		if (element instanceof PendingUpdateAdapter) {
 			return new StyledString(Messages.outline_computingSymbols);
 		}
 		if (element instanceof Throwable throwable) {
@@ -329,14 +329,14 @@ public class SymbolsLabelProvider extends LabelProvider
 		}
 		return res;
 	}
-	
+
 	private boolean isDeprecated(List<SymbolTag> tags) {
 		if(tags != null){
 			return tags.contains(SymbolTag.Deprecated);
 		}
 		return false;
 	}
-	
+
 	@Override
 	public void restoreState(IMemento aMemento) {
 	}
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/outline/SymbolsLabelProvider.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/outline/SymbolsLabelProvider.java
@@ -60,6 +60,7 @@ import org.eclipse.ui.ISharedImages;
 import org.eclipse.ui.internal.progress.ProgressManager;
 import org.eclipse.ui.navigator.ICommonContentExtensionSite;
 import org.eclipse.ui.navigator.ICommonLabelProvider;
+import org.eclipse.ui.progress.PendingUpdateAdapter;
 
 import com.google.common.collect.RangeMap;
 import com.google.common.collect.TreeRangeMap;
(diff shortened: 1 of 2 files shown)


org.eclipse.lsp4e/src/org/eclipse/lsp4e/outline/SymbolsLabelProvider.java
```diff
@@ -119,7 +120,7 @@ public Image getImage(Object element) {
 		if (element == null){
 			return null;
 		}
-		if (element == LSSymbolsContentProvider.COMPUTING) {
+		if (element == LSSymbolsContentProvider.COMPUTING || element instanceof PendingUpdateAdapter) {
```
rubenporras, line 123: Hi @ghentschke , 

I have read the class but I still do not understand the change. Which code is creating objects of this type in the context of LSP4E?

mickaelistria:
Shouldn't we just replace the `COMPUTING` object (remove it) in favor or PendingUpdateAdapter ?

rubenporras:
> Shouldn't we just replace the `COMPUTING` object (remove it) in favor or PendingUpdateAdapter ?

This has also been my first thought when I look at the PR. That org.eclipse.lsp4e.outline.LSSymbolsContentProvider.COMPUTING would be an instance of `PendingUpdateAdapter`, but what would be the relation with **ctd-lsp**?

ghentschke:
> Shouldn't we just replace the COMPUTING object (remove it) in favor or PendingUpdateAdapter ?

 > This has also been my first thought when I look at the PR. That org.eclipse.lsp4e.outline.LSSymbolsContentProvider.COMPUTING would be an instance of PendingUpdateAdapter, but what would be the relation with ctd-lsp?

 I assume that the replacment of the COMPUTING  object would be a better appproach. In cdt-lsp we use the  `PendingUpdateAdapter` object while the asynchronous fetch of the translation unit symbols via LS in ongoing. This is the default object returned by `org.eclipse.ui.progress.DeferredTreeContentManager` in `getChildren` as placeholder:
 ```java
	/**
	 * Returns the child elements of the given element, or in the case of a deferred
	 * element, returns a placeholder. If a deferred element is used, a job is
	 * created to fetch the children in the background.
	 *
	 * @param parent The parent object.
	 * @return Object[] or <code>null</code> if parent is not an instance of
	 *         IDeferredWorkbenchAdapter.
	 */
	public Object[] getChildren(final Object parent) {
		IDeferredWorkbenchAdapter element = getAdapter(parent);
		if (element == null) {
			return null;
		}
		PendingUpdateAdapter placeholder = createPendingUpdateAdapter();
		startFetchingDeferredChildren(parent, element, placeholder);
		return new Object[] { placeholder };
	}
```

File: org.eclipse.lsp4e/src/org/eclipse/lsp4e/outline/SymbolsLabelProvider.java (hunk 1 above)
Comment (ghentschke) [reply org.eclipse.lsp4e/src/org/eclipse/lsp4e/outline/SymbolsLabelProvider.java]: It comes from `org.eclipse.ui.progress.DeferredTreeContentManager.getChildren(Object)` which is used to calculate the symbols for the Project Explorer tree asynchronously:
```java
	public Object[] getChildren(final Object parent) {
		IDeferredWorkbenchAdapter element = getAdapter(parent);
		if (element == null) {
			return null;
		}
		PendingUpdateAdapter placeholder = createPendingUpdateAdapter();
		startFetchingDeferredChildren(parent, element, placeholder);
		return new Object[] { placeholder };
	}
```

mickaelistria (approved):

org.eclipse.lsp4e/src/org/eclipse/lsp4e/outline/SymbolsLabelProvider.java
```diff
@@ -252,7 +252,7 @@ public String getText(Object element) {
 	@Override
 	public StyledString getStyledText(Object element) {
 
-		if (element == LSSymbolsContentProvider.COMPUTING || element instanceof PendingUpdateAdapter) {
+		if (element instanceof PendingUpdateAdapter) {
```
rubenporras, line 255: would it be good for UI consistency to replace LSP4E custom message with 
```
if (element instanceof PendingUpdateAdapter pendingUpdate) {
  pendingUpdate.getLaber(null);
}
```

rubenporras (approved):

org.eclipse.lsp4e/src/org/eclipse/lsp4e/outline/SymbolsLabelProvider.java
mickaelistria: I think it might even better for consistency to use directly a `WorkbenchLabelProvider` as super and return `super.getStyledText()` in case the element is not a LS symbol.
But the current patch is clearly a good improvement already. Let's merge.

Issue #741, opened 2023-08-02
ItemDefaults client capability is not implemented


angelozerr:
We have supported ItemDefaults with our XMLLanguage Server to speed the completion for lemmix maven which can return big list (artifact completion) but we cannot benefit with  feature since LSP4E client doesntsend thepropre completion ItemDefaults capability.

In vscode it send:

```
 "completionList": {
                    "itemDefaults": [
                        "commitCharacters",
                        "editRange",
                        "insertTextFormat",
                        "insertTextMode"
                    ]
                }
```

mickaelistria:
There is support for itemDefaults, see https://github.com/eclipse/lsp4e/blob/92abe97f21f72dca1f07bb6e16fa452b1763c272/org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/completion/LSContentAssistProcessor.java#L230 and #599  . Can you please elaborate what is missing?

angelozerr:
The language server needs to know if the LSP client support Item defaults to generate completion with default. In other words it misses the LSP4E completion client capability, 

See LSP4IJ https://github.com/redhat-developer/intellij-quarkus/blob/a0e94debf4d88d0866ab081491f8fe81ae547304/src/main/java/com/redhat/devtools/intellij/lsp4ij/internal/SupportedFeatures.java#L58C26-L58C43

mickaelistria:
ah ok, thanks. I'm patchingit.

angelozerr:
And LSP4E  don't support item default with editRange which IMHO is the most used to avoid generating a big JSON with the same edit range. See LSP4IJ https://github.com/redhat-developer/intellij-quarkus/blob/a0e94debf4d88d0866ab081491f8fe81ae547304/src/main/java/com/redhat/devtools/intellij/lsp4ij/operations/completion/LSPCompletionContributor.java#L148

angelozerr:
To check that it is working, I suggest that you open completion with XML, it should return completion list with default editRange and not an editRange per completion item.

PR #782, opened 2023-08-25
Declare itemDefaults capabilities

mickaelistria:
Fixes https://github.com/eclipse/lsp4e/issues/741


org.eclipse.lsp4e/src/org/eclipse/lsp4e/internal/SupportedFeatures.java
```diff
@@ -84,6 +85,10 @@ public class SupportedFeatures {
 		completionItemCapabilities.setResolveSupport(new CompletionItemResolveSupportCapabilities(List.of("documentation", "detail", "additionalTextEdits"))); //$NON-NLS-1$ //$NON-NLS-2$ //$NON-NLS-3$
 		final var completionCapabilities = new CompletionCapabilities(completionItemCapabilities);
 		completionCapabilities.setContextSupport(Boolean.TRUE);
+		completionCapabilities.setCompletionList(new CompletionListCapabilities(List.of("commitCharacters", //$NON-NLS-1$
+                        "editRange", //$NON-NLS-1$
```
angelozerr, line 89: You don't support it I think.

angelozerr:
It looks good, but I suggest really to test that XML language server completion returns default value for edit range and  check completion continue to work.

PR #786, opened 2023-08-31
Json parameters can now be substituted.

### LINKED ISSUE #756: Enable string interpolation for Json launch parameters
# Steps to reproduce # 
1. Open the Debug Configuration Wizard
2. Create a new configuration for the "Debug Adapter Launcher"
3. Enter valid json in the "Launch parameters" textbox. For example: 
```
{
    "webRoot":"${workspace_loc:/some_project}",
}
```
The plugin does not perform string interpolation for the json launch parameters. In contrast to VS Code, which does (see here: https://github.com/microsoft/vscode-chrome-debug#launch and here: https://github.com/firefox-devtools/vscode-firefox-debug#launch) 

Without string interpolation, it makes no sense to commit the launch configuration to git, because the configurations depend on the directory structure of the commiter. By enabling string interpolation, committing the launch configurations into git makes sense.

Si-So:
I fixed the issue and also created a unit test for it. 

Fixes #756

## Design considerations ##
- A json object must be at the top level
    - This is the same behavior as before 
    - The launch.json from VC Code also requires a json object at the top level. See: https://github.com/microsoft/vscode/blob/main/src/vs/workbench/contrib/debug/common/debugSchemas.ts
- There is no check whether variables have been defined more than once
    - This is the same behavior as before 
    - Variables can be defined multiple times in the json
    - Variables can be defined multiple times by defining it once inside the json and defining it again outside of the json. See:  https://github.com/eclipse-wildwebdeveloper/wildwebdeveloper/blob/5fe786143580c705602f73f166910284478ac448/org.eclipse.wildwebdeveloper/src/org/eclipse/wildwebdeveloper/debug/AbstractHTMLDebugDelegate.java#L47C20-L47C20)

### DIFF
--- a/org.eclipse.lsp4e.debug/src/org/eclipse/lsp4e/debug/launcher/JsonParserWithStringSubstitution.java
+++ b/org.eclipse.lsp4e.debug/src/org/eclipse/lsp4e/debug/launcher/JsonParserWithStringSubstitution.java
@@ -1,0 +1,87 @@
+package org.eclipse.lsp4e.debug.launcher;
+
+import java.util.ArrayList;
+import java.util.LinkedHashMap;
+import java.util.Map;
+
+import org.eclipse.core.runtime.CoreException;
+import org.eclipse.core.variables.IStringVariableManager;
+
+import com.google.gson.JsonArray;
+import com.google.gson.JsonElement;
+import com.google.gson.JsonObject;
+import com.google.gson.JsonParser;
+import com.google.gson.JsonPrimitive;
+
+public class JsonParserWithStringSubstitution {
+
+	IStringVariableManager stringVariableManager;
+
+	/**
+	 * @param stringVariableManager that should be used to substitute variables in
+	 *                              strings.
+	 */
+	public JsonParserWithStringSubstitution(IStringVariableManager stringVariableManager) {
+		this.stringVariableManager = stringVariableManager;
+	}
+
+	/**
+	 * Substitutes variables in all string values within the given json.
+	 *
+	 * @param json as {@link String}
+	 * @return Returns json object as a {@link Map}. Keys are of type
+	 *         {@link String}, values are of type {@link Object}.
+	 * @throws IllegalStateException is thrown if top level element is not a
+	 *                               {@link JsonObject}.
+	 * @throws CoreException         is thrown if undefined variable was referenced
+	 *                               in json.
+	 */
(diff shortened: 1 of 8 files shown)


Si-So:
> * There is no check whether variables have been defined more than once

I just realized that a later variable definition will overwrite a previous definition. Thus, the debugger will never see a variable only once. Since the json is parsed last, the definitions from the json will overwrite everything else.

Si-So:
It looks to me that the build failure was not caused by my commit.

mickaelistria:
You need to also bump the version of the bundle you modify as it's the 1st change happening since last release.

org.eclipse.lsp4e.debug/src/org/eclipse/lsp4e/debug/launcher/JsonParserWithStringSubstitution.java
```diff
@@ -0,0 +1,87 @@
+package org.eclipse.lsp4e.debug.launcher;
```
mickaelistria, line 1: Please add a copyright header (same as other filesbut with  updated year and your copyright owner (most likely your employer)

org.eclipse.lsp4e.debug/src/org/eclipse/lsp4e/debug/launcher/JsonParserWithStringSubstitution.java
```diff
@@ -0,0 +1,87 @@
...
+import org.eclipse.core.runtime.CoreException;
+import org.eclipse.core.variables.IStringVariableManager;
+
+import com.google.gson.JsonArray;
+import com.google.gson.JsonElement;
+import com.google.gson.JsonObject;
+import com.google.gson.JsonParser;
+import com.google.gson.JsonPrimitive;
+
+public class JsonParserWithStringSubstitution {
```
mickaelistria, line 16: This makes it public API, which is probably undesired. Can you please try to move it to an internal package?

mickaelistria (changes requested):

org.eclipse.lsp4e.debug/src/org/eclipse/lsp4e/debug/launcher/JsonParserWithStringSubstitution.java
Si-So: Done

mickaelistria:
Thanks!

Si-So:
Thank you for merging :)

PR #787, opened 2023-08-31
Add .settings to gitignore. The .settings folder is created by eclipse and should be ignored.

Si-So:
See title

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ bin/
 .META-INF_MANIFEST.MF
 dash-licenses/
 *.log
+**/.settings


mickaelistria:
We actually like to keep the .settings as it contains settings that we want to share with all contributors (some things like null-analysis, some severity settings...)

Si-So:
My eclipse created in all folders the following two files: 
 .settings/org.eclipse.core.resources.prefs
 .settings/org.eclipse.m2e.core.prefs
 
Content of the `org.eclipse.core.resources.prefs` file: 
```
eclipse.preferences.version=1
encoding/<project>=UTF-8
```

Content of the `org.eclipse.m2e.core.prefs` file: 
```
activeProfiles=
eclipse.preferences.version=1
resolveWorkspaceProjects=true
version=1
```

Would it be okay for me to commit these two files?

PR #788, opened 2023-09-01
Added eclipse settings to repo

Si-So:


mickaelistria:
I think we need a preliminary commit tweaking the pom.xml so that `.settings` folder are excluded from Git timestamp/qualifier computation. Something like `jgit.ignore` as described in https://wiki.eclipse.org/Tycho/Reproducible_Version_Qualifiers

Si-So:
That is a good idea. Unfortunately, I got a few new tasks from my manager today and I don't have the time to do it. :(

Therefore, I am closing this PR.

PR #792, opened 2023-09-05
Make default implementation of configuration conform to specification

brychcy:
LSP Specification says "The order of the returned configuration settings correspond to the order of the passed ConfigurationItems (e.g. the first item in the response is the result for the first configuration item in the params). [...] If the client can’t provide a configuration setting for a given scope then null needs to be present in the returned array."

### DIFF
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageClientImpl.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageClientImpl.java
@@ -13,7 +13,7 @@
  *******************************************************************************/
 package org.eclipse.lsp4e;
 
-import java.util.Collections;
+import java.util.ArrayList;
 import java.util.List;
 import java.util.concurrent.CompletableFuture;
 import java.util.function.Consumer;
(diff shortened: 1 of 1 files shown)


org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageClientImpl.java
```diff
@@ -74,7 +74,11 @@ protected final LanguageServer getLanguageServer() {
 	@Override
 	public CompletableFuture<List<Object>> configuration(ConfigurationParams configurationParams) {
 		// override as needed
-		return CompletableFuture.completedFuture(Collections.emptyList());
+		List<Object> list = new ArrayList<>();
```
rubenporras, line 77: we should correctly size the list when creating its object using `configurationParams.getItems().size()`, could you do that?

rubenporras (changes requested):
Thanks for the PR.

In addition to the improvement when creating the ArrayList, could you also mention in the PR the relevant part of the spec?

Is it part of https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#workspace_configuration? If so, could you extract the relevant part of it or explain it in your own words?

brychcy:
> Is it part of https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification
/#workspace_configuration? 

Yes

> If so, could you extract the relevant part of it or explain it in your own words?

Hmm, i had tried to do that, see the quoted sentences in both the first comment of this PR and the commit message?

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageClientImpl.java
brychcy: done (but shouldn't really matter, these lists should always be tiny)

rubenporras: Thanks. I realize that, it is mostly a matter of getting into a good habit :)

rubenporras (approved):

PR #798, opened 2023-09-12
Fix undefined triggerCharacter for TriggerKind.Invoked

### LINKED ISSUE #794: TriggerKind is always TriggerCharacter if char is on completionTriggerChars list
During work eclipse-wildwebdeveloper/wildwebdeveloper#1316 I realized that LSP4E ignore way how code assist is invoked, and always mark completion computation as TriggerCharacter if character under cursor is inside completionTriggerChars

Also triggerCharacter is always pushed to server, which is also wrong. For TriggerKind.Invoked should be undefined

Bug has been introduced via https://github.com/eclipse/lsp4e/commit/7c770c390b3148a57818aa19c5b2060825369531

ghentschke:
triggerCharacter won't be send via json rpc in case of TriggerKind.Invoked: "context":{"triggerKind":1}

fixes #794

### DIFF
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LSPEclipseUtils.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LSPEclipseUtils.java
@@ -234,7 +234,7 @@ public final class LSPEclipseUtils {
 				} else {
 					// According to LSP 3.17 specification: the triggerCharacter in CompletionContext is undefined if
 					// triggerKind != CompletionTriggerKind.TriggerCharacter
-					param.setContext(new CompletionContext(CompletionTriggerKind.Invoked, null));
+					param.setContext(new CompletionContext(CompletionTriggerKind.Invoked));
 				}
 			} catch (BadLocationException e) {
 				LanguageServerPlugin.logError(e);
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LSPEclipseUtils.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LSPEclipseUtils.java
@@ -234,7 +234,7 @@ public final class LSPEclipseUtils {
 				} else {
 					// According to LSP 3.17 specification: the triggerCharacter in CompletionContext is undefined if
 					// triggerKind != CompletionTriggerKind.TriggerCharacter
-					param.setContext(new CompletionContext(CompletionTriggerKind.Invoked, positionCharacter));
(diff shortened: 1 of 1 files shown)


org.eclipse.lsp4e/src/org/eclipse/lsp4e/LSPEclipseUtils.java
```diff
@@ -228,13 +228,13 @@ public static CompletionParams toCompletionParams(URI fileUri, int offset, IDocu
 		if (document.getLength() > 0) {
 			try {
 				int positionCharacterOffset = offset > 0 ? offset-1 : offset;
-				String positionCharacter = document.get(positionCharacterOffset, 1);
-				if (Chars.contains(completionTriggerChars, positionCharacter.charAt(0))) {
-					param.setContext(new CompletionContext(CompletionTriggerKind.TriggerCharacter, positionCharacter));
+				String triggerCharacter = document.get(positionCharacterOffset, 1);
```
rubenporras, line 231: I would leave the name of the variable as it was before. Would that be good for you?

ghentschke: I'am fine with the old name as well.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LSPEclipseUtils.java
```diff
@@ -228,13 +228,13 @@ public static CompletionParams toCompletionParams(URI fileUri, int offset, IDocu
...
+				if (Chars.contains(completionTriggerChars, triggerCharacter.charAt(0))) {
+					param.setContext(new CompletionContext(CompletionTriggerKind.TriggerCharacter, triggerCharacter));
 				} else {
 					// According to LSP 3.17 specification: the triggerCharacter in CompletionContext is undefined if
 					// triggerKind != CompletionTriggerKind.TriggerCharacter
-					param.setContext(new CompletionContext(CompletionTriggerKind.Invoked, positionCharacter));
+					param.setContext(new CompletionContext(CompletionTriggerKind.Invoked, null));
```
rubenporras, line 237: As I mentioned in https://github.com/eclipse/lsp4e/pull/747#discussion_r1296968755, I think we do not need to pass `null`, could we leave it out and use the one argument constructor?

ghentschke: Ah, good point. Haven't seen that there is a one argument constructor.

rubenporras (approved):
Thanks a lot for the fix

PR #802, opened 2023-09-13
Adding Did Change Configuration feature

ppank5:
The intention of this PR is adding the Did Change Configuration feature of the LSP. The initial/last configuration is sent after the initialized notification message is received, similarly to VS Code. Then each preference store change triggers a `didChangeConfiguration` message to the server. This behavior provides that the language server acquires a meaningful configuration before starting analyzing any file/workspace/project coming from the LSP4E client.
The plugin developers can add their own implementation via a new extension point called `configurationHandler` and by implementing the `IConfigurationHandler ` interface. This interface comes with a default implementation that basically sends empty messages that can be easily neglected by the language server.
I think the idea is understandable from the code, I tried to add meaningful comments as well, however, feedbacks are welcome.

### DIFF
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/IConfigurationHandler.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/IConfigurationHandler.java
@@ -1,0 +1,74 @@
+package org.eclipse.lsp4e;
+
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+
+import org.eclipse.jdt.annotation.NonNull;
+import org.eclipse.jface.preference.IPreferenceStore;
+import org.eclipse.ui.plugin.AbstractUIPlugin;
+
+/**
+ * Configuration handler that enables to send and update
+ * the configuration of the language server.
+ * It also contains a default implementation that can be instantiated.
+ */
+public interface IConfigurationHandler {
+
+	/**
+	 * Returns a {@link Map} containing all preference names and their values.
+	 * Usually used during the initialization of the language server to send
+	 * all the configuration at once.
+	 * @return
+	 */
+	@SuppressWarnings("null")
+	default @NonNull Map<String, Object> getConfiguration() {
+		return Collections.emptyMap();
+	}
+
+	/**
+	 * Returns a {@link Map} containing all preference names and their values.
+	 * Used to send multiple configuration change to the language server at once.
+	 * @return
+	 */
+	@SuppressWarnings("null")
+	default @NonNull Map<String, Object> getConfiguration(final List<String> prefList) {
+		return Collections.emptyMap();
+	}
+
+	/**
+	 * Returns a {@link Map} containing the requested preference name and its value,
+	 * usually used by a preference change listener.
+	 * The Map will be converted to JSON by LSP4J before the transmission.
+	 * @param preferenceName name of preference
+	 * @return {@link Map} containing the requested preference name and its value
+	 */
+	@SuppressWarnings("null")
+	default @NonNull Map<String, Object> getConfiguration(final String prefName) {
+		return Collections.emptyMap();
+	}
+
+	/**
+	 * Returns the preference store containing the configuration of the language server.
+	 * This must be overridden by the plugin developer, usually with the actual
+	 * {@code Activator.getDefault().getPreferenceStore()}
+	 * @see AbstractUIPlugin#getPreferenceStore()
+	 * @return the preference store
+	 */
+	public @NonNull IPreferenceStore getPreferenceStore();
+
+	/**
+	 * Default implementation that can be instantiated
+	 */
+	final static class DefaultConfigurationHandler implements IConfigurationHandler {
+		public DefaultConfigurationHandler() {
+			// Do nothing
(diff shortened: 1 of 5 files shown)


org.eclipse.lsp4e/src/org/eclipse/lsp4e/IConfigurationHandler.java
```diff
@@ -0,0 +1,74 @@
...
+
+	/**
+	 * Returns a {@link Map} containing the requested preference name and its value,
+	 * usually used by a preference change listener.
+	 * The Map will be converted to JSON by LSP4J before the transmission.
+	 * @param preferenceName name of preference
+	 * @return {@link Map} containing the requested preference name and its value
+	 */
+	@SuppressWarnings("null")
+	default @NonNull Map<String, Object> getConfiguration(final String prefName) {
```
zulus, line 47: Some servers asks for concrete preference (for example vue language-server) in this case getConfiguration can be for example boolean

angelozerr:
Please note that WWD manages HTML, CSS, JS settings https://github.com/eclipse-wildwebdeveloper/wildwebdeveloper/blob/bd4d90213bee415549706c875dbe3ce348e26bda/org.eclipse.wildwebdeveloper/src/org/eclipse/wildwebdeveloper/html/HTMLLanguageClient.java#L42

I wonder if this PR will simplify the code of WWD.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServerWrapper.java
```diff
@@ -142,6 +145,15 @@ public void dirtyStateChanged(IFileBuffer buffer, boolean isDirty) {
 
 	};
 
+	private final IPropertyChangeListener propertyChangeListener = new IPropertyChangeListener() {
+		@Override
+		public void propertyChange(PropertyChangeEvent event) {
+			final IConfigurationHandler configHandler = serverDefinition.getConfigurationHandler();
+			languageServer.getWorkspaceService().didChangeConfiguration(
```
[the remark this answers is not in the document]
ppank5: I think any LS can ask for any section, however the LS is not aware of what configuration sections are available in/at the client. So by default the configuration is shared, just need to ask the client properly.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/IConfigurationHandler.java
ppank5: Sorry I don't get this. What do you mean by "can be"? The return value or the parameter or what do you mean?

ppank5:
> Please note that WWD manages HTML, CSS, JS settings https://github.com/eclipse-wildwebdeveloper/wildwebdeveloper/blob/bd4d90213bee415549706c875dbe3ce348e26bda/org.eclipse.wildwebdeveloper/src/org/eclipse/wildwebdeveloper/html/HTMLLanguageClient.java#L42
> 
> I wonder if this PR will simplify the code of WWD.

Based on the code you linked, the WWD LS has to ask for the configuration in every case. Contrarily, the didChangeConfiguration message is a notification about the change of the settings. The former mechanism is not influenced by the latter one. So the LS can have both if the developers like. In our case, the user can have such preferences that influence the code analysis at the server side. So we needed the configuration of the client asap, preferably before starting the analysis. Therefore, we "copied" the VSCode behavior by sending the whole client configuration right after the initialization messages.
Regarding the WWD LS, I think the developers can experiment a bit with this idea. As I see a conversion is needed between `Settings` and `Map<String, Object>`, basically everything else is just boilerplate code.

mickaelistria:
I'm a bit unsure about this one as there are many many things that could trigger a configuration change, and I'm unsure about whether 1 API could successfully encompass all cases. For example. a configuration can be derived from one or more preference store, with or without translation of preference ids, or can be derived from a configuration file, or can be triggered by some custom menu action, or even by a codeAction that would trigger a client side action...
In such case, it's relatively easy to place the right listener already where it fits best (and LSP4E cannot do much to know "where it fits best"), and then to call `LanguageServers.forDocument().withPreferredServerId(lsID).execute(ls -> ls.didChangeConfiguration(...))`.
Unless I'm mistaken, I think the missing brick is more a way to store the current configuration for a language server instance. But I don't think a new API is necessary for it, it could maybe be just a field in the LanguageServerWrapper ?

zulus:
I'm thinking about something different for eclipse-wildwebdeveloper/wildwebdeveloper#1335

1. Additional extension point to collect prefixes (sections) and IConfigurationProvider that can deliver entire tree/map or single value, how it will be stored, we don't care
2. Separate alias mapping (can be in xml directly in same extension point)
3. Basic implementation for IConfigurationProvider that store in IEclipsePreference (PreferenceStore) (boolean, numbers, string) values directly, and do json serialize/deserialize for others
4. Default implementation for "workspace/configuration" that use data from IConfigurationProvider's for LS that prefer PULL's. I read somewhere that is a trend now ;)
5. Helpers for LS that prefer settings in PUSH mode :  initializationOptions or after initialize + watchers

mickaelistria:

PR #808, opened 2023-09-14
Fix and unit test for #804: Race condition in LanguageServerWrapper

### LINKED ISSUE #804: Race condition in LanguageServerWrapper on exception in StreamConnectionProvider.start()
Hi everyone,

I have observed the following behavior when I use a custom StreamConnectionProvider that tries to contact a remote language server using websockets: When the server is down, eventually lsp4e stops trying to recreate the language server connection. This is caused by an exception thrown in StreamConnectionProvider.start() that causes the following race condition:

- In https://github.com/eclipse/lsp4e/blob/9cdd9be00121f8c7fc386763e886d4e6a5f98787/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServerWrapper.java#L326 the start()-exception is handled by calling stop(), which sets `inititalizeFuture = null`
- In https://github.com/eclipse/lsp4e/blob/9cdd9be00121f8c7fc386763e886d4e6a5f98787/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServerWrapper.java#L259 initializeFuture is assigned the value from `CompletableFuture.supplyAsync`, which might happen before or after the start()-Exception has been handled.

Sometimes, the initializeFuture gets assigned the failed future from line 259 after the stop-Method has been called in line 326. When this happens, LanguageServerWrapper.start() does not start a new language server any more, because it only checks for initializeFuture == null in https://github.com/eclipse/lsp4e/blob/9cdd9be00121f8c7fc386763e886d4e6a5f98787/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServerWrapper.java#L256.

One possible solution might be to amend line 256 with something like `&& !this.initializeFuture.isCompletedExceptionally()`.

ChristophKaser:
Fixes #804

### DIFF
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServerWrapper.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServerWrapper.java
@@ -322,10 +322,13 @@ public class LanguageServerWrapper {
 				FileBuffers.getTextFileBufferManager().addFileBufferListener(fileBufferListener);
 			}).exceptionally(e -> {
 				LanguageServerPlugin.logError(e);
-				initializeFuture.completeExceptionally(e);
 				stop();
-				return null;
+				throw new RuntimeException(e);
 			});
+			if (this.initializeFuture.isCompletedExceptionally()) {
+				// This might happen if an exception occurred and stop() was called before this.initializeFuture was assigned
+				this.initializeFuture = null;
+			}
 		}
 	}
 
--- a/org.eclipse.lsp4e.test/fragment.xml
+++ b/org.eclipse.lsp4e.test/fragment.xml
@@ -105,6 +105,17 @@
             contentType="org.eclipse.lsp4e.test.singleton-ls-content-type"
             id="org.eclipse.lsp4e.test.singletonLS">
       </contentTypeMapping>
+       <server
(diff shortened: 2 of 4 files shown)


org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServerWrapper.java
```diff
@@ -322,10 +322,13 @@ public synchronized void start() throws IOException {
 				FileBuffers.getTextFileBufferManager().addFileBufferListener(fileBufferListener);
 			}).exceptionally(e -> {
 				LanguageServerPlugin.logError(e);
-				initializeFuture.completeExceptionally(e);
 				stop();
-				return null;
+				throw new RuntimeException(e);
```
rubenporras, line 326: is there a reason for now just `throw e`?

ChristophKaser: Yes, unfortunately e is of type `Throwable` and `Function.apply` can only throw an unchecked exception.

mickaelistria (approved):
This looks good to me. I leave the final say (and merge if possible) to @rubenporras

rubenporras (approved):

rubenporras:
thanks

ChristophKaser:
Thanks for merging!

Issue #809, opened 2023-09-14
Bug in 'Open Symbol in Workspace' dialog filter


ghentschke:
When filtering for a symbol in workspace via the 'Open Symbol in Workspace' dialog, the requested symbol won't be filtered:
![image](https://github.com/eclipse/lsp4e/assets/123444711/c37a35e2-743b-4117-82a5-f07f41113dc1)
even matches are in the list from the LS, the matches won't be shown on top of the list:
![image](https://github.com/eclipse/lsp4e/assets/123444711/da866c35-eb36-4393-9772-f808dcffc7f8)
I've to scroll a long way down to find them

bug comes with PR #805
The bug came along with this change:
![image](https://github.com/eclipse/lsp4e/assets/123444711/fb8f9c8e-6bc5-40df-947c-a9580b9609ce)

When I am typing `counter` the filter seems to work not for all entries:
![image](https://github.com/eclipse/lsp4e/assets/123444711/6711e2aa-750a-452e-824a-8a2192f7f105)
![image](https://github.com/eclipse/lsp4e/assets/123444711/b5e96f32-c7ac-4826-974a-1ba43531a775)

mickaelistria:
I've recently removed all filtering from client side because in the latest discussion (on jdt-ls end), there was some consensus that filtering has to happen on server side to match things like wildcards or qualified names. The example was about requesting `java.awt.List` which was leading to empty list of symbols.
However, there may be a glitch with it: in case we add a character to the end, it might be that the result from the LS are kept instead of re-requesting the LS, and since there is no filtering, some older response remain.
If this is the case, I think LSP4E should always re-ask the LS for the symbols when modifying the query.

PR #811, opened 2023-09-14
Re-ask LS for workspaceSymbols on every change

mickaelistria:
Do not perform client-side filtering in any case, do not reuse existing responses even if query pattern grows.

Fixes https://github.com/eclipse/lsp4e/issues/809


mickaelistria:
@ghentschke Can you please check it fixes the issue you reported in #809 ?

ghentschke:
Yes, it works. Thank you for the quick fix!!!

Issue #738, opened 2023-07-29
LSP Listener to be notified that a didChange has been done


angelozerr:
In WWD we use XML language server to consume a LSP command from the XML language server to manage auto close tag.

When we consume this LSP command we must be sure that the didChange has been done before consume it, because we must be sure that TextDocument on XML language server side has been updated.

Today we use the uggly code with Thread.sleep https://github.com/eclipse-wildwebdeveloper/wildwebdeveloper/blob/4998f935382c703a7776f37053993eae19848942/org.eclipse.wildwebdeveloper.xml/src/org/eclipse/wildwebdeveloper/xml/internal/autoclose/XMLAutoCloseTagReconciler.java#L129 which could create a deadlock https://github.com/eclipse/lemminx-maven/issues/332#issuecomment-1385482003

PR #812, opened 2023-09-18
Register documentcontentsynchronizer as prenotified listener 

zulus:
Will be called before reconcilers installed via for example addDocumentListener. 

For now reconcilers have to wait couple ms to make sure they will be called after didChange notification

### DIFF
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServerWrapper.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServerWrapper.java
@@ -604,7 +604,7 @@ public class LanguageServerWrapper {
 				TextDocumentSyncKind syncKind = initializeFuture == null ? null
 						: serverCapabilities.getTextDocumentSync().map(Functions.identity(), TextDocumentSyncOptions::getChange);
 				final var listener = new DocumentContentSynchronizer(this, languageServer, theDocument, syncKind);
-				theDocument.addDocumentListener(listener);
+				theDocument.addPrenotifiedDocumentListener(listener);
 				LanguageServerWrapper.this.connectedDocuments.put(uri, listener);
 			}
 		}).thenApply(theVoid -> this);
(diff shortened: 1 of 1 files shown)


org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServerWrapper.java
```diff
@@ -604,7 +604,7 @@ private boolean supportsWorkspaceFolderCapability() {
 				TextDocumentSyncKind syncKind = initializeFuture == null ? null
 						: serverCapabilities.getTextDocumentSync().map(Functions.identity(), TextDocumentSyncOptions::getChange);
 				final var listener = new DocumentContentSynchronizer(this, languageServer, theDocument, syncKind);
-				theDocument.addDocumentListener(listener);
+				theDocument.addPrenotifiedDocumentListener(listener);
```
rubenporras, line 607: the javadoc of  `addPrenotifiedDocumentListener` and `removePrenotifiedDocumentListener` says 	

>  This method is not for public use.

Is that not something we need to respect?

zulus: If not, DefaultDocumentAdapter could be used. It connect to Document via this API and expose TextChangeListener

zulus: Or just tell adopters to not use IDocumentListener and migrate to ITextListener

rubenporras: @mickaelistria , what do you think about this PR?

mickaelistria: I'm not much familiar with this API. If it ensures that `didChange` is called before the reconciliers are invoked, why not... I don't really see how it can hurt, particularly as it doesn't change the state of the IDocument.

rubenporras: I also do not see how it can hurt. @zulus , could we follow on the comment "For now reconcilers have to wait couple ms to make sure they will be called after didChange notification" and change the reconcilers to not wait as part of this PR?

zulus: @rubenporras as an example see #738 If IReconciler use standard path -> ITextViewer -> ITextInputListener -> IDocumentListener, his listener will be installed before LanguageServerWrapper, because both  ConnectDocumentToLanguageServerSetupParticipant and LanguageServerWrapper are async (awit for initialize, delayed excutor etc..)

PR #815, opened 2023-09-19
Fix TokenTypeMapper: Return token from TokenProvider when possible

jnt0r:


org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/semanticTokens/TokenTypeMapper.java
```diff
@@ -39,7 +39,7 @@ public IToken apply(final String tokenType) {
 		if (tmPresentationReconciler != null) {
 			ITokenProvider tokenProvider = tmPresentationReconciler.getTokenProvider();
 			if (tokenProvider != null) {
-				tokenProvider.getToken(tokenType);
+				return tokenProvider.getToken(tokenType);
```
rubenporras, line 42: we would still need to check that the value returned by `tokenProvider.getToken(tokenType)` is not null, right?

jnt0r: The getToken method is the only method not annotated with @Nullable in TokenProvider class.
I can still add the null check.

rubenporras: I see, it returns `ITokenProvider.DEFAULT_TOKEN`, which is defined as 	`IToken DEFAULT_TOKEN = new Token(null)`;

I would then check that it is not the default token, as maybe the next getToken we are now not calling could return something more meaningful.

jnt0r: I don't know. If there is a TokenProvider and it does not return a token, this may be intentional. Then returning a token from the default theme might conflict with the expected behaviour.

rubenporras (approved):

PR #817, opened 2023-09-20
Fix completion

mickaelistria:
In case where the additionalTextEdits adds longer text then document.length() - offset.

Do not shift offset because of additionalTextEdit until we're done reading interesting bits of this document at the target location (eg indentation information)


mickaelistria:
Dup of #816

PR #819, opened 2023-09-21
InsertTextMode.AdjustIndentation when cursor (and completion) is not at the start of the line

ChristophKaser:
Assuming I have a file with the following content:
```html
<p>
   foo<di|
```
(| is the position of the cursor)
and I invoke content assist, which provides a completion `"<div>\n</div>"` with InsertTextMode.AdjustIndentation  .

I would expect the applied completion to look like this:

```html
<p>
   foo<div>
   </div>|
```
But atm I get this:

```html
<p>
   foo<div>
</div>|
```

VSCode on the other hand applies the completion with correct indentation.

This PR fixes the indentation calculation and adds a unit test for it.


mickaelistria:
Thank you!

PR #824, opened 2023-09-25
[PoC] Fix #823 - Add support for willRenameFiles/willCreateFiles/willDeleteFiles

### LINKED ISSUE #823: Add support for willRenameFiles/willCreateFiles/willDeleteFiles
As in subject. LSP4E currently only support basic inline refactoring in editor, could also support RefactoringParticipant inside project explorer

zulus:
This is a poc for rename. Unfortunately changes produced by LSPEclipseUtils.toCompositeChange arent correct:

1. TextDocumentChanges isn't grouped per file, so preview is broken (same document is re-used)
2. If files are opened in editors, change leave them in dirty state
3. In theory "will*" operations should be invoked sequentially, if not two servers can report same change, some de-duplication should be invoked

### DIFF
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/rename/LSPRenameParticipant.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/rename/LSPRenameParticipant.java
@@ -1,0 +1,124 @@
+/*******************************************************************************
+ * Copyright (c) 2023 Dawid Pakuła and others.
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *  Dawid Pakuła - initial implementation
+ *******************************************************************************/
+package org.eclipse.lsp4e.operations.rename;
+
+import java.nio.file.FileSystems;
+import java.util.List;
+import java.util.concurrent.CompletableFuture;
+
+import org.eclipse.core.resources.IFile;
+import org.eclipse.core.runtime.CoreException;
+import org.eclipse.core.runtime.IProgressMonitor;
+import org.eclipse.core.runtime.OperationCanceledException;
+import org.eclipse.lsp4e.LSPEclipseUtils;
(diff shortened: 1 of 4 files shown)


zulus:
>     1. TextDocumentChanges isn't grouped per file, so preview is broken (same document is re-used)

Fixed on platform level (already merged): https://github.com/eclipse-platform/eclipse.platform.ui/pull/1172

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/rename/LSPRenameParticipant.java
```diff
@@ -0,0 +1,124 @@
...
+			}
+			FileOperationOptions willRename = f.getWorkspace().getFileOperations().getWillRename();
+			if (willRename == null) {
+				return false;
+			}
+			if (willRename.getFilters() == null || willRename.getFilters().isEmpty()) {
+				return true;
+			}
+			return willRename.getFilters().stream().anyMatch(filter -> {
+				return FileSystems.getDefault().getPathMatcher("glob:" + filter.getPattern().getGlob()) //$NON-NLS-1$
```
angelozerr, line 56: `FileSystems.getDefault().getPathMatcher("glob:" `have  a limitation, when you define `**/*.ts` and your base dir is 

`C:/Users/`

 * `C:/Users/foo/bar.ts`  will match `**/*.ts`
 * `C:/Users/bar.ts`  will NOT match `**/*.ts` although vscode support it and it sa very commons usecases.

More its is better to cache the globfile pattern.

I suggest that you see my test code at https://github.com/redhat-developer/lsp4ij/blob/main/src/test/java/com/redhat/devtools/lsp4ij/features/filewatchers/FileSystemWatcherManagerTest.java which are the same test than vscode.

In otherwords I suggest that you copy/paste my https://github.com/redhat-developer/lsp4ij/blob/main/src/main/java/com/redhat/devtools/lsp4ij/features/filewatchers/PathPatternMatcher.java and use it.

PathPatternMatcher is used to manage glob pattern for didChangeFilesWatcher and willRenameFiles, etc

Here my current PR to support willRenameFiles https://github.com/redhat-developer/lsp4ij/pull/198

Hope it will help.

mickaelistria:

PR #832, opened 2023-10-02
Enable console for vscode-js-debug

mickaelistria:
* Fix some bugs in case DAP process doesn't exist
* Buffer streams by default so they're not discarded when received before console shows

### DIFF
--- a/org.eclipse.lsp4e.debug/src/org/eclipse/lsp4e/debug/debugmodel/DSPDebugTarget.java
+++ b/org.eclipse.lsp4e.debug/src/org/eclipse/lsp4e/debug/debugmodel/DSPDebugTarget.java
@@ -556,7 +556,11 @@ public class DSPDebugTarget extends DSPDebugElement implements IDebugTarget, IDe
 			}
 		}
 		try {
-			CompletableFuture<ThreadsResponse> threads2 = getDebugProtocolServer().threads();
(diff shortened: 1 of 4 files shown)


org.eclipse.lsp4e.debug/src/org/eclipse/lsp4e/debug/debugmodel/DSPDebugTarget.java
```diff
@@ -628,18 +632,20 @@ public void thread(ThreadEventArguments args) {
...
-				outputted = true;
-			} else if (OutputEventArgumentsCategory.STDERR.equals(args.getCategory())) {
-				dspStreamsProxy.getErrorStreamMonitor().append(output);
-				outputted = true;
-			}
+		DSPStreamsProxy dspStreamsProxy;
+		if (process == null) {
+			process(null); // create dummy process to have a console
+		}
+		dspStreamsProxy = process.getStreamsProxy();
```
jonahgraham, line 639: Move declaration to here of `dspStreamsProxy`

mickaelistria:
Thanks, I'm going to merge it.
I'm not yet confident that this is the best approach, as I'm also working on other issues related to process mapping; but I believe it's still a noticeable enhancement over the current state, so let's merge it.

PR #833, opened 2023-10-03
Provide a Java process command builder.

angelozerr:
This PR provides the capability to build a Java process command with a builder to start a language server written in Java. The advantage of this builder is:

 * share same code for any language server which need to start a Java process
 * in the future, LSP4E could provide an UI (for language server written in Java) to fill debug port. In LSP4IJ we provide this feature and debug info are stored in the language server settings (so no need to implement a custom UI for a language server)
 
image

 * in the future LSP4E could provide a JRE combo to choose which JRE to use to start the language server
 * M2E provide an extension point to override the classpath (to add lemminx-maven JAR in XMLLanguageServer). This extension point could be hosted in LSP4E (we have for instance a MicroProfile language server which add extra JARs to provide Quarkus support into MicroProfile LS)

This builder could simplify code:

 * for XML language server https://github.com/eclipse-wildwebdeveloper/wildwebdeveloper/blob/92ece9fc35e4e028eb249cacf445512bbafac835/org.eclipse.wildwebdeveloper.xml/src/org/eclipse/wildwebdeveloper/xml/internal/XMLLanguageServer.java#L91
 * for Quarkus language server https://github.com/jbosstools/jbosstools-quarkus/blob/b073d1801df135c37656492481cdeab26fb5cb83/plugins/org.jboss.tools.quarkus.lsp4e/src/org/jboss/tools/quarkus/lsp4e/QuarkusLanguageServer.java#L42
 * for Qute language server https://github.com/jbosstools/jbosstools-quarkus/blob/b073d1801df135c37656492481cdeab26fb5cb83/plugins/org.jboss.tools.quarkus.lsp4e/src/org/jboss/tools/quarkus/lsp4e/internal/ls/qute/QuteLanguageServer.java#L45

As you can notice the code is the same (ex : computeJavaPath), etc

### DIFF
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/server/java/JavaProcessCommandBuilder.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/server/java/JavaProcessCommandBuilder.java
@@ -1,0 +1,85 @@
+/*******************************************************************************
+ * Copyright (c) 2023 Red Hat Inc. and others.
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *  Angelo ZERR (Red Hat Inc.) - initial implementation
+ *******************************************************************************/
+package org.eclipse.lsp4e.server.java;
+
+import java.io.File;
+import java.util.ArrayList;
+import java.util.List;
+
+import org.eclipse.core.runtime.Platform;
+
+/**
+ * A builder to create Java process command.
+ */
+public class JavaProcessCommandBuilder {
+
+    private String javaPath;
+
+    private String debugPort;
+
+    private boolean debugSuspend;
+    private String jar;
+
+    private String cp;
+
+    public JavaProcessCommandBuilder() {
+        setJavaPath(computeJavaPath());
+    }
+
+    public JavaProcessCommandBuilder setJavaPath(String javaPath) {
+        this.javaPath = javaPath;
+        return this;
+    }
+
+    public JavaProcessCommandBuilder setDebugPort(String debugPort) {
+        this.debugPort = debugPort;
+        return this;
+    }
+
+    public JavaProcessCommandBuilder setDebugSuspend(boolean debugSuspend) {
+        this.debugSuspend = debugSuspend;
+        return this;
+    }
+
+    public JavaProcessCommandBuilder setJar(String jar) {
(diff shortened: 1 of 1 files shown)


mickaelistria:
I find there is almost always a few things, like a system property, a Java version, some classpath, some other configuration file... to look at that make that there is no strong benefit in having such rationalization here. For instance, inLemMinX, we wouldn't use it as you mention it has some extension to contribute to classpath and those extensions wouldn't necessary make sense for other Java-process-based language servers.

angelozerr:
>  For instance, inLemMinX, we wouldn't use it as you mention it has some extension to contribute to classpath and those extensions wouldn't necessary make sense for other Java-process-based language servers.

You will use it https://github.com/eclipse-wildwebdeveloper/wildwebdeveloper/blob/92ece9fc35e4e028eb249cacf445512bbafac835/org.eclipse.wildwebdeveloper.xml/src/org/eclipse/wildwebdeveloper/xml/internal/XMLLanguageServer.java#L91

And as I tried to explain, for override classpath, I think LSP4E should integrate the M2E extension point which override classpaths and builder could use it, It will avoid coding https://github.com/eclipse-wildwebdeveloper/wildwebdeveloper/blob/92ece9fc35e4e028eb249cacf445512bbafac835/org.eclipse.wildwebdeveloper.xml/src/org/eclipse/wildwebdeveloper/xml/internal/XMLLanguageServer.java#L169 and another LS (like Quarkus could benefit with this extension point) I mean that builder will use this extension point to prefill classpath.

The builder create a standard command list but you can add another command if you wish.

mickaelistria:
> And as I tried to explain, for override classpath, I think LSP4E should integrate the M2E extension point which override classpaths and builder could use it,

I disagree here: the way to extend a Java-based process can differ from a Java process to another. The problem of how to best start a Java LS is not so generic, and the how to extend it is even less. I'm not under the impression that the problems are really generic and hard to solve with existing API, so I'm not convinced LSP4E should try to go further/more specific here.

akurtakov:
What is the status of this one ?

angelozerr:
>What is the status of this one ?

IMHO I think it can be a good idea to have that. We provide that in LSP4IJ an dit is very helpfull. For instance you don't need to take care of generation of debug string command. In LSP4IJ we have a generic debug port and suspend fields and the Java builder  generate the proper command (no need to write a custom UI to fill debugging info for Java process).

The Java builder from LSP4IJ will provide an extension point to add some JAR in classpath which is very helpfull too. I tried to convince LSP4E to adopt this idea by giving concrete sample like WWD for staring lemminx, Qute LS, MicroProfile LS, etc 

All those language server have the same code.

mickaelistria:

PR #834, opened 2023-10-03
Support termination of virtual DSPProcess

mickaelistria:
The DSPProcess not always maps to an actual process, but when it does, we use it to control termination actions and state, and if it doesn't we failback to an internal state.

This makes that the "Terminate" button on the Console works.


org.eclipse.lsp4e.debug/src/org/eclipse/lsp4e/debug/console/DSPProcess.java
```diff
@@ -50,16 +53,19 @@ public <T> T getAdapter(Class<T> adapter) {
...
 	@Override
 	public boolean isTerminated() {
-		return target.isTerminated();
+		return handle.map(h -> !h.isAlive()).orElse(terminated);
 	}
 
 	@Override
 	public void terminate() throws DebugException {
+		terminated = true;
+		handle.ifPresent(ProcessHandle::destroyForcibly);
```
jonahgraham, line 67: Please don't destroyForcibly without trying a clean destroy first. For embedded targets this causes the target boards to be left in an unknown state. 

I, and others, have been working with VSCode to make sure they don't do that either. In the VSCode case the target is terminated and then if the adapter process doesn't quit within 2 seconds the adapter process is forcibly destroyed.

jonahgraham (changes requested):

org.eclipse.lsp4e.debug/src/org/eclipse/lsp4e/debug/console/DSPProcess.java
mickaelistria: I think latest patch implements it.

jonahgraham (approved):
Thanks for making that change.

PR #835, opened 2023-10-03
Introduce universal configuration provider

zulus:
This pr implements:

1. didChangeConfiguration auto registration
2. Support for old style configuration watchers
3. Basic API to implements settings provider on top of EclipsePreferences

### DIFF
--- a/org.eclipse.lsp4e/META-INF/MANIFEST.MF
+++ b/org.eclipse.lsp4e/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: Language Server Protocol client for Eclipse IDE (Incubation)
 Bundle-SymbolicName: org.eclipse.lsp4e;singleton:=true
-Bundle-Version: 0.17.3.qualifier
+Bundle-Version: 0.18.0.qualifier
 Bundle-RequiredExecutionEnvironment: JavaSE-17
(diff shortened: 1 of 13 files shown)


zulus:
See attached implementation for wildwebdeveloper. After this, for example vue language server (because support didChangeConfiguration auto-registration) without any other modification will be able to consume js/ts settings managed by eclipse

zulus:
@angelozerr @mickaelistria any comments?

mickaelistria:
I will try to review it soon

zulus:
knock knock ;)

mickaelistria:
I think the part adding "didChangeConfiguration auto registration" is good and due to properly support LSP.
However, I'm skeptical about adding a custom extension point and native support for preference-based configuration directly in LSP4E now, as there is only 1 LS that has expressed the need for it, and the mapping rules happens to be simple for this LS, but may not be in most cases (ie a configuration int from the LS could be turned into a string for the IDE preferences and other more complex mapping can take place, some of them than cannot be captured by LSP4E).
So can you please trim out that part the particular part about Eclipse Preferences and move it to client ( https://github.com/eclipse-wildwebdeveloper/wildwebdeveloper/pull/1356 ) to see how things behave? I believe it would keep things simpler and more powerful if LSP4E only adds extensibility so consumers can push/fetch/react to configuration in their own code.

zulus:
Sure, I had a plan to maybe in future implement default provider for "proxy.*" configs, but even they don't use IEclipsePreference ;)

mickaelistria:

PR #837, opened 2023-10-04
Do not try to refresh threads to compute resume/suspend

mickaelistria:
Those operations typically work on the list of *known* threads, there is no need to refresh the threads at that time, and in case DAP's threads() request doesn't run well on non-suspended targets, then the former implementation could lead to bogus states.

### DIFF
--- a/org.eclipse.lsp4e.debug/src/org/eclipse/lsp4e/debug/debugmodel/DSPDebugTarget.java
+++ b/org.eclipse.lsp4e.debug/src/org/eclipse/lsp4e/debug/debugmodel/DSPDebugTarget.java
@@ -21,23 +21,22 @@ import java.io.InputStream;
 import java.io.OutputStream;
 import java.io.PrintWriter;
 import java.util.Arrays;
-import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Map;
+import java.util.Map.Entry;
 import java.util.Objects;
 import java.util.Set;
 import java.util.TreeMap;
-import java.util.Map.Entry;
 import java.util.concurrent.CompletableFuture;
-import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
-import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.function.Predicate;
 import java.util.function.Supplier;
 import java.util.function.UnaryOperator;
+import java.util.stream.Collectors;
 
 import org.eclipse.core.resources.IMarkerDelta;
 import org.eclipse.core.runtime.CoreException;
@@ -79,7 +78,6 @@ import org.eclipse.lsp4j.debug.TerminateArguments;
 import org.eclipse.lsp4j.debug.TerminatedEventArguments;
 import org.eclipse.lsp4j.debug.Thread;
 import org.eclipse.lsp4j.debug.ThreadEventArguments;
-import org.eclipse.lsp4j.debug.ThreadsResponse;
 import org.eclipse.lsp4j.debug.launch.DSPLauncher;
 import org.eclipse.lsp4j.debug.services.IDebugProtocolClient;
 import org.eclipse.lsp4j.debug.services.IDebugProtocolServer;
@@ -130,10 +128,6 @@ public class DSPDebugTarget extends DSPDebugElement implements IDebugTarget, IDe
 	 * they are up to date (against the {@link #refreshThreads} flag).
 	 */
 	private final Map<Integer, DSPThread> threads = Collections.synchronizedMap(new TreeMap<>());
-	/**
-	 * Set to true to update the threads list from the debug adapter.
-	 */
-	private final AtomicBoolean refreshThreads = new AtomicBoolean(true);
 
 	private boolean fTerminated = false;
 	private boolean fSentTerminateRequest = false;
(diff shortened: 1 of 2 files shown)


mickaelistria:
@jonahgraham This has become an issue with threads() not answering (and `theads.get()` stuck) in vscode-js-debug. I think my proposal makes sense, but I would feel better if you can review it quickly to at least tell me the reasoning doesn't seem too wrong.

mickaelistria:
@jonahgraham This implements your proposal and seems to work fine for vscode-js-debug.

jonahgraham:
(We need automated tests as that would have immediately identified the probable regression here - I will talk to Renesas to see if I can get some funded time to work on that.)

mickaelistria:
OK, here is another attempt which I believe matches the spec (query threads on stopped or on threads event)

org.eclipse.lsp4e.debug/src/org/eclipse/lsp4e/debug/debugmodel/DSPDebugTarget.java
```diff
@@ -632,8 +587,24 @@ public void exited(ExitedEventArguments args) {
...
+		return getDebugProtocolServer().threads().thenAcceptAsync(threadsResponse -> {
+			var threadIds = Arrays.stream(threadsResponse.getThreads()).map(Thread::getId).collect(Collectors.toSet());
+			if (threads.keySet().removeIf(Predicate.not(threadIds::contains))) {
+				fireChangeEvent(DebugEvent.CONTENT);
+			}
+			for (Thread thread : threadsResponse.getThreads()) {
+				DSPThread dspThread = threads.get(thread.getId());
+				if (dspThread == null) {
+					dspThread = new DSPThread(this, thread.getId());
+					fireChangeEvent(DebugEvent.CONTENT);
```
jonahgraham, line 603: I think doing this in the loop means that there could be a bunch of flickering if lots of threads cause lots of fireChangeEvents.

org.eclipse.lsp4e.debug/src/org/eclipse/lsp4e/debug/debugmodel/DSPDebugTarget.java
```diff
@@ -632,8 +587,24 @@ public void exited(ExitedEventArguments args) {
 
 	@Override
 	public void thread(ThreadEventArguments args) {
-		refreshThreads.set(true);
-		fireChangeEvent(DebugEvent.CONTENT);
+		triggerUpdateThreads();
```
jonahgraham, line 590: Note that in case where multiple threads are started in quick succession this will cause multiple threads() calls to be made in quick succession.

org.eclipse.lsp4e.debug/src/org/eclipse/lsp4e/debug/debugmodel/DSPDebugTarget.java
```diff
@@ -632,8 +587,24 @@ public void exited(ExitedEventArguments args) {
...
+	private CompletableFuture<?> triggerUpdateThreads() {
+		return getDebugProtocolServer().threads().thenAcceptAsync(threadsResponse -> {
+			var threadIds = Arrays.stream(threadsResponse.getThreads()).map(Thread::getId).collect(Collectors.toSet());
+			if (threads.keySet().removeIf(Predicate.not(threadIds::contains))) {
+				fireChangeEvent(DebugEvent.CONTENT);
+			}
+			for (Thread thread : threadsResponse.getThreads()) {
+				DSPThread dspThread = threads.get(thread.getId());
+				if (dspThread == null) {
+					dspThread = new DSPThread(this, thread.getId());
```
jonahgraham, line 602: I may be missing something here, but when is this new dspThread added to threads?

jonahgraham (changes requested):
This looks better, and I have re-read the spec and AFAICT this does implement the spec, but in practice debug adapters don't need to send an initial stopped or thread event to kick things off. VSCode always follows "configurationDone" with a "threads" - which means the debug adapter doesn't need to provide a stopped or thread event to get its initial state correct.

---

I am concerned over the removal of the synchronization in this code. IIUC the update you have here the contents of `thenAcceptAsync` (in triggerUpdateThreads) may be running in multiple threads simultaneously as triggerUpdateThreads may be called multiple times in a row before the async part of previous calls has finished.

org.eclipse.lsp4e.debug/src/org/eclipse/lsp4e/debug/debugmodel/DSPDebugTarget.java
mickaelistria: Ack

mickaelistria:
New patch set re-adds synchronization and implements your 2 suggestions/fixes. For the multiple request to threads() in case multiple notification are sent; I'm not sure what would be the best solution.
Do you think it's something we need to fix right now? If it's acceptable, I can leave a comment about this in the code and we can optimize it later when needed.

jonahgraham:
> Do you think it's something we need to fix right now?

No.

jonahgraham:
This looks ok to me, just this [comment](https://github.com/eclipse/lsp4e/pull/837#pullrequestreview-1660284511) unresolved:

> This looks better, and I have re-read the spec and AFAICT this does implement the spec, but in practice debug adapters don't need to send an initial stopped or thread event to kick things off. VSCode always follows "configurationDone" with a "threads" - which means the debug adapter doesn't need to provide a stopped or thread event to get its initial state correct.

The current version won't display threads correctly for such debug adapters, but they will work in VSCode.

mickaelistria:
OK, I've added an instruction to trigger a threads update upon initialization.

jonahgraham (approved):

mickaelistria:
Thank you very much for you review @jonahgraham !

Issue #839, opened 2023-10-05
Restart LS via LanguageServerWrapper


ddscharfe:
In cdt-lsp we want to restart the language server when a configuration file changes (https://github.com/eclipse-cdt/cdt-lsp/issues/214). 

This was already discussed in #726. 
Calling `LanguageServerWrapper::start` and then `LanguageServerWrapper::stop` is not sufficient, because this will not refresh the open editors instantly. 
What we'd like to have is a method `LanguageServerWrapper::restart()`, which stops and starts the language server and reconnects the documents. `LanguageServerWrapper::start()` already supports this, but only restarts if the server is inactive. 

I will provide a PR if there are no objections to this feature.

sebthom:
This is implemented since [0.24.3](https://github.com/eclipse/lsp4e/releases/tag/0.24.3)

PR #840, opened 2023-10-06
[#839] Add restart method

ddscharfe:
Implements #839

### DIFF
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServerWrapper.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServerWrapper.java
@@ -242,9 +242,33 @@ public class LanguageServerWrapper {
 	 * @throws IOException
 	 */
 	public synchronized void start() throws IOException {
+		start(false);
+	}
+
+	/**
+	 * Restarts a language server. If language server is not started, calling this
+	 * method is the same as calling {@link #start()}.
+	 *
+	 * @throws IOException
+	 * @since 0.18
+	 */
+	public synchronized void restart() throws IOException {
+		start(true);
+	}
+
+	/**
+	 * Starts a language server and triggers initialization. If language server is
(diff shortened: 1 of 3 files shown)


org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServerWrapper.java
```diff
@@ -242,9 +242,32 @@ private List<WorkspaceFolder> getRelevantWorkspaceFolders() {
...
 	 */
 	public synchronized void start() throws IOException {
+		start(false);
+	}
+
+	/**
+	 * Restarts a language server. If language server is not started, calling this
+	 * method is the same as calling {@link #start()}.
+	 *
+	 * @throws IOException
```
mickaelistria, line 252: Please add `@since 0.18`, while LSP4E is not best at API versioning, those are getting stable and we should try to leverage good practices whenever prossible.

mickaelistria (changes requested):
Version must be bumped to 0.18.0 since it's the 1st change, API addition, since last release of LSP4E.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServerWrapper.java
ddscharfe: Okay, this makes sense. I wasn't aware that LanguageServerWrapper is API, using it from our cdt plugin gives discouraged access warning.

mickaelistria: > I wasn't aware that LanguageServerWrapper is API

As mentioned above, LSP4E is not really clean in term of API ;) But lately it seems like there is a consensus that LanguageServers and LanguageServerWrapper can be considered as API.

mickaelistria:
The version in pom file also needs to be modified.

mickaelistria:
Thank you!

PR #841, opened 2023-10-06
Document content synchronizer

rubenporras:
Prevent failure of DocumentContentSynchronizer.<init> if the document
does not have an URI registered in EFS and it is not a file uri.

Also improve the existing log to include the URI that we are failing to
deal with.


rubenporras:
I wonder if we actually need a warning or if we just can take 0L as default without bothering the user with a warning. What do you think @mickaelistria ?

mickaelistria:
I don't think I ever encountered this use-case, so I don't have anything relevant to answer here. I think this patch is fine as it, so if you think it's worth merging it now, please do. If you think it's better to not log, then don't log and merge it. I fully trust your judgement here ;)

rubenporras:
Thanks, I think the code can handle the case quite well by taking the default value of `0L`, so I will not give the warning.

PR #844, opened 2023-10-12
Improved parsing of completions with InsertTextFormat.Snippet

ChristophKaser:
Hello everyone,

we are making extensive use of completions with InsertTextFormat.Snippet and had some problems with completions that try to insert keys with $ in them, because `\$` was inserted as is.

So I tried to improve the parsing of snippets to be closer to the spec (https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#snippet_syntax) and/or closer to what VSCode does.

Since the snippet format is fairly complex, I extracted the parser into its own class, so I could keep the current parser offset in a field.

This PR fixes
- variables with default values (like `${TM_SELECTED_TEXT:defaultval}`}
- escaping of special characters like dollar, brackets and comma with \ (like `special\$key`)

The following complicated parts of the snippet format are not yet implemented:
- Variable transforms (like `${TM_FILENAME/(.*)\..+$/$1/}`)
- Nested tabstops in placeholders (like `${1:Place${2:holder}}`)
- Nested tabstops in default values (like `${TM_SELECTED_TEXT:Place${1:holder}}`)

### DIFF
--- a/org.eclipse.lsp4e.test/META-INF/MANIFEST.MF
+++ b/org.eclipse.lsp4e.test/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: Tests for language server bundle (Incubation)
 Bundle-SymbolicName: org.eclipse.lsp4e.test;singleton:=true
-Bundle-Version: 0.15.6.qualifier
+Bundle-Version: 0.15.7.qualifier
 Fragment-Host: org.eclipse.lsp4e
 Bundle-Vendor: Eclipse.org
 Bundle-RequiredExecutionEnvironment: JavaSE-17
--- a/org.eclipse.lsp4e.test/pom.xml
+++ b/org.eclipse.lsp4e.test/pom.xml
@@ -8,7 +8,7 @@
 	</parent>
 	<artifactId>org.eclipse.lsp4e.test</artifactId>
 	<packaging>eclipse-test-plugin</packaging>
-	<version>0.15.6-SNAPSHOT</version>
+	<version>0.15.7-SNAPSHOT</version>
 
 	<properties>
 		<os-jvm-flags /> <!-- for the default case -->
--- a/org.eclipse.lsp4e.test/src/org/eclipse/lsp4e/test/completion/CompleteCompletionTest.java
+++ b/org.eclipse.lsp4e.test/src/org/eclipse/lsp4e/test/completion/CompleteCompletionTest.java
@@ -27,6 +27,7 @@ import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashSet;
 import java.util.List;
+import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ExecutionException;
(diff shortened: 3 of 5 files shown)


angelozerr:
I did a similar work for IJ and I created a snippet parser at https://github.com/redhat-developer/intellij-quarkus/blob/main/src/main/java/com/redhat/devtools/intellij/lsp4ij/operations/completion/snippet/LspSnippetParser.java

This parser parse characters by characters the snippet syntax and don't use some string substring.

IMHO I think this parser should belong to LSP4J. See my comments at https://github.com/eclipse-lsp4j/lsp4j/issues/149#issuecomment-1636071597

mickaelistria:
@ChristophKaser Can you please check whether you can reproduce the test failure exposed by https://ci.eclipse.org/lsp4e/job/lsp4e-github/job/PR-844/1/display/redirect locally?

ChristophKaser:
> @ChristophKaser Can you please check whether you can reproduce the test failure exposed by https://ci.eclipse.org/lsp4e/job/lsp4e-github/job/PR-844/1/display/redirect locally?

Yes, I can reproduce it locally. I had missed that test class and only tested against CompleteCompletionTest to save time, sorry.
This seems to be one of the cases I did not implement (Nested expressions in placeholders), I will try to at least support variables there.

ChristophKaser:
I don't see any test failures any more (the build failure is due to some issue with the version number that I don't really understand).

BTW, it seems the test testVariableReplacement is duplicated in CompleteCompletionTest and VariableReplacementTest.

ChristophKaser:
> I did a similar work for IJ and I created a snippet parser at https://github.com/redhat-developer/intellij-quarkus/blob/main/src/main/java/com/redhat/devtools/intellij/lsp4ij/operations/completion/snippet/LspSnippetParser.java
> 
> This parser parse characters by characters the snippet syntax and don't use some string substring.
> 
> IMHO I think this parser should belong to LSP4J. See my comments at [eclipse-lsp4j/lsp4j#149 (comment)](https://github.com/eclipse-lsp4j/lsp4j/issues/149#issuecomment-1636071597)

I wish I had seen that before I implemented the parser from scratch ;)
However, there are some differences between my implementation and yours, especially if the snippet is not syntactically valid. I tried to reproduce the way VScode handles those snippets (by keeping invalid expressions as-is).

org.eclipse.lsp4e.debug/META-INF/MANIFEST.MF
```diff
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: Debug Adapter client for Eclipse IDE (Incubation)
 Bundle-SymbolicName: org.eclipse.lsp4e.debug;singleton:=true
-Bundle-Version: 0.15.4.qualifier
```
mickaelistria, line 5: This one doesn't need to be changed as you didn't modify it. Only the bundles you change needs to be upversioned (if it wasn't already done since last release)

mickaelistria (changes requested):
The version bump do touch the wrong bundles. According to the former build error, only org.eclipse.lsp4e.tests needs to be upversioned.

ChristophKaser:
> The version bump do touch the wrong bundles. According to the former build error, only org.eclipse.lsp4e.tests needs to be upversioned.

Sorry, I got mixed up in my changes and the rebase...
I have now rebased my branch on master, squashed the commits to avoid multiple commits for one logical change and bumped the version of only org.eclipse.lsp4e.tests

mickaelistria:
Thanks, it's all good now!

PR #855, opened 2023-10-18
Ensure that active highlights are removed when a source view is closed

joaodinissf:
This commit resolves a bug related to erroneously persisted highlights
in source files. It includes a unit test that verifies the correctness
of the fix, and a few minor improvements to the testing framework.

---

Bug Description: Before this change, source highlights would sometimes
persist after a source view was closed. This would happen, for example,
after the following sequence of steps:

1. Open a source. Ensure that some keyword is highlighted.
2. Activate the option "Toggle Split Editor (Horizontal)".
3. In the newly opened source view, highlight a different keyword.
4. At this point, both sets of highlights (the one from step 1 and the
one from step 3) should be visible in both source views. This is correct
behavior.
5. Close one of the source views, while keeping the other open. Before
this commit, both sets of highlights would persist. This is incorrect:
only the set of highlights related to the source view that remains open
should be displayed.

Cause: Each instance of HighlightReconcilingStrategy adds/removes
highlights independently, and keeps track internally of all the
highlights it has created. Before this commit, uninstalling a
HighlightReconcilingStrategy did not remove the active highlights.

Solution: When a source view is closed, all active highlights associated
with its HighlightReconcilingStrategy should be removed. We accomplish
this by calling `removeOccurrenceAnnotations()` during the `uninstall()`
method of `HighlightReconcilingStrategy.java`.

Testing: A unit test was created which approximately replicates the
sequence of steps outlined above. To implement this unit test, the
following changes were made in `org.eclipse.lsp4e.test`:

1. In `HighlightTest.java`, the function `assertAnnotationDoesNotExist`
was created (for consistency, the existing function
`assertHasAnnotation` was renamed to `assertAnnotationExists`).
2. The `mockDocumentHighlights` returned by the
`MockTextDocumentService` are now `Position`-dependent. The unit tests
in `HighlightTest.java` were adapted accordingly.

### DIFF
--- a/org.eclipse.lsp4e.test/META-INF/MANIFEST.MF
+++ b/org.eclipse.lsp4e.test/META-INF/MANIFEST.MF
@@ -25,5 +25,7 @@ Require-Bundle: org.eclipse.core.runtime,
  org.eclipse.jdt.annotation;bundle-version="[2.0.0,3.0.0)",
  org.eclipse.ui.tests.harness,
  org.eclipse.ui.monitoring,
- org.eclipse.core.variables
+ org.eclipse.core.variables,
+ org.eclipse.e4.ui.model.workbench,
+ org.eclipse.e4.ui.workbench
 Automatic-Module-Name: org.eclipse.lsp4e.test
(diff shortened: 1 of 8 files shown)


rubenporras (approved):
Thanks

rubenporras:
@joaodinissf , you need to bump the version in the MANIFEST.MF and pom.xml of org.eclipse.lsp4e.tests.mock for the build to pass.

rubenporras (approved):

Issue #859, opened 2023-10-20
Limited Functionality of LSP Editors in MultiPart Editors


ddscharfe:
LSP editors, when embedded in MultiPage editors, are not fully functional. While some operations such as formatting work correctly, other features including call hierarchy, find references, renaming, don`t work at all.

sebthom:
@ddscharfe can this issue be closed?

PR #860, opened 2023-10-20
Limited Functionality of LSP Editors in MultiPart Editors #859

ddscharfe:
Fix for #859.

There are multiple occasions in the code where editor parts are casted to ITextEditor. As a result, lsp editors embedded in MultiPage editors don't have the full functionality.
Additionally many handlers use UI::getActiveTextEditor, IMHO getting the editor from ExecutionEvent would be cleaner.

As most handlers (except one) work with ITextEditor, I added a convenience execution method to `LSPDocumentAbstractHandler` for this use case. I refactored the existing handlers to use this method, which was basically removing the casts.

I created a new method UI::asTextEditor as central place for obtaining ITextEditor from an IEditorPart. This method could additionally try to adapt to ITextEditor, which would enable more use cases, but I didn't want to introduce this for this PR.

### DIFF
--- a/org.eclipse.lsp4e/META-INF/MANIFEST.MF
+++ b/org.eclipse.lsp4e/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: Language Server Protocol client for Eclipse IDE (Incubation)
 Bundle-SymbolicName: org.eclipse.lsp4e;singleton:=true
-Bundle-Version: 0.18.0.qualifier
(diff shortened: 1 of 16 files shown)


ddscharfe:
Converted to draft, because I have to fix some additional stuff regarding the property testers.

ddscharfe:
I added testing the `activeEditorInput` for the `org.eclipse.lsp4e.textSelectionHasLanguageServer` core expression and therefore also adapted `HasLanguageServerPropertyTester`.  I also added another utility method `UI::asTextViewer` similar to `UI::asTextEditor`.

mickaelistria:
All good to me, thanks a lot!

Issue #725, opened 2023-07-07
CodeActions pop-up refreshed too often


mickaelistria:
Using eclipseide-jdtls, I see that the quickfixes/codeActions popup gets refreshed too often. I'm not sure it's LSP4E's fault only, but since there are recent changes in the popup refresh and I cannot reproduce it without LSP4E-based editor, it can be related.

mickaelistria:
Another symptom is that pressing Escape to close the popup makes the popup disappear, then reappear immediately.

mickaelistria:
Here is the call stack I see on each popup, and so that I see (hit the breakpoint in refreshProposals) repeatedly without good reason
```
Thread [LS-com.redhat.eclipseide.jdtlsclient.inProcessServer#listener-0] (Suspended (entry into method refreshProposals in LSPCodeActionQuickAssistProcessor))	
	LSPCodeActionQuickAssistProcessor.refreshProposals(IQuickAssistInvocationContext) line: 204	
	LSPCodeActionQuickAssistProcessor.lambda$10(IQuickAssistInvocationContext, Void, Throwable) line: 180	
	0x0000000101a75c78.accept(Object, Object) line: not available	
	CompletableFuture<T>.uniWhenComplete(Object, BiConsumer<? super T,Throwable>, UniWhenComplete<T>) line: 863	
	CompletableFuture$UniWhenComplete<T>.tryFire(int) line: 841	
	RemoteEndpoint$1(CompletableFuture<T>).postComplete() line: 510	
	RemoteEndpoint$1(CompletableFuture<T>).complete(T) line: 2179	
	RemoteEndpoint.handleResponse(ResponseMessage) line: 212	
	RemoteEndpoint.consume(Message) line: 193	
	LanguageServerWrapper.lambda$3(MessageConsumer, URI, Message) line: 309	
	0x0000000100a3ac78.consume(Message) line: not available	
	StreamMessageProducer.handleMessage(InputStream, StreamMessageProducer$Headers) line: 194	
	StreamMessageProducer.listen(MessageConsumer) line: 94	
	ConcurrentMessageProcessor.run() line: 113	
	Executors$RunnableAdapter<T>.call() line: 572	
	FutureTask<V>.run() line: 317	
	ThreadPoolExecutor.runWorker(ThreadPoolExecutor$Worker) line: 1144	
	ThreadPoolExecutor$Worker.run() line: 642	
	Thread.runWith(Object, Runnable) line: 1596	
[...]

PR #868, opened 2023-11-10
Prevent recurring pop-up

mickaelistria:
Do not retrigger LS queries (and reactions) if the context has not changed.

Fixes https://github.com/eclipse/lsp4e/issues/725


mickaelistria:
Test failure is not related, test pass locally, let's merge.

PR #869, opened 2023-11-11
Fix broken key description

Bananeweizen:
Localized string exists, but was not referenced correctly.


Bananeweizen:
Looks like this before: 
grafik

mickaelistria:
Thank you!

PR #871, opened 2023-11-11
[build] Fix sha256 checksums for highlight.js

vrubezhny:
Need to set `<skipCache>true</skipCache>` for  `download-maven-plugin` , otherwise,
when building locally  (not removing `org.eclipse.lsp4e/resources` folder contents) ,
the download plugin skips the required files from being downloaded and, because of that,
fails  on `sha256` checksum calculation.

### DIFF
--- a/org.eclipse.lsp4e/pom.xml
+++ b/org.eclipse.lsp4e/pom.xml
@@ -9,9 +9,13 @@
 	<packaging>eclipse-plugin</packaging>
 	<version>0.18.1-SNAPSHOT</version>
 
+	<properties>
+		<forceSkipCache>false</forceSkipCache>
+	</properties>
+
 	<build>
 		<plugins>
-			<plugin>
+			<!-- plugin>
 				<groupId>org.apache.maven.plugins</groupId>
 				<artifactId>maven-clean-plugin</artifactId>
 				<version>3.3.2</version>
@@ -22,17 +26,18 @@
 						</fileset>
 					</filesets>
 				</configuration>
-			</plugin>
+			</plugin -->
 			<plugin>
 				<groupId>com.googlecode.maven-download-plugin</groupId>
 				<artifactId>download-maven-plugin</artifactId>
 				<!-- 1.7.0 is broken, see https://github.com/maven-download-plugin/maven-download-plugin/releases/tag/1.7.0 -->
-				<version>1.7.1</version>
+				<version>1.6.8</version>
 				<configuration>
 					<!-- https://maven-download-plugin.github.io/maven-download-plugin/docsite/1.6.8/wget-mojo.html -->
 					<alwaysVerifyChecksum>true</alwaysVerifyChecksum>
 					<overwrite>true</overwrite>
-					<skipCache>false</skipCache>
+					<!-- skipCache>${forceSkipCache}</skipCache -->
+					<followRedirects>tue</followRedirects>
 					<readTimeOut>5000</readTimeOut>
 				</configuration>
 				<executions>
@@ -44,7 +49,7 @@
 						</goals>
 						<configuration>
 							<url>https://unpkg.com/@highlightjs/cdn-assets@11.9.0/highlight.min.js</url>
-							<sha256>837a6fa5b0c736b52bbde2b2b6190f305da3fc9ed41681db5321507057b5c846</sha256>
+							<sha256>f178f9e890cb9129fc25944d1ea1f1f4d556af0783ec5a867a930d5cdccabab4</sha256>
 							<outputDirectory>resources/highlight.min.js</outputDirectory>
 						</configuration>
 					</execution>
@@ -56,7 +61,7 @@
 						</goals>
 						<configuration>
 							<url>https://unpkg.com/@highlightjs/cdn-assets@11.9.0/styles/default.min.css</url>
-							<sha256>fbde0ac0921d86c356c41532e7319c887a23bd1b8ff00060cab447249f03c7cf</sha256>
+							<sha256>66e72de7f9149fe9ea64702180eb56c25e2a886825ffab790b0ad6d18a33bca5</sha256>
 							<outputDirectory>resources/highlight.min.js/styles</outputDirectory>
 						</configuration>
 					</execution>
@@ -68,7 +73,7 @@
 						</goals>
 						<configuration>
 							<url>https://unpkg.com/@highlightjs/cdn-assets@11.9.0/styles/dark.min.css</url>
-							<sha256>bf437be81145907d1d081f1b52be1c1d254df00ff309a3a8a4cb92989595ff9c</sha256>
+							<sha256>8e52f19e63ccdc4625637db7de06d5494a5aea0694d2c56248f1d46e20aa0885</sha256>
 							<outputDirectory>resources/highlight.min.js/styles</outputDirectory>
 						</configuration>
 					</execution>


sebthom:
This will prevent offline builds and potentially slows down rebuilds. I would prefer if a maven property is used whose default value is true and that can be overriden on the command line in case checksum errors occur to skip the cache.

vrubezhny:
@sebthom agree.  Using a maven property would be a better solution.

vrubezhny:
That's strange... Everything should work normally with
```
		<alwaysVerifyChecksum>true</alwaysVerifyChecksum>
		<overwrite>true</overwrite>
```
... if a check sum doesn't match for a file - such file is to be re-downloaded. and this looks like works for me locally until I have to change the files (when a new version arrives)...

I download the new files with  `download-maven-plugin` (skipping the caching) and calculate their sha256 (with fedora's `sah256sum` utility) - this gives me the new sha256 check sums as follows:
``` 
$ sha256sum  highlight.min.js styles/*
f178f9e890cb9129fc25944d1ea1f1f4d556af0783ec5a867a930d5cdccabab4  highlight.min.js
8e52f19e63ccdc4625637db7de06d5494a5aea0694d2c56248f1d46e20aa0885  styles/dark.min.css
66e72de7f9149fe9ea64702180eb56c25e2a886825ffab790b0ad6d18a33bca5  styles/default.min.css
```
I save these sha256 check sums to `pom.xml` and repeatedly build with the mentioned preferences (always check and override) - everything is checked OK and builds successfully. 
Then I push my PR to GitHub and... sha256 verification fails when building the PR on Jenkins (I suppose it should re-download the files there from scratch as `/resources/highlight.min.js/` directory is added to the ignore list and should never be cloned from the repo on build) ...

So, I remove the contents of my local `org.eclipse.lsp4e/resources.highlighth.min.js/` and repeat the build locally - and see exactly the same error as Jenkins has: the previous version of the file  "magically" appeared in `resources/highlight.js/` with its `8137d51d5ac38acaaa3bfad9f96eb88428c76b39a528cf95070c04c50daf2776` sha256 check sum - so I suppose the file was not re-downloaded (while it should be re-downloaded, because the new sha256 differs), but is taken (probably) from the download plugin's cache.

And this looks like a problem in ` download-maven-plugin` - no way to make it work, other than setting it to `<skipCache>true</skipCache >`, but this means that the cache itself will never be correctly updated and no "offline" builds possible

mickaelistria:
I'd prefer we avoid skipCache=true; and instead try to get the cache to work. As explained by @sebthom , the cache -which seems persisted between rebuilds- is supposed to drastically improve builds.
But one issue with those highlight.js files is that it looks like they can change remotely without a new version (that is not great, but it's OK) without changing URL, and the WGet plugin doesn't seem to like it: I think it has stored the older file in cache and will use it anyway and reports that the file it got (locally from cache) doesn't match the expected checksum (from remote), which is indeed wrong.
Maybe it's worth reporting this issue to highlight.js?
Those cases are relatively non-standard though. I would rather see a way to trigger a `master` build with an option to clear the local cache occasionally when we notice such issue.

vrubezhny:
@mickaelistria I don't know a way then to force downloader to download a fresh version of Highlight.js on the "clean"  run (as Jenkins job does).

Problem is that if a newer file version exists on CDN and we put its sha256 into downloader's config,   that new version doesn't  get downloaded, instead the downloader picks up an old cached version, checks its sha256 against the pom.xml value and says " sha256 is not correct so I give up".

I'm empty of ideas on how can I make the downloader to work correctly.

 The only solution I see now is getting rid of using downloader plugin by storing the current latest Highlight.js js/css files to LSP4E repo and creating a GihHub action that will check periodically  for possible updates and creating according PRs to update the js/css to newer version( (the only problem is that someone will have to create the IP Team requests for the update PRs created by such an action).

vrubezhny:
PS: For better understanding of the problem... Currently Jenkins job builds against Highlight.js v.11.8.0, not the latest v11.9.0 because downloader pick ups that old version from its cache and doesn't pick up the newer one if I provide its URL and sha256 checksums in the downloader's config. It always picks up the older version from cache and fails to download the version we've requested.

mickaelistria:
In the Jenkins file, it's possible to add a boolean parameter "clearM2Workspace", default to `false`, and add a line `sh '[[ $clearM2Workspace == true ]] && rm -r $WORKSPACE/.m2/repository`

vrubezhny:
> In the Jenkins file, it's possible to add a boolean parameter "clearM2Workspace", default to `false`, and add a line `sh '[[ $clearM2Workspace == true ]] && rm -r $WORKSPACE/.m2/repository`

It's not about `$WORKSPACE/.m2/repository`, it's about ``$WORKSPACE/.m2/repository/.cache/download-maven-plugin/*`, no need to kill the whole repository - it's probably cached and restored between the build runs.

mickaelistria:
> It's not about $WORKSPACE/.m2/repository, it's about $WORKSPACE/.m2/repository/.cache/download-maven-plugin/*`, no need to kill the whole repository - 

OK, the flag could only delete this area.

> And problem of the downloader is that it doesn't check the source URL from which a file was downloaded - it checks the only file name (and path probably) and its check sum. If it could check the URI (from which a file was requested for download) - there wouldn't be such a problem.

It does check the URL, but the issue is that the content pointed by this URL can change, so the remote checksum changes too while the cache keeps the older artifact with the older checksum. So old file is found and verified with older checksum; and if you change the checksum then you get the old file and the newer checksum. Removing the cache directory allows to get the newer file in.

vrubezhny:
> It does check the URL, but the issue is that the content pointed by this URL can change,

Nope. Every new version of Highlight.js is downloaded from a different URL

mickaelistria:
> Nope. Every new version of Highlight.js is downloaded from a different URL

What I'm saying is the other way round: sometimes, content of a given URL for 1 highlight.js URL, the content can change slightly over time. The current timestamp we have for the current URL we have were the correct pair some time ago, but remote content seems to have changed a bit and thus both are changed; and the cache doesn't support the case of remote content changing for a given URL.

vrubezhny:
> > Nope. Every new version of Highlight.js is downloaded from a different URL
> 
> What I'm saying is the other way round: sometimes, content of a given URL for 1 highlight.js URL, the content can change slightly over time. The current timestamp we have for the current URL we have were the correct pair some time ago, but remote content seems to have changed a bit and thus both are changed; and the cache doesn't support the case of remote content changing for a given URL.

This could happen, of course, but the case I currently have is about downloader  is picking up from its cache a file downloaded from:
- URL: https://unpkg.com/@highlightjs/cdn-assets@**_11.8.0_**/highlight.min.js
- that has sha256: 837a6fa5b0c736b52bbde2b2b6190f305da3fc9ed41681db5321507057b5c846

...while it was requested to download a file with the same name:
- URL: https://unpkg.com/@highlightjs/cdn-assets@**_11.9.0_**/highlight.min.js
- and sha256: f178f9e890cb9129fc25944d1ea1f1f4d556af0783ec5a867a930d5cdccabab4

mickaelistria:
OK. Could it be a consequence of https://github.com/maven-download-plugin/maven-download-plugin/issues/260 then?

vrubezhny:
> OK. Could it be a consequence of [maven-download-plugin/maven-download-plugin#260](https://github.com/maven-download-plugin/maven-download-plugin/issues/260) then?

It could be, but we're using v.1.7.1 (where the issue is supposed to be fixed)

mickaelistria:
OK, I see you're right that the cache seems to be per filename; I was mentioning how the earlier cache (the one I used to maintain) was working but didn't see it was changed to something else, which is unfortunately known as bugged.
The fix for download-maven-plugin was merged 3 days ago: https://github.com/maven-download-plugin/maven-download-plugin/commit/e96de15b405266b332fa2e45d9ed0c58a0eabfc4 ; I suggest we do like some other projects (as linked on the issue) and revert to 1.6.8 until a good 1.7.x comes out.

org.eclipse.lsp4e/pom.xml
```diff
@@ -9,9 +9,13 @@
...
+		<forceSkipCache>false</forceSkipCache>
+	</properties>
+
 	<build>
 		<plugins>
-			<plugin>
+			<!-- plugin>
 				<groupId>org.apache.maven.plugins</groupId>
 				<artifactId>maven-clean-plugin</artifactId>
 				<version>3.3.2</version>
```
mickaelistria, line 21: Just delete if you think it's not worth keeping it.

vrubezhny: Yeah. It's just a draft yet.

vrubezhny:
Closing as outdated.
Also, it looks like the problem we faced (not re-downloaded file with the same name and new sha256) is definitely fixed in download-maven-plugin v.1.8.1

PR #878, opened 2023-11-22
Make sure that dspParameters map is always initialized

akurtakov:
Broken by eclipse/lsp4e#756 .
Fixes https://github.com/redhat-developer/eclipseide-jdtls/issues/111


akurtakov:
@rubenporras  What do you think about this one? I would like to see it a version so we can contribute it to 2023-12

PR #881, opened 2023-11-28
let LSJavaCompletionProposal implement extension interface to enable linked editing mode

martinlippert:
Accepting a language server completion proposal that is defined as a snippet in the Java source editor does not trigger the linked editing mode.

The problem is that the implementation of the `LSJavaCompletionProposal` is somewhat limited and doesn't implement the `ICompletionProposalExtension2` yet, so that the completion proposal doesn't get all the necessary information for entering the linked editing mode.

This PR adds the necessary implementation of `ICompletionProposalExtension2` to the `LSJavaCompletionProposal` and the linked editing mode is triggered nicely when inside the Java source file editor.


mickaelistria:
@martinlippert I don't use the LSP4E/JDT bindings, and AFAIK, it's not part of some stack I'm supposed to support for my employer. So I don't think I'll even try to review here; I trust you're doing the best thing as you're more expert than me on this particular area.

Issue #865, opened 2023-11-01
Remove usage of Guave 


vogella:
IIRC I saw recently a discussion from @akurtakov that the usage of Guave in lsp4e would be desired. Opening this issue to capture this. 

@ptziegler does the same at the moment for WindowBuilder see for example https://github.com/eclipse-windowbuilder/windowbuilder/pull/615

vogella:
Is it actually used? https://github.com/search?q=repo%3Aeclipse%2Flsp4e+guava&type=code does not show any usage in Java files.

akurtakov:
Guava is not in the packagename. You should search for com.google.common .

vogella:
Still only 8 usages. Maybe @ptziegler can help here also after his experience with wb?

ptziegler:
From what I can tell, only the following part of Guava is used is lsp4e:

- Chars.contains(\<Char Array\>, \<Char\>)
- Functions.identity()
- Strings.nullToEmpty(\<String\>), Strings.isNullOrEmpty(\<String\>)
- ThreadFactoryBuilder
- Ranges & RangeMap

If necessary, the first four could probably be rewritten to not use Guava anymore, but except for Functions.identity() (which should be identical to java.util.function.Function.identity()), I don't think there is a drop-in replacement.

The usage of ranges pretty much requires Guava (or a library with similar functionality). Whether a migration is worth not carrying a 2MB dependency around, I can't answer.
The reason I work on this for WindowBuilder is primarily because:
- we still use e.g. the Guava Predicates, which have also been introduced with Java 8
- we use both Guava and Apache Commons to do basically the same thing (create & manage collections)

Both reason don't seem to apply here.

sebthom:
We are using guava in tm4e (CacheBuilder, GsonBuilder, Iterators, Strings, Lists) which is usually used together with lsp4e. I don't see us removing it from tm4e without introducing other new dependencies. So I don't really see the benefit of removing it from lsp4e.

laeubi:
> We are using guava in tm4e (CacheBuilder, GsonBuilder, Iterators, Strings, Lists) which is usually used together with lsp4e.

See
- https://github.com/eclipse/tm4e/issues/607

by the way m2e uses guava as well for the sake of `CacheBuilder` it might be good to extract that part into a separate artifact or have an Eclipse plain vanilla replacement for that.

laeubi:
Also see 
[...]

PR #886, opened 2023-12-23
reduce guava code usage

sebthom:
This PR addresses #865 and removes most usages of the guava library.

The only location left is at https://github.com/eclipse/lsp4e/blob/680b76572d29eecc0d3ce0047930d2cacf436424/org.eclipse.lsp4e/src/org/eclipse/lsp4e/outline/SymbolsLabelProvider.java#L185-L227  where com.google.common.collect.RangeMap is used. 

I cannot really wrap my head around that code introduced by @mickaelistria with https://github.com/eclipse/lsp4e/commit/ae5a762cf56bdaf7ce6596cf570a658079e8f79c. Maybe he is able to change the implementation to not require guava.


mickaelistria:
The change here doesn't make things simpler. So I wouldn't merge it yet if it's not enabling us to fully remove guava.
The `getMaxServerity` is a way to get the highest severity that affects a range: LSP diagnostics are attached to ranges, some can be overlapping, and the RangeMap allows to easily access the diagnostics for a range, including the included or overlapping ranges. I'm not aware of a better structure for that in standard JDK.

PR #888, opened 2023-12-29
Fix closing quick outline dialog after selection

travkin79:
The quick outline view (in my case used with the C/C++ editor based on LSP) behaved odd. It did not close after selecting an element symbol) in an open file. There was also an x button missing.

I fixed that.


rubenporras:
Thanks

PR #892, opened 2024-01-12
Prevent error logging for RequestCancelledExceptions

joaodinissf:
If the language server throws a `CancellationException` upon cancelling a request, this should not be logged as an error by the `LanguageServerPlugin`.

A check to ensure this behavior already exists in some locations (e.g., `org.eclipse.lsp4e.operations.completion.LSContentAssistProcessor.computeCompletionProposals(ITextViewer, int)`).

However, other locations where this check should be performed do not include it. This commit introduces the missing checks.


mickaelistria:
Thank you!

Issue #885, opened 2023-12-18
Popup "Problem occurred" when closing projects that use faces-config.xml or persistence.xml


erik-brangs:
In my Eclipse 2023-09 installation at work, when I close a project that has a ``faces-config.xml`` file or a ``persistence.xml``, I get a popup with the title ``Problem Occurred`` and a text like ``'Update markers from diagnostics' has encountered a problem. Resource '$projectName/src/main/resources/META-INF/persistence.xml' does not exist``. I'm mostly using multi-module projects.

I also have stack trace for a case with ``faces-config-xml``:

```
!ENTRY org.eclipse.core.resources 4 368 2023-10-12 13:17:20.356
!MESSAGE Resource '/[redacted]/src/main/webapp/WEB-INF/faces-config.xml' does not exist.
!STACK 0
java.lang.Exception: Resource '/[redacted]/src/main/webapp/WEB-INF/faces-config.xml' does not exist.
	at org.eclipse.core.internal.resources.ResourceException.provideStackTrace(ResourceException.java:42)
	at org.eclipse.core.internal.resources.ResourceException.<init>(ResourceException.java:38)
	at org.eclipse.core.internal.resources.Resource.checkExists(Resource.java:367)
	at org.eclipse.core.internal.resources.Resource.checkAccessible(Resource.java:241)
	at org.eclipse.core.internal.resources.Resource.findMarkers(Resource.java:1031)
	at org.eclipse.lsp4e.operations.diagnostics.LSPDiagnosticsToMarkers$1.runInWorkspace(LSPDiagnosticsToMarkers.java:140)
	at org.eclipse.core.internal.resources.InternalWorkspaceJob.run(InternalWorkspaceJob.java:43)
	at org.eclipse.core.internal.jobs.Worker.run(Worker.java:63)
```

I tried updating to Eclipse 2023-12 but that failed due to a dependency conflict that does not make sense to me.

erik-brangs:
The problem still occurs in Eclipse 2023-12.

mickaelistria:
@erik-brangs Can you please try submitting a PR to fix it? I suspect it's mostly a matter of adding a `resource.exists()` somewhere.

erik-brangs:
I haven't seen this issue since upgrading to the version that contained the PR.

PR #893, opened 2024-01-12
Add rough sketch for a bugfix for #885

erik-brangs:
As requested by @mickaelistria in #885 , here is a pull request that highlights where exists() checks might be missing.

The fix is probably wrong because I wasn't actually able to reproduce the problem. The associated test seems to trigger a different problem.

I suppose the original problem is triggered by the order of operations. I don't know enough to reproduce it in a test case.

### DIFF
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/diagnostics/LSPDiagnosticsToMarkers.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/diagnostics/LSPDiagnosticsToMarkers.java
@@ -140,6 +140,10 @@ public class LSPDiagnosticsToMarkers implements Consumer<PublishDiagnosticsParam
 
 			@Override
 			public IStatus runInWorkspace(IProgressMonitor monitor) throws CoreException {
+				if (!resource.exists()) {
+					return Status.OK_STATUS;
+				}
+
 				final var toDeleteMarkers = new HashSet<IMarker>(
 						Arrays.asList(resource.findMarkers(markerType, true, IResource.DEPTH_ZERO)));
 				toDeleteMarkers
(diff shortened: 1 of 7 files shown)


erik-brangs:
The test `testDiagnosticsForMarkerUpdateAfterProjectClose` now reproduces the original problem.

However, I don't know if this fully solves the problem with missing `exists()` checks. I've added some more `exists()` checks in other places where errors occurred when trying to write the tests. However, I can't reproduce those reliably. This also applies to the `testDiagnosticsForMarkerUpdateAfterDeletedFile` test.

The markers rule seems to be `null` in the tests. Is this normal?

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/diagnostics/LSPDiagnosticsToMarkers.java
```diff
@@ -166,12 +170,16 @@ public IStatus runInWorkspace(IProgressMonitor monitor) throws CoreException {
...
-						resource.createMarker(markerType, markerAttributes);
+						if (resource.exists()) {
+							Map<String, Object> markerAttributes = computeMarkerAttributes(document, diagnostic, resource);
+							resource.createMarker(markerType, markerAttributes);
+						}
 					}
 					for (Entry<IMarker, Diagnostic> entry : toUpdate.entrySet()) {
-						Map<String, Object> markerAttributes = computeMarkerAttributes(document, entry.getValue(), resource);
-						updateMarker(markerAttributes, entry.getKey());
+						if (resource.exists()) {
```
mickaelistria, line 179: couldn't it be `if (entry.getKey().exists())` here instead?

erik-brangs: Yes, you're right.

PR #901, opened 2024-01-25
Fix #900 Listen to ESC, enter, double-click, and single click

### LINKED ISSUE #900: Impossible to select an element in the C++ LSP Quick Outline view with a keyboard
I observed a very strange behavior when using the C++ LSP quick outline view. Selecting an element with the up and down keys doesn't work. Hitting one of the keys closes the view. It would ne nice if this view would behave similar to JDT's quick outline view for Java code.

**Details**
As soon as the selection in the quick outline view changes, e.g. when using the up and down keys in order to select the element to navigate to, the view is closed immediately. Users have no chance to select the desired element with the up and down keys.
(Clicking on an element with the mouse seems to work as expected.)

This issue is related to PR #888.

travkin79:
Listen to ESC, enter, double-click, and single click.
Improve selection listener concept.
Ensure, elements can be selected with keyboard or mouse. Extract method for revealing the selected element in editor.

### DIFF
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/symbols/LSPSymbolInFileDialog.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/symbols/LSPSymbolInFileDialog.java
@@ -11,8 +11,6 @@
  *******************************************************************************/
 package org.eclipse.lsp4e.operations.symbols;
 
-import java.util.concurrent.atomic.AtomicBoolean;
-
 import org.eclipse.core.resources.IFile;
 import org.eclipse.jdt.annotation.NonNull;
 import org.eclipse.jface.dialogs.PopupDialog;
@@ -34,9 +32,18 @@ import org.eclipse.lsp4j.SymbolInformation;
 import org.eclipse.lsp4j.jsonrpc.messages.Either;
 import org.eclipse.osgi.util.NLS;
 import org.eclipse.swt.SWT;
+import org.eclipse.swt.events.KeyEvent;
+import org.eclipse.swt.events.KeyListener;
+import org.eclipse.swt.events.MouseAdapter;
+import org.eclipse.swt.events.MouseEvent;
+import org.eclipse.swt.events.SelectionEvent;
+import org.eclipse.swt.events.SelectionListener;
+import org.eclipse.swt.graphics.Point;
 import org.eclipse.swt.widgets.Composite;
 import org.eclipse.swt.widgets.Control;
 import org.eclipse.swt.widgets.Shell;
+import org.eclipse.swt.widgets.Tree;
+import org.eclipse.swt.widgets.TreeItem;
 import org.eclipse.ui.dialogs.FilteredTree;
 import org.eclipse.ui.dialogs.PatternFilter;
 import org.eclipse.ui.internal.navigator.NavigatorContentService;
@@ -48,6 +55,7 @@ public class LSPSymbolInFileDialog extends PopupDialog {
 
 	private final OutlineViewerInput outlineViewerInput;
 	private final ITextEditor textEditor;
+	private TreeViewer treeViewer;
 
 	public LSPSymbolInFileDialog(@NonNull Shell parentShell, @NonNull ITextEditor textEditor,
 			@NonNull IDocument document, @NonNull LanguageServerWrapper wrapper) {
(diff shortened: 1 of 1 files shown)


org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/symbols/LSPSymbolInFileDialog.java
```diff
@@ -63,48 +73,98 @@ protected Control createDialogArea(Composite parent) {
...
+		treeViewer.setLabelProvider(new NavigatorDecoratingLabelProvider(contentService.createCommonLabelProvider()));
+		treeViewer.setUseHashlookup(true);
+
+		final Tree tree = treeViewer.getTree();
+
+		// listen to ESC key
+		tree.addKeyListener(new KeyListener() {
+			@Override
+			public void keyPressed(KeyEvent e)  {
+				if (e.character == 0x1B) { // ESC
```
rubenporras, line 97: Is it possible to use org.eclipse.swt.SWT.ESC?

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/symbols/LSPSymbolInFileDialog.java
```diff
@@ -63,48 +73,98 @@ protected Control createDialogArea(Composite parent) {
...
+				gotoSelectedElement();
+			}
+		});
+
+		// listen to left mouse button's single click
+		tree.addMouseListener(new MouseAdapter() {
+			@Override
+			public void mouseUp(MouseEvent e) {
+
+				if (tree.getSelectionCount() < 1)
```
rubenporras, line 124: could we combine both conditions in one if and use {} even if there is only one statement ?

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/symbols/LSPSymbolInFileDialog.java
```diff
@@ -63,48 +73,98 @@ protected Control createDialogArea(Composite parent) {
...
+					return;
+
+				if (e.button != 1)
+					return;
+
+				if (tree.equals(e.getSource())) {
+					Object o= tree.getItem(new Point(e.x, e.y));
```
rubenporras, line 131: since this object has not a meaningful name I suggest we remove the variable as in my suggestion below

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/symbols/LSPSymbolInFileDialog.java
```diff
@@ -63,48 +73,98 @@ protected Control createDialogArea(Composite parent) {
...
+					TreeItem selection= tree.getSelection()[0];
+					if (selection.equals(o))
```
rubenporras, line 133:
```suggestion
					if (selection.equals(tree.getItem(new Point(e.x, e.y))))
```

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/symbols/LSPSymbolInFileDialog.java
```diff
@@ -48,6 +57,7 @@ public class LSPSymbolInFileDialog extends PopupDialog {
 
 	private final OutlineViewerInput outlineViewerInput;
 	private final ITextEditor textEditor;
+	private TreeViewer treeViewer;
```
rubenporras, line 60: I suggest that to keep the diff smaller, we leave the name as `viewer`

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/symbols/LSPSymbolInFileDialog.java
```diff
@@ -113,12 +173,23 @@ protected Control createDialogArea(Composite parent) {
...
 			}
-		});
 
-		viewer.setInput(outlineViewerInput);
-		return filteredTree;
+			close();
+		}
+	}
+
+	protected Object getSelectedElement() {
```
rubenporras, line 182: Could it be private?

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/symbols/LSPSymbolInFileDialog.java
```diff
@@ -63,48 +73,98 @@ protected Control createDialogArea(Composite parent) {
...
+		});
+
+		treeViewer.setInput(outlineViewerInput);
+		return filteredTree;
+	}
+
+	private void gotoSelectedElement() {
+		Object selectedElement= getSelectedElement();
 
+		if (selectedElement instanceof Either<?, ?> either) {
```
rubenporras, line 154: I think it is better to move this to  `getSelectedElement()`

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/symbols/LSPSymbolInFileDialog.java
```diff
@@ -113,12 +173,23 @@ protected Control createDialogArea(Composite parent) {
...
+			close();
+		}
+	}
+
+	protected Object getSelectedElement() {
+		if (treeViewer == null) {
+			return null;
+		}
+
+		IStructuredSelection selection = treeViewer.getStructuredSelection();
```
rubenporras, line 187: I am not sure if treeViewer could be made asynchronously null (since it is now a class field), to be sure that we do not throw an NPE, I would assign if first to a local variable, then check the local variable, and if not null, continue using that local variable

rubenporras (changes requested):

travkin79:
It seems, there is again a time-out problem. Unfortunately, I cannot trigger a new build.

rubenporras (changes requested):

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/symbols/LSPSymbolInFileDialog.java
```diff
@@ -78,33 +88,77 @@ protected Control createDialogArea(Composite parent) {
...
+
+		viewer.setInput(outlineViewerInput);
+		return filteredTree;
+	}
+
+	private void gotoSelectedElement() {
+		Object selectedElement= getSelectedElement();
+
+		if (selectedElement != null) {
+			close();
```
rubenporras, line 151: should it not go after the catch on the block contaning `textEditor.selectAndReveal(offset, endOffset - offset);`?

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/symbols/LSPSymbolInFileDialog.java
```diff
@@ -78,33 +88,75 @@ protected Control createDialogArea(Composite parent) {
...
-			// selection is pre-selected in the tree view when the outline popup is opens
-			if (isFirstSelectionEvent.get()) {
-				isFirstSelectionEvent.set(false);
-				return;
+		final Tree tree = viewer.getTree();
+
+		tree.addKeyListener(new KeyListener() {
+			@Override
+			public void keyPressed(KeyEvent e)  {
+				if (e.character == org.eclipse.swt.SWT.ESC) {
```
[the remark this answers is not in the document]
travkin79: Oh yes, you're absolutely right. I was sloppy here.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/symbols/LSPSymbolInFileDialog.java
travkin79: Good question. My rationale was, if I want the editor to jump to a certain position and I have a user-confirmed selection in the quick outline view, i.e. the user did hit enter or clicked on the selected element (that are the only cases when `goToSelectedElement()` is called), then I want the quick outline view to be closed, even then if I detect some problem like a missing range. Otherwise, the view keeps being open and the user has no opportunity to close it (except of hitting ESC key).

Do you agree?

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/symbols/LSPSymbolInFileDialog.java
```diff
@@ -113,12 +165,31 @@ protected Control createDialogArea(Composite parent) {
 				} catch (BadLocationException e) {
 					LanguageServerPlugin.logError(e);
 				}
-				close();
 			}
-		});
 
-		viewer.setInput(outlineViewerInput);
-		return filteredTree;
+			close();
```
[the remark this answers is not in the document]
travkin79: See my next comment

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/symbols/LSPSymbolInFileDialog.java
travkin79: Do you mean, I should keep the `close()` operation as last statement in the block `if (range != null) {...}` and create PR for moving it where it is now, after `if (selectedElement != null) {`?

rubenporras: Yes, that would be great

travkin79: Ok, I change the close() call as you requested and created a PR #904 for moving the close operation as I suggested.

rubenporras (approved):

rubenporras:
thanks

PR #903, opened 2024-01-27
[#902] Do not split line endings in edits

### LINKED ISSUE #902: Blank editor after formatting source code via LS
When formatting a source code file in the editor, the editor can become blank caused by a `BadLocationException` during `TextViewer.setRedraw`. See this[ cdt-lsp issue ](https://github.com/eclipse-cdt/cdt-lsp/issues/239) for further details

After some investigation I found out that it will be triggered, when the `TextEdit` response on a `textDocument/formatting` request leads to a splitting of the `\r\n` line ending. See this [comment](https://github.com/eclipse-cdt/cdt-lsp/issues/239#issuecomment-1913091960).

This could be fixed in `org.eclipse.lsp4e.LSPEclipseUtils.applyEdits(IDocument, List<? extends TextEdit>)`. I can provide a PR for that.

ghentschke:
Splitting \r\n line endings leads to a BadLocationException while TextViewer.setRedraw is executed during a document.replace operation triggered by a TextEdit.

fixes #902


mickaelistria:
Do you think you can create a test case for this change? It seems pretty easy to mess this code up later, so a test would consolidate the fix.

mickaelistria:
Thank you!

PR #904, opened 2024-01-29
Fix closing quick outline view

travkin79:
There are a few cases, where updating the selection / cursor position in an open editor might fail. In those cases the solution from PR #901 might not work and the user could only close the quick outline view with the ESC key (the dialog has no x Button). Instead of first updating the editor's selection and then closing the quick outline view, I suggest to change the order, i.e. first close the quick outline view and then update the editor's selection.

### DIFF
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/symbols/LSPSymbolInFileDialog.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/symbols/LSPSymbolInFileDialog.java
@@ -148,6 +148,8 @@ public class LSPSymbolInFileDialog extends PopupDialog {
 		Object selectedElement= getSelectedElement();
 
 		if (selectedElement != null) {
+			close();
+
 			Range range = null;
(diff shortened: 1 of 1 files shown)


rubenporras:
can you rebase?

rubenporras:
@mickaelistria , any opinion on this PR?

mickaelistria:
I have not tried it but I'm afraid it would close the dialog to close is navigating it with the keyboard, ie hitting down until one finds the element they want, which is a common use case (at least to me).
I think it's OK to close the dialog in case of a double-click, but for other operations (simple click, keyboard navigation), the dialog should remain open.

travkin79:
Hi @mickaelistria,
I explicitly tested both cases, (a) navigating in the quick outline view to the desired element with the mouse (simple click on the desired element) and (b) using the keyboard's up and down buttons to select the element to navigate to. Both cases work with this version. You're welcome to try it yourself.

The dialog is only closed if the user selected a certain element and did hit enter or clicked on it (or hits the ESC key).

PR #906, opened 2024-01-30
Indicate when content assist proposals are incomplete.

andrewL-avlq:
Add a place-holder proposal that indicates when the list of content assist proposals is incomplete.

### DIFF
--- a/org.eclipse.lsp4e.test/src/org/eclipse/lsp4e/test/completion/AbstractCompletionTest.java
+++ b/org.eclipse.lsp4e.test/src/org/eclipse/lsp4e/test/completion/AbstractCompletionTest.java
@@ -11,7 +11,7 @@
  *******************************************************************************/
 package org.eclipse.lsp4e.test.completion;
 
-import static org.junit.Assert.*;
+import static org.junit.Assert.assertEquals;
 
 import java.util.ArrayList;
 import java.util.List;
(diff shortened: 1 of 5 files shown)


mickaelistria:
Can you please share a screenshot of how this looks like?

mickaelistria:
OK. What about have the item directly saying "Continue typing for more proposals..."? Also, can we find an interesting emoji to add here in order to showcase this is not a regular entry? Something like ➕ ?

andrewL-avlq:
No problem changing the item text. Do you want to leave the additional info as is?

mickaelistria:
> No problem changing the item text. Do you want to leave the additional info as is?

Yes, i think we can keep it, and improve it later.

andrewL-avlq:
Using existing icons from the Eclipse platform could look something like one of these:
2024_01_31_00_40_49_Greenshot

Or if the license allows, we could use something from https://www.iconfinder.com/search?q=more&price=free&license=gte__2 or something like https://www.iconfinder.com/icons/326714/pencil_translate_write_icon

Do any of these appeal (I quite like the pencil icon)? Or do you have a preferred site to get icons from?

mickaelistria:
I personally like more emojis than Eclipse icons nowadays because they're consistent with my OS, are zoomable, are screen-readable...  So I would just add an emoji to the label.
Using Eclipse icons makes sense too, although it doesn't have most of the benefits mentioned above.
Using external icons would require some IP effort and would bring more visual inconsistency with OS and Eclipse workbench, I think we should avoid that.

Overall, I think either emoji or Eclipse icons are fine. Your call ;)

PR #914, opened 2024-02-13
Fix NPE if ITextViewer is nullified during call to .underline()

joaodinissf:
Previously, it was possible for the `viewer` variable used in `LSPDocumentLinkPresentationReconcilingStrategy.underline()` to be nullified externally, which caused a null-pointer exception for the call `viewer.getTextWidget()`.

This commit solves this problem by saving the relevant viewer as a variable local to the
`LSPDocumentLinkPresentationReconcilingStrategy.underline()` function.


rubenporras (approved):
thanks

PR #919, opened 2024-02-14
Build against LSP4J 0.22.0

akurtakov:


rubenporras:
@mickaelistria , shall we have a last 2.24.x release before upgrading LSP4J? I would be for doing it

akurtakov:
Looking at https://github.com/eclipse/lsp4e/compare/0.24.7...master doesn't strike me as anything major in the last 2 weeks to deserve a release. Am I missing smth?

rubenporras:
You are right, there is nothing in the last two weeks which is a major. The reason I propose to do a minor release before the LSP4J upgrade, is that consumers can get the latest fixes without upgrading LSP4J.

akurtakov:
So if you want to do a release do one and please merge this one.  We will need a release right after that so we deliver latest lsp4j/lsp4e combo to simrel.

rubenporras:
Okay, I will do one shortly and will merge this one then.

rubenporras:
@akurtakov 
(org.eclipse.lsp4e/0.18.4.202402140838). Expected to have bigger x.y.z than what is available in baseline (0.18.4.202402140832)
could you bump the number?

PR #923, opened 2024-02-21
Support click on inlayhint label parts

BoykoAlex:
`InlayHint` may have a number of `InlayHintLabelPart`s. This is necessary support to determine which label part has been clicked on to execute the command corresponding to the clicked label part.

Includes a test for a single label part

### DIFF
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/inlayhint/LSPLineContentCodeMining.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/inlayhint/LSPLineContentCodeMining.java
@@ -24,6 +24,7 @@ import org.eclipse.jface.text.ITextViewer;
 import org.eclipse.jface.text.codemining.LineContentCodeMining;
 import org.eclipse.lsp4e.LSPEclipseUtils;
 import org.eclipse.lsp4e.LanguageServerWrapper;
+import org.eclipse.lsp4e.LanguageServers.LanguageServerDocumentExecutor;
 import org.eclipse.lsp4e.command.CommandExecutor;
 import org.eclipse.lsp4j.Command;
 import org.eclipse.lsp4j.ExecuteCommandOptions;
@@ -122,8 +123,13 @@ public class LSPLineContentCodeMining extends LineContentCodeMining {
 					if(command != null && command.getCommand() != null && !command.getCommand().isEmpty()) {
 						ExecuteCommandOptions provider = wrapper.getServerCapabilities().getExecuteCommandProvider();
 						if (provider != null && provider.getCommands().contains(command.getCommand())) {
-							wrapper.execute(ls -> ls.getWorkspaceService()
-									.executeCommand(new ExecuteCommandParams(command.getCommand(), command.getArguments())));
+							LanguageServerDocumentExecutor.forDocument(document).computeAll((w, ls) -> {
+								if (w == this.wrapper) {
+									return ls.getWorkspaceService()
+											.executeCommand(new ExecuteCommandParams(command.getCommand(), command.getArguments()));
+								}
+								return CompletableFuture.completedFuture(null);
+							});
 						} else  {
 							CommandExecutor.executeCommandClientSide(command, document);
 						}
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/inlayhint/LSPLineContentCodeMining.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/inlayhint/LSPLineContentCodeMining.java
@@ -26,7 +26,6 @@ import org.eclipse.lsp4e.LSPEclipseUtils;
 import org.eclipse.lsp4e.LanguageServerWrapper;
 import org.eclipse.lsp4e.LanguageServers.LanguageServerDocumentExecutor;
 import org.eclipse.lsp4e.command.CommandExecutor;
-import org.eclipse.lsp4j.Command;
 import org.eclipse.lsp4j.ExecuteCommandOptions;
 import org.eclipse.lsp4j.ExecuteCommandParams;
 import org.eclipse.lsp4j.InlayHint;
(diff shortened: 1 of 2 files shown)


BoykoAlex:
What do you think of this PR @mickaelistria @rubenporras ? I wasn't able to write a proper unit test for the multi part label... However, testing it manually with multi part label from LS each part with its own command - worked very well for me.
I've tried to find GEF/GMF DirectEditRequest logic that would find from the click where to place the cursor in the popped up text cell editor (i.e. find char under mouse) but looks like this code is gone... Nevertheless, what I've implemented seem to do the job... GC would only be created if more than one label part present and multi-label clicking seem to perform acceptable.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/inlayhint/LSPLineContentCodeMining.java
```diff
@@ -91,4 +112,76 @@ private static org.eclipse.jface.text.Position toPosition(Position position, IDo
...
+	public final Consumer<MouseEvent> getAction() {
+		return me -> {
+			String title= getLabel();
+			if (title != null && !title.isEmpty()) {
+				findLabelPart(me).ifPresent(labelPart -> {
+					final Command command = labelPart.getCommand();
+					if(command != null && command.getCommand() != null && !command.getCommand().isEmpty()) {
+						ExecuteCommandOptions provider = wrapper.getServerCapabilities().getExecuteCommandProvider();
+						if (provider != null && provider.getCommands().contains(command.getCommand())) {
+							wrapper.execute(ls -> ls.getWorkspaceService()
```
rubenporras, line 125: should we not use LanguageServerDocumentExecutor.LanguageServerDocumentExecutor(IDocument) and execute the command through it so that we ensure the order of the executions for the given document?

BoykoAlex: Shall I change it to:

```java
	public final Consumer<MouseEvent> getAction() {
		return me -> {
			String title= getLabel();
			if (title != null && !title.isEmpty()) {
				findLabelPart(me).ifPresent(labelPart -> {
					final Command command = labelPart.getCommand();
					if(command != null && command.getCommand() != null && !command.getCommand().isEmpty()) {
						ExecuteCommandOptions provider = wrapper.getServerCapabilities().getExecuteCommandProvider();
						if (provider != null && provider.getCommands().contains(command.getCommand())) {
							LanguageServerDocumentExecutor.forDocument(document).computeAll((wrapper, ls) -> {
								if (wrapper == this.wrapper) {
									return ls.getWorkspaceService()
											.executeCommand(new ExecuteCommandParams(command.getCommand(), command.getArguments()));
								}
							});
						} else  {
							CommandExecutor.executeCommandClientSide(command, document);
						}
					}
				});
			}
		};
	}

```

BoykoAlex:
Any update on this PR? Shall I indeed use `LanguageServerDocumentExecutor.LanguageServerDocumentExecutor(IDocument)` to execute a command from inlay label part?

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/inlayhint/LSPLineContentCodeMining.java
rubenporras: Sorry for having missed your comment. Yes, I think something like that is good. Maybe use just `(w, ls)` instead of `(wrapper, ls)`, as I think that is the commonly used abbreviation in the codebase.

mickaelistria: IMO, explicit names that one can read out of a context are always better than abbreviation which require deep knowledge of the overall code. I think we should avoid mandating such "shortening" of names in general and most often stick with the contributor's style. I find project that welcome and handle more diverse style are more comfortable to work with than the ones that force a style (Maven is my worst experience in OSS contribution because of that)

rubenporras: Yes, I agree that they are usually better. Only this context, `wrapper` already exists as a class field, so I would tend to give a different name if possible. In any case, it was only a suggestion, not a requirement :)

mickaelistria: >  Only this context, wrapper already exists as a class field,

Ah OK, I had missed that point.

rubenporras:
@BoykoAlex , if you rebase, it should build again.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/inlayhint/LSPLineContentCodeMining.java
```diff
@@ -91,4 +113,81 @@ private static org.eclipse.jface.text.Position toPosition(Position position, IDo
 		return new org.eclipse.jface.text.Position(start, 1);
 	}
 
+	@Override
+	public final Consumer<MouseEvent> getAction() {
+		return me -> {
+			String title= getLabel();
+			if (title != null && !title.isEmpty()) {
+				findLabelPart(me).ifPresent(labelPart -> {
```
rubenporras, line 121: would that be a bit more readable by reducing the number of nested ifs?
```suggestion
				findLabelPart(me).map(InlayHintLabelPart::getCommand).map(Command::getCommand).ifPresent(command -> {
```

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/inlayhint/LSPLineContentCodeMining.java
```diff
@@ -91,4 +113,81 @@ private static org.eclipse.jface.text.Position toPosition(Position position, IDo
...
+				Image image = null;
+				GC gc = null;
+				Font font = null;
+				try {
+					image = new Image(display, 1, 1);
+					gc = new GC(image);
+					font = new Font(display, fontData);
+					gc.setFont(font);
+					Point origin = new Point(0, 0);
+					for (InlayHintLabelPart l : inlayHint.getLabel().getRight()) {
```
rubenporras, line 159:
```suggestion
					for (InlayHintLabelPart labelPart : inlayHint.getLabel().getRight()) {
```

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/inlayhint/LSPLineContentCodeMining.java
```diff
@@ -91,4 +113,81 @@ private static org.eclipse.jface.text.Position toPosition(Position position, IDo
...
+						}
+					}
+				});
+			}
+		};
+	}
+
+	private Optional<InlayHintLabelPart> findLabelPart(MouseEvent me) {
+		if (inlayHint.getLabel().isRight()) {
+			if (inlayHint.getLabel().getRight().size() == 1) {
```
rubenporras, line 144: Could we extract inlayHint.getLabel().getRight() to a local variable and replace the three accesses in this method?
```suggestion
			if (inlayHint.getLabel().getRight().size() == 1) {
```

rubenporras (approved):
I only have minor comments. Would you like to merge as it is or would you like to modify the PR a little bit?

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/inlayhint/LSPLineContentCodeMining.java
BoykoAlex: Sounds good, did this.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/inlayhint/LSPLineContentCodeMining.java
BoykoAlex: Sounds good, did this.

PR #924, opened 2024-02-22
Support refresh inlayhints

BoykoAlex:


org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageClientImpl.java
```diff
@@ -166,24 +166,29 @@ public CompletableFuture<ShowDocumentResult> showDocument(ShowDocumentParams par
...
+	private void updateCodeMinings() {
+		IWorkbenchPage activePage = UI.getActivePage();
+		if (activePage == null) {
+			return;
+		}
+		IEditorReference[] editors = activePage.getEditorReferences();
+		for (IEditorReference ref : editors) {
+			var editor = ref.getEditor(false);
+			var codeMiningSourceViewer = Adapters.adapt(editor, ITextViewer.class, true);
+			if (codeMiningSourceViewer != null && codeMiningSourceViewer instanceof SourceViewer) {
```
rubenporras, line 178:
```suggestion
			if (textViewer != null && textViewer instanceof SourceViewer sourceViewer) {
```

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageClientImpl.java
```diff
@@ -166,24 +166,29 @@ public CompletableFuture<ShowDocumentResult> showDocument(ShowDocumentParams par
...
+				((SourceViewer) codeMiningSourceViewer).updateCodeMinings();
```
rubenporras, line 179:
```suggestion
				sourceViewer.updateCodeMinings();
```

rubenporras (approved):
I think it would be nicer if we rename the variable and avoid one cast using one of the new features that Java 17 offers, but the change is also fine as it is.
Shall we merge or do you want to apply the suggestions first?

rubenporras (approved):
thanks

PR #933, opened 2024-02-28
[#932] prevent reconcile on every dirty region

### LINKED ISSUE #932: Performance issue when formatting a large file
Given is a large file (~5000 lines). When performing a formatting with many formatted lines (e.g. by changing the indent with) via LSP, the folding and semantic highlighting reconciler are triggered for every queued change event in the document. Since both reconcilers perform on the whole document they should be performed only once, when the document has been modfied and not for each quewed document change.

A formatting with many formatted lines result in many queued change events:
```txt
I[09:48:20.585] <-- textDocument/didChange
I[09:48:20.585] <-- textDocument/didChange
I[09:48:20.585] <-- textDocument/didChange
I[09:48:20.585] <-- textDocument/didChange
I[09:48:20.586] <-- textDocument/didChange
I[09:48:20.586] <-- textDocument/didChange
...
```
which results in many (thousands!) consecutive formatting and semantic highlighting reconciler runs:
```
I[09:48:21.774] <-- textDocument/semanticTokens/full("483")
I[09:48:21.779] --> reply:textDocument/semanticTokens/full("483") 4 ms
I[09:48:21.835] <-- textDocument/foldingRange("484")
I[09:48:21.917] --> reply:textDocument/foldingRange("484") 81 ms
I[09:48:22.340] <-- textDocument/semanticTokens/full("485")
...
```

ghentschke:
This change prevents reconcile run on every dirty region which can be many thousands in a large source file (> 5000 lines) Since the reconciler operate on the whole document it needed to be executed only when the file document changes.

fixes #932

### DIFF
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/semanticTokens/SemanticHighlightReconcilerStrategy.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/semanticTokens/SemanticHighlightReconcilerStrategy.java
@@ -256,12 +256,10 @@ public class SemanticHighlightReconcilerStrategy
 	}
 
 	private void fullReconcile() {
-		IDocument theDocument = document;
-		var ts = DocumentUtil.getDocumentModificationStamp(theDocument);
-		if (disabled || !isInstalled || ts == timestamp) { // Skip any processing
+		if (disabled || !isInstalled) { // Skip any processing
 			return;
 		}
-		timestamp = ts;
+		IDocument theDocument = document;
 		cancelSemanticTokensFull();
 		if (theDocument != null) {
 			long modificationStamp = DocumentUtil.getDocumentModificationStamp(theDocument);
@@ -295,12 +293,24 @@ public class SemanticHighlightReconcilerStrategy
 
 	@Override
 	public void reconcile(final DirtyRegion dirtyRegion, final IRegion subRegion) {
-		fullReconcile();
+		fullReconcileOnce();
 	}
 
 	@Override
 	public void reconcile(final IRegion partition) {
-		fullReconcile();
+		fullReconcileOnce();
+	}
+
+	/**
+	 * Because a full reconcile will be performed always on the whole document,
+	 * prevent consecutive reconciling on dirty regions or partitions if the document has not changed.
+	 */
+	private void fullReconcileOnce() {
+		var ts = DocumentUtil.getDocumentModificationStamp(document);
+		if (ts != timestamp) {
+			fullReconcile();
+			timestamp = ts;
+		}
 	}
 
 	@Override
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/semanticTokens/SemanticHighlightReconcilerStrategy.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/semanticTokens/SemanticHighlightReconcilerStrategy.java
@@ -105,6 +105,8 @@ public class SemanticHighlightReconcilerStrategy
 	 */
 	private volatile long documentTimestampAtLastAppliedTextPresentation;
 
+	private volatile long timestamp = 0;
+
 	private CompletableFuture<Optional<VersionedSemanticTokens>> semanticTokensFullFuture;
 
 	public SemanticHighlightReconcilerStrategy() {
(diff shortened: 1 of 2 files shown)


rubenporras:
I am not sure we not covering for some other fault with this PR. 

For example: 
- the folding reconciler: it has a subregion as input parameter. Should we not reconcile only the changed regions? If that takes a bit too long, why is that a problem? Should reconcile not happen in a background thread?
- the semantic reconciler: even thought we do not implement it yet, we could also reconcile only changed regions.

Besides, will this PR not stop reconcilers from doing the job until the user saves the file? If so, I think that would be a regression.

ghentschke:
> * the folding reconciler: it has a subregion as input parameter. Should we not reconcile only the changed regions? If that takes a bit too long, why is that a problem? Should reconcile not happen in a background thread?
> * the semantic reconciler: even thought we do not implement it yet, we could also reconcile only changed regions.
> 

The problem/difficulty is, at least for clangd, a format with many modified lines (e.g. by changing the indent width) via LSP could result in a event queue with many thousands dirty regions. I've noticed that it takes very long (several minutes) until this queue has been processed in background threads (these thousands of changed regions could be separate issue as well). 
IMO performing the folding and the semantic highlighter on the whole document (as it is currently) has in many cases a better performance than processing each dirty region. 
But I'll check that.

> Besides, will this PR not stop reconcilers from doing the job until the user saves the file? If so, I think that would be a regression.

IMO not, because `DocumentUtil.getDocumentModificationStamp` returns a new timestamp when the document gets edited.

rubenporras:
> DocumentUtil.getDocumentModificationStamp returns a new timestamp when the document gets edited.

DocumentUtil.getDocumentModificationStamp returns a new timestamp when the document gets edited but not when formats are applied using the formatter?

Is this desired behaviour or some corner case that has been forgoten?

ghentschke:
> DocumentUtil.getDocumentModificationStamp returns a new timestamp when the document gets edited but not when formats are applied using the formatter

No, it returns a new timestamp when the document has been modified. This includes also formatting changes. That was the problem here. 
The method:
https://github.com/eclipse/lsp4e/blob/8df7e08de54f325a40835480e6fdf9a1cd281cfe/org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/folding/LSPFoldingReconcilingStrategy.java#L314
gets called on every dirty region. This can be several thousands in a large file with many formatting changes. I made this PR to prevent the repetitive call of the folding reconcile which operates on the whole document.

rubenporras:
Okay, I see. Is the problem that the method `void reconcile(IRegion subRegion)` is not using the `subRegion` because the LSP has no option to only calculate the folding ranges for a subregion?

ghentschke:
> Okay, I see. Is the problem that the method void reconcile(IRegion subRegion) is not using the subRegion because the LSP has no option to only calculate the folding ranges for a subregion?

Yes

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/semanticTokens/SemanticHighlightReconcilerStrategy.java
```diff
@@ -254,10 +256,12 @@ private void cancelSemanticTokensFull() {
...
 
 	private void fullReconcile() {
-		if (disabled || !isInstalled) { // Skip any processing
+		IDocument theDocument = document;
+		var ts = DocumentUtil.getDocumentModificationStamp(theDocument);
+		if (disabled || !isInstalled || ts == timestamp) { // Skip any processing
 			return;
 		}
-		IDocument theDocument = document;
+		timestamp = ts;
```
rubenporras, line 264: Now that I understand your PR better, should we not have the check these two methods?

```
	@Override
	public void reconcile(final DirtyRegion dirtyRegion, final IRegion subRegion) {
		fullReconcile();
	}

	@Override
	public void reconcile(final IRegion partition) {
		fullReconcile();
	}
```

In the end it will end the reconciliation no matter where we add the check and the exit, but I think it makes more clear what it the purpose of the check and exit

ghentschke: Okay, do mean something like this?:
```java
	@Override
	public void reconcile(final DirtyRegion dirtyRegion, final IRegion subRegion) {
                var ts = DocumentUtil.getDocumentModificationStamp(document);
               if (ts != timestamp) {
		  fullReconcile();
                  timestamp = ts;
                }
	}
```

rubenporras (approved):

ghentschke:
@rubenporras  Thank you!

PR #935, opened 2024-02-28
[#934] fix removed folding annotations on document change

### LINKED ISSUE #934: Folding annotations gets removed on large files while editing
When opening a large file with in a LSP backed generic text editor (e.g. C file with clangd) and starting to edit this file, the folding annotations will  be removed:

image

ghentschke:
The method
ProjectionAnnotationModel.modifyAnnotations(Annotation[], Map<? extends Annotation, ? extends Position>, Annotation[]) does not consider Position changes in any cases. On large annotations the positions won't get updated properly which leads to removed folding annotations.

fixes #934

### DIFF
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/folding/LSPFoldingReconcilingStrategy.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/folding/LSPFoldingReconcilingStrategy.java
@@ -139,6 +139,7 @@ public class LSPFoldingReconcilingStrategy
 	private void applyFolding(List<FoldingRange> ranges) {
 		// these are what are passed off to the annotation model to
 		// actually create and maintain the annotations
+		final var modifications = new ArrayList<Annotation>(); // not used anymore
 		final var deletions = new ArrayList<FoldingAnnotation>();
 		final var existing = new ArrayList<FoldingAnnotation>();
 		final var additions = new HashMap<Annotation, Position>();
@@ -150,7 +151,7 @@ public class LSPFoldingReconcilingStrategy
 			if (ranges != null) {
 				Collections.sort(ranges, Comparator.comparing(FoldingRange::getEndLine));
 				for (FoldingRange foldingRange : ranges) {
-					updateAnnotation(deletions, existing, additions, foldingRange.getStartLine(),
+					updateAnnotation(modifications, deletions, existing, additions, foldingRange.getStartLine(),
 							foldingRange.getEndLine(), FoldingRangeKind.Imports.equals(foldingRange.getKind()));
 				}
 			}
@@ -228,7 +229,8 @@ public class LSPFoldingReconcilingStrategy
 	 *            the end line number
 	 * @throws BadLocationException
 	 */
-	private void updateAnnotation(List<FoldingAnnotation> deletions, List<FoldingAnnotation> existing, Map<Annotation, Position> additions, int line, Integer endLineNumber, boolean collapsedByDefault)
+	private void updateAnnotation(List<Annotation> modifications, List<FoldingAnnotation> deletions,
+			List<FoldingAnnotation> existing, Map<Annotation, Position> additions, int line, Integer endLineNumber, boolean collapsedByDefault)
 			throws BadLocationException {
 		int startOffset = document.getLineOffset(line);
 		int endOffset = document.getLineOffset(endLineNumber) + document.getLineLength(endLineNumber);
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/folding/LSPFoldingReconcilingStrategy.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/folding/LSPFoldingReconcilingStrategy.java
@@ -238,7 +238,7 @@ public class LSPFoldingReconcilingStrategy
 		final var newPos = new Position(startOffset, endOffset - startOffset);
 		if (!existing.isEmpty()) {
 			FoldingAnnotation existingAnnotation = existing.remove(existing.size() - 1);
-			updateAnnotations(existingAnnotation, newPos, deletions);
+			updateAnnotations(existingAnnotation, newPos, modifications, deletions);
 		} else {
 			additions.put(new FoldingAnnotation(collapsedByDefault), newPos);
 		}
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/folding/LSPFoldingReconcilingStrategy.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/folding/LSPFoldingReconcilingStrategy.java
@@ -167,7 +167,8 @@ public class LSPFoldingReconcilingStrategy
 			}
 			// send the calculated updates to the annotations to the
 			// annotation model
-			theProjectionAnnotationModel.modifyAnnotations(deletions.toArray(new Annotation[1]), additions, new Annotation[0]);
+			theProjectionAnnotationModel.modifyAnnotations(deletions.toArray(new Annotation[1]), additions,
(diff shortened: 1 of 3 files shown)


org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/folding/LSPFoldingReconcilingStrategy.java
```diff
@@ -269,9 +269,7 @@ protected void updateAnnotations(Annotation existingAnnotation, Position newPos,
 				Position oldPos = theProjectionAnnotationModel.getPosition(foldingAnnotation);
 				// only update the position if we have to
 				if (!newPos.equals(oldPos)) {
-					oldPos.setOffset(newPos.offset);
-					oldPos.setLength(newPos.length);
-					modifications.add(foldingAnnotation);
```
ghentschke, line 274: modifications can be removed formt eh arguments list of this method.  But this would be an API break.

rubenporras:
Are we fixing a problem with ProjectionAnnotationModel.modifyAnnotations? Or how is it that "On large annotations the positions won't get updated properly which leads to removed folding annotations"?

ghentschke:
> Are we fixing a problem with ProjectionAnnotationModel.modifyAnnotations?

Not really. IMO we havent't used the proper method. And `theProjectionAnnotationModel.modifyAnnotationPosition(foldingAnnotation, newPos)` 
does exactly what we want: update the position of a Annotation.

rubenporras:
> > Are we fixing a problem with ProjectionAnnotationModel.modifyAnnotations?
> 
> Not really. IMO we haven't used the proper method. And `theProjectionAnnotationModel.modifyAnnotationPosition(foldingAnnotation, newPos)` does exactly what we want: update the position of a Annotation.

I guess that you mean that call `projectionAnnotationModel.modifyAnnotations(deletions.toArray(new Annotation[1]), additions, modifications.toArray(new Annotation[0]))` is not the proper method. Why not?

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/folding/LSPFoldingReconcilingStrategy.java
rubenporras: I do not think there will be many subclasses of LSPFoldingReconcilingStrategy, so I would prefer that we modify the signature accordingly (or probably just remove the method, as it does not update anything any longer.)

@mickaelistria , what do you think?

akurtakov: FYI Mickael is not available this week.

rubenporras: Since Mickael is not available this week, could you mark the method as deprecated for removal and document on the javadoc that the parameter is unused? Then we can cleanup on a later iteration.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/folding/LSPFoldingReconcilingStrategy.java
```diff
@@ -151,7 +150,7 @@ private void applyFolding(List<FoldingRange> ranges) {
 			if (ranges != null) {
 				Collections.sort(ranges, Comparator.comparing(FoldingRange::getEndLine));
 				for (FoldingRange foldingRange : ranges) {
-					updateAnnotation(modifications, deletions, existing, additions, foldingRange.getStartLine(),
+					updateAnnotation(deletions, existing, additions, foldingRange.getStartLine(),
```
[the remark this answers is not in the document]
ghentschke: > we still need to call it, otherwise we would be breaking the customized code

Hm, I am not sure I am understand you here.  The modifications is always empty, because its not written anymore in `LSPFoldingReconcilingStrategy.updateAnnotations(existingAnnotation, newPos, deletions)`. The `updateAnnotation` which is called here in line 153 is a private method.

Theres no need to call the deprecated  `LSPFoldingReconcilingStrategy.updateAnnotations(existingAnnotation, newPos, modifications, deletions)` method here.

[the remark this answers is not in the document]
rubenporras: > Hm, I am not sure I am understand you here. The modifications is always empty, because its not written anymore in `LSPFoldingReconcilingStrategy.updateAnnotations(existingAnnotation, newPos, deletions)`

Yes, the method as we have rewritten it will return empty annotations, but if some subclass subclassed the method, we still need to call it, otherwise we would break the customization.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/folding/LSPFoldingReconcilingStrategy.java
```diff
@@ -167,8 +167,7 @@ private void applyFolding(List<FoldingRange> ranges) {
 			}
 			// send the calculated updates to the annotations to the
 			// annotation model
-			theProjectionAnnotationModel.modifyAnnotations(deletions.toArray(new Annotation[1]), additions,
-					modifications.toArray(new Annotation[0]));
+			theProjectionAnnotationModel.modifyAnnotations(deletions.toArray(new Annotation[1]), additions, new Annotation[0]);
```
rubenporras, line 170: We still need to pass `modifications`  to `theProjectionAnnotationModel.modifyAnnotations`, as a subclass can have added some there, should we not?

ghentschke: applyFolding is a private method. Subclasses can't have added something here. Or not?

ghentschke: Ah, I think I got your point here now :-)

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/folding/LSPFoldingReconcilingStrategy.java
```diff
@@ -238,7 +238,7 @@ private void updateAnnotation(List<Annotation> modifications, List<FoldingAnnota
 		final var newPos = new Position(startOffset, endOffset - startOffset);
 		if (!existing.isEmpty()) {
 			FoldingAnnotation existingAnnotation = existing.remove(existing.size() - 1);
-			updateAnnotations(existingAnnotation, newPos, modifications, deletions);
+			updateAnnotations(existingAnnotation, newPos, deletions);
```
rubenporras, line 241:
```suggestion
			updateAnnotations(existingAnnotation, newPos, modifications, deletions);
```

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/folding/LSPFoldingReconcilingStrategy.java
```diff
@@ -139,7 +139,7 @@ public void reconcile(IRegion subRegion) {
 	private void applyFolding(List<FoldingRange> ranges) {
 		// these are what are passed off to the annotation model to
 		// actually create and maintain the annotations
-		final var modifications = new ArrayList<Annotation>();
+		final var modifications = new ArrayList<Annotation>(); // not used anymore
```
rubenporras, line 142:
```suggestion
		final var modifications = new ArrayList<Annotation>(); // not used anymore, can be removed later with the deprecated updateAnnotations method
```

rubenporras (approved):
Thanks a lot!

ghentschke:
@rubenporras thank you for quick review!

PR #937, opened 2024-02-29
Releng improvements

akurtakov:


pom.xml
```diff
@@ -129,23 +128,6 @@
 					</execution>
 				</executions>
 			</plugin>
-			<plugin>
-				<groupId>org.jacoco</groupId>
```
rubenporras, line 133: we are removing jacoco because our jenkins job does not publish its results, so it is useless. Am I correct?

akurtakov: Yes. And even if publishing I don't think anyone is/will look at test coverage results.

akurtakov: I try to put detailed commit messages with this info as can be seen at https://github.com/eclipse/lsp4e/pull/937/commits/56d6c0ee3c74d80bb25048f2f707591116bd2484

rubenporras (approved):

pom.xml
akurtakov: As you have approved the changes I'm resolving this one.

PR #940, opened 2024-03-01
[#930] fix broken semantic highlight unit test

### LINKED ISSUE #930: Master build broken due to failed unit test
The CI build job fails due to a filed unit test: 
```
array lengths differed, expected.length=3 actual.length=0; arrays first differed at element [0]; expected:<StyleRange {0, 4, fontStyle=normal, foreground=Color {133, 153, 0, 255}}> but was:<end of array>
```
The failed test is 
https://github.com/eclipse/lsp4e/blob/78fffa519b11b5c56bcc612e5133a4cef8adfb95/org.eclipse.lsp4e.test/src/org/eclipse/lsp4e/test/semanticTokens/SemanticHighlightReconcilerStrategyTest.java#L51

ghentschke:
due to the
https://github.com/eclipse/tm4e/commit/5c780dc312c764f59c1a3d52202e6f135840030c commit in tm4e the unit test fails.

fixes #930

### DIFF
--- a/org.eclipse.lsp4e.test/src/org/eclipse/lsp4e/test/semanticTokens/SemanticHighlightReconcilerStrategyTest.java
+++ b/org.eclipse.lsp4e.test/src/org/eclipse/lsp4e/test/semanticTokens/SemanticHighlightReconcilerStrategyTest.java
@@ -61,11 +61,11 @@ public class SemanticHighlightReconcilerStrategyTest {
 		DisplayHelper.sleep(display, 2_000); // Give some time to the editor to update
 
 		StyleRange[] styleRanges = textViewer.getTextWidget().getStyleRanges();
-
+		//default theme is dark, that's why we have to expect a blue foreground color:
 		List<StyleRange> expectedStyleRanges = Arrays.asList(//
-				new StyleRange(0, 4, SemanticTokensTestUtil.GREEN, null), //
-				new StyleRange(15, 4, SemanticTokensTestUtil.GREEN, null), //
-				new StyleRange(24, 7, SemanticTokensTestUtil.GREEN, null)//
+				new StyleRange(0, 4, SemanticTokensTestUtil.BLUE, null), //
+				new StyleRange(15, 4, SemanticTokensTestUtil.BLUE, null), //
+				new StyleRange(24, 7, SemanticTokensTestUtil.BLUE, null)//
 		);
 		assertArrayEquals(expectedStyleRanges.toArray(), styleRanges);
 	}
--- a/org.eclipse.lsp4e.test/src/org/eclipse/lsp4e/test/semanticTokens/SemanticHighlightReconcilerStrategyTest.java
+++ b/org.eclipse.lsp4e.test/src/org/eclipse/lsp4e/test/semanticTokens/SemanticHighlightReconcilerStrategyTest.java
@@ -8,7 +8,7 @@
  *******************************************************************************/
 package org.eclipse.lsp4e.test.semanticTokens;
 
-import static org.junit.Assert.assertArrayEquals;
+import static org.junit.Assert.assertEquals;
 
 import java.util.Arrays;
 import java.util.List;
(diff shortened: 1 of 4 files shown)


org.eclipse.lsp4e.test/src/org/eclipse/lsp4e/test/semanticTokens/SemanticHighlightReconcilerStrategyTest.java
```diff
@@ -61,11 +61,11 @@ public void testKeyword() throws CoreException {
 		DisplayHelper.sleep(display, 2_000); // Give some time to the editor to update
 
 		StyleRange[] styleRanges = textViewer.getTextWidget().getStyleRanges();
-
+		//default theme is dark, that's why we have to expect a blue foreground color:
```
mickaelistria, line 64: I don't think this is generally true. The default theme may depend on other OS settings. We need something more flexible (eg evaluating whether theme is dark or not), or more constraining (eg setting the theme) here as we cannot reliably assume what the default theme is.

ghentschke: Agree. Wouldn't it be enough to check the start and length of the style range?

mickaelistria: Yes, checking ranges + `color != widget.getBackgroundColor()` would be good enough

org.eclipse.lsp4e.test/src/org/eclipse/lsp4e/test/semanticTokens/SemanticHighlightReconcilerStrategyTest.java
```diff
@@ -61,12 +62,20 @@ public void testKeyword() throws CoreException {
...
-				new StyleRange(0, 4, SemanticTokensTestUtil.GREEN, null), //
-				new StyleRange(15, 4, SemanticTokensTestUtil.GREEN, null), //
-				new StyleRange(24, 7, SemanticTokensTestUtil.GREEN, null)//
-		);
-		assertArrayEquals(expectedStyleRanges.toArray(), styleRanges);
+		assertEquals(3, styleRanges.length);
+		
+		assertEquals(0, styleRanges[0].start);
+		assertEquals(4, styleRanges[0].length);
+		assertTrue(styleRanges[0].foreground != backgroundColor);
```
mickaelistria, line 71: Just nitpicking, can you please use `assertNotEquals`  instead?

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/semanticTokens/TokenTypeMapper.java
```diff
@@ -42,6 +44,9 @@ public IToken apply(final String tokenType) {
 				return tokenProvider.getToken(tokenType);
 			}
 		}
-		return TMUIPlugin.getThemeManager().getDefaultTheme().getToken(tokenType);
+		final ITheme[] defaultTheme = new ITheme[1];
+		// Needs to be executed in UI-thread:
```
mickaelistria, line 48: Wasn't it fixed in TM4E directly?

ghentschke: Yes, but I've seen that LSP4E is using the TM4E release for building.

mickaelistria: I think it would be OK to move LSP4E to use latest snapshot of TM4E until its next release. What do you think?

ghentschke: I agree.

rubenporras: I was planning to do a release once https://github.com/eclipse/lsp4e/pull/935 is merged for simrel. If you agree, I think we should wait for that before using TM4E snapshot.

ghentschke: Sorry, I've seen your comment tool late. It's fine for me. I'll provide a new commit...

PR #941, opened 2024-03-04
[#930] fix flaky unit test

ghentschke:
part of #930


org.eclipse.lsp4e.test/src/org/eclipse/lsp4e/test/references/FindReferencesTest.java
```diff
@@ -74,11 +78,14 @@ public void log(UiFreezeEvent event) {
 
 	@Before
 	public void setUp() throws CoreException {
+		shell = new Shell();
```
mickaelistria, line 81: Can't we use `textViewer.getTextWidget().getDisplay()` directly here instead of instantiating a new shell?

ghentschke: I think so

PR #942, opened 2024-03-04
Make formatting operations available to consumers

ava-fred:
This commit adds org.eclipse.lsp4e.operations.format to the list of exported packages to enable consumers to implement formatting for objects which do not readily adapt to an IFile. There is also a small refactor of the LSPFormatFilesHandler class with the same goal.

### DIFF
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/format/LSPFormatFilesHandler.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/format/LSPFormatFilesHandler.java
@@ -98,11 +98,7 @@ public class LSPFormatFilesHandler extends AbstractHandler {
 					}
 				});
 				docProvider.changed(doc);
-				try {
-					docProvider.saveDocument(monitor, file, doc, true);
-				} catch (CoreException e) {
-					LanguageServerPlugin.logError(e);
-				}
+				saveDocument(docProvider, file, monitor);
 			});
 		} catch (InterruptedException ex) {
 			Thread.currentThread().interrupt();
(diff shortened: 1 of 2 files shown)


org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/format/LSPFormatFilesHandler.java
```diff
@@ -98,11 +98,7 @@ protected void formatFile(final @NonNull IFile file, final IProgressMonitor moni
 					}
 				});
 				docProvider.changed(doc);
-				try {
-					docProvider.saveDocument(monitor, file, doc, true);
-				} catch (CoreException e) {
-					LanguageServerPlugin.logError(e);
-				}
+                saveDocument(docProvider, file, monitor);
```
rubenporras, line 101: it looks like the commit uses spaces for the changed lines instead of tabs, which are used in the rest of the file. Could you change that?

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/format/LSPFormatFilesHandler.java
```diff
@@ -118,6 +114,14 @@ protected IDocumentProvider getDocumentProvider(IFile file) {
 		return DocumentProviderRegistry.getDefault().getDocumentProvider(new FileEditorInput(file));
 	}
 
+    protected void saveDocument(IDocumentProvider docProvider, IFile file, IProgressMonitor monitor) {
```
rubenporras, line 117: same comment regarding space/tabs on these lines

rubenporras (approved):
thanks

rubenporras:
@mickaelistria , any class that you want to make private before exporting the package?

mickaelistria:
> any class that you want to make private before exporting the package?

No, I've become very laxist on this topic ;)

PR #945, opened 2024-03-09
add missing array outline icon

sebthom:
The LSPImages class references a non-existing array.png:
https://github.com/eclipse/lsp4e/blob/838b4c98cde21b4241aea1b9d67b78724fbf652e/org.eclipse.lsp4e/src/org/eclipse/lsp4e/ui/LSPImages.java#L101

This results in a red dot being displayed for array/lists in the outline:

image


mickaelistria:
Thank you!

PR #947, opened 2024-03-15
Create job to monitor the initialization of the language server

joaodinissf:
This commit creates the method `createInitializeLanguageServerJob()` in the LanguageServerWrapper, which returns a Job that can be used to monitor the initialization process of a language server.


org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServerWrapper.java
```diff
@@ -713,13 +738,7 @@ protected CompletableFuture<LanguageServer> getInitializedServer() {
...
 		if (initializeFuture != null && !this.initializeFuture.isDone()) {
 			if (Display.getCurrent() != null) { // UI Thread
-				final var waitForInitialization = new Job(Messages.initializeLanguageServer_job) {
-					@Override
-					protected IStatus run(IProgressMonitor monitor) {
-						initializeFuture.join();
-						return Status.OK_STATUS;
-					}
-				};
+				final Job waitForInitialization = createInitializeLanguageServerJob();
```
rubenporras, line 741: @mickaelistria , the PR makes sense to me as it is, but I wonder if we need a blocking dialog if we introduce a normal progress bar as part of this PR. To me blocking dialogs is something we should avoid, and I do not know about any feature that would require it.

Could we remove this block as part of the PR? I think it was introduced by you in https://github.com/eclipse/lsp4e/commit/9c23de59f4d9d656051b930e56b008b6a4da951e#diff-a27e5243bbfe3244b1bd63bf3fee0cbac1747e63479eee932391b08b4bb62257, so maybe you know what was the use case behind.

Regards

joaodinissf: @mickaelistria Have you had a chance to go over this PR? Thanks in advance.

mickaelistria: > I wonder if we need a blocking dialog if we introduce a normal progress bar as part of this PR. To me blocking dialogs is something we should avoid, and I do not know about any feature that would require it.

I don't think the blocking dialog is really intentional. If you want some other dialog to report progress, that's most likely fine to implement.
But that seems like a topic for a separate dedicated PR.

mickaelistria (approved):
The change looks good to me.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServerWrapper.java
```diff
@@ -713,13 +756,7 @@ protected CompletableFuture<LanguageServer> getInitializedServer() {
...
```
rubenporras, line 759: @joaodinissf , is initializeFutureNumberOfStages = 7 correct also for this case?

rubenporras: I think it can happen that the number of stages is not accurate for the UI pop-up, but since I think we should remove the pop-up in a new PR, we can merge this one nevertheless.

rubenporras (approved):

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServerWrapper.java
joaodinissf: I'll create a separate PR for the removal of the pop-up.

PR #950, opened 2024-03-20
Fix build failure by removing dependency on org.eclipse.osgi.services

joaodinissf:


rubenporras:
thanks.

rubenporras:
@mickaelistria , this fixes the current build failures, I do not see a problem with the PR, but if you see one, please tell me :)

PR #953, opened 2024-03-22
Refactor Preview for WorkspaceEdit with `needsConfirmation` flag

BoykoAlex:
`WorkspaceEdit` has a map of change annotations where any change annotation is associated with edits from the workspace edit. If any of change annotations has `needsConfirmation` flag on then VSCode would bring up the refactor preview with annotated changes unchecked.

This PR proposes similar (but not the same, there are limitations) behaviour into Eclipse.
If there is at lease one change annotation that requires confirmation then refactor wizard is opened with the Preview page opened. All changes would be checked marked as I couldn't find how to control this piece. Another limitation is Eclipse text edits are in reverse order (they are applied in reverse order for a reason) but the preview doesn't allow to set the tree sorter on the preview tree.

@rubenporras @mickaelistria @martinlippert feedback, ideas are welcomed :-)

### DIFF
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LSPEclipseUtils.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LSPEclipseUtils.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2016, 2022 Red Hat Inc. and others.
+ * Copyright (c) 2016, 2024 Red Hat Inc. and others.
  * This program and the accompanying materials are made
  * available under the terms of the Eclipse Public License 2.0
  * which is available at https://www.eclipse.org/legal/epl-2.0/
@@ -70,9 +70,12 @@ import org.eclipse.core.resources.ResourcesPlugin;
 import org.eclipse.core.runtime.Adapters;
 import org.eclipse.core.runtime.CoreException;
 import org.eclipse.core.runtime.IPath;
+import org.eclipse.core.runtime.IProgressMonitor;
 import org.eclipse.core.runtime.NullProgressMonitor;
+import org.eclipse.core.runtime.OperationCanceledException;
 import org.eclipse.core.runtime.Path;
 import org.eclipse.core.runtime.Platform;
+import org.eclipse.core.runtime.Status;
 import org.eclipse.core.runtime.content.IContentType;
 import org.eclipse.core.runtime.content.IContentTypeManager;
 import org.eclipse.jdt.annotation.NonNull;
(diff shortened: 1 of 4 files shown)


org.eclipse.lsp4e/src/org/eclipse/lsp4e/LSPEclipseUtils.java
```diff
@@ -925,25 +933,79 @@ public static void applyWorkspaceEdit(WorkspaceEdit wsEdit, String label) {
...
-					Position start = range.getStart() != null ? range.getStart() : new Position(0, 0);
-					if (Display.getCurrent() != null) {
-						open(uri.toString(), new Range(start, start));
-					} else {
-						UI.getDisplay().asyncExec(() -> open(uri.toString(), new Range(start, start)));
-					}
-				});
+
+		if (wsEdit.getChangeAnnotations() != null && wsEdit.getChangeAnnotations().values().stream().anyMatch(ca -> ca.getNeedsConfirmation())) {
+			Refactoring refactoring = new Refactoring() {
```
rubenporras, line 938: Could we move lines 938-973 into `runRefactorWizardOperation` passing the `change` to that method instead of the `wizard` as now?

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LSPEclipseUtils.java
```diff
@@ -925,25 +933,79 @@ public static void applyWorkspaceEdit(WorkspaceEdit wsEdit, String label) {
...
+					RefactoringWizard.NO_BACK_BUTTON_ON_STATUS_DIALOG
+			) {
+
+				@Override
+				protected void addUserInputPages() {
+					//no inputs required
+				}
+
+			};
+			if (Display.getCurrent() == null) {
```
rubenporras, line 974: for consistency within this method, could we invert the condition on `Display.getCurrent()` in the new or the old code so that they use the same condition and the the if/else is in the same order?

rubenporras (commented):
Hi Alex, 
the PR makes sense to me. I looks good to me but I do not know the Refactoring API at all.
My only comments are just some minor improvements on the code structure. If @martinlippert or @mickaelistria don't have further comments I would be happy if we merge it

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LSPEclipseUtils.java
```diff
@@ -925,25 +933,79 @@ public static void applyWorkspaceEdit(WorkspaceEdit wsEdit, String label) {
...
+				UI.getDisplay().asyncExec(() -> runRefactorWizardOperation(wizard, label));
+			} else {
+				runRefactorWizardOperation(wizard, label);
 			}
```
mickaelistria, line 978: I think it would be better to put the UI thread check directly in `runRefactorWizardOperation`.

mickaelistria (commented):
I had a quick look and it seems good overally,
I'm disappointed that there isn't a utility for that already in Platform, something like `new ChangeRefactoringWizard(change)`, nor nothing alike (as far as I search it). So in absence of such a more straightforward API, this implementation seems to use the correct bits.
I support the overall comments about factorizing more stuff into the `runRefactorWizardOperation`.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LSPEclipseUtils.java
BoykoAlex: Yes, definitely!

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LSPEclipseUtils.java
BoykoAlex: I'll create runOnUiThread(Runnable) in the UI class and use it here in both places

BoykoAlex:
@rubenporras @mickaelistria Thank you both very much for the prompt review :-)
I have added a commit to address the review comments with a few more changes:
1. Even if the workspace change is for one file that is opened in the editor in the case of `needsConfirmation` flag set to true we would still bring up the refactor preview. (I think this is rather obvious change)
2. Added `runOnUIThread(Runnable)` to the `UI` class add used it in `LSPEclipseUtils`
3. `ServerMessageHandler` use current active shell as the parent for `MessageDialog` instead of creating a new one - don't think you'd ever need to create a parent shell in Eclipse for a dialog. Creation of a new shell leaves its bounds undetermined the consequence of which is dialog box moves to bottom-right every time a new one is opened.

mickaelistria (approved):

rubenporras (approved):

rubenporras:
thank you

PR #954, opened 2024-03-28
Use Rename and MoveRename ResourceChange classes if possible

BoykoAlex:
If the change is rename and then modify the content of a file. This translates into
1. Delete the file
2. Create the new named file 
3. Set its new content 

This doesn't let to see the diffs in the refactor preview between the old file name file and the new name file side by side. If the Rename or MoveRename ResourceChange is used then refactor preview has more useful content.

This PR suggests to use Rename and MoveRename if this possible, i.e. both resources are in the workspace.


mickaelistria:
Thank you!

PR #968, opened 2024-04-16
Prevent NPE by asynchronous threads after the plug-in has been stop

rubenporras:


rubenporras:
@mickaelistria , I am wondering if instead of not logging anything after the plugin has been stop, we should rather use `this` as a fallback. I can imagine we are still able to log even if the plug-in has been stop, but probably that is not desired.

rubenporras:
Failed tests are flaky tests. The job passed already twice.

PR #970, opened 2024-04-17
Fix #969 Add collecting of declarations in addition to definitions

### LINKED ISSUE #969: Open Declaration (Ctrl + Click / F3) does not collect declarations
I checked an issue in a CDT LSP editor concerning the "open declaration" feature. In a C++ code example, I was able to jumpt to a method's definition, but not to its declaration.

I checked the LSP4E implementation in `org.eclipse.lsp4e.operations.declaration.OpenDeclarationHyperlinkDetector` and it turns out, the [LSP endpoint textDocument/declaration](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_declaration) is not used at all (see source code excerpt). I think, in addition to the definitions (and type definitions and implementations), the declarations should be collected here, too.

image

Are there any objections?

travkin79:
Instead of collecting only definitions (and type definitions and implementations), I modified the code to also collect declarations for the command "open declaration" in LSP4E. For example, in C++, that makes a difference, since methods can be defined, e.g. in .cpp files and declared in .hpp files.

### DIFF
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LSPEclipseUtils.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LSPEclipseUtils.java
@@ -100,6 +100,7 @@ import org.eclipse.lsp4j.CompletionContext;
 import org.eclipse.lsp4j.CompletionParams;
 import org.eclipse.lsp4j.CompletionTriggerKind;
 import org.eclipse.lsp4j.CreateFile;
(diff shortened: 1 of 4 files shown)


rubenporras (approved):
Thanks

travkin79:
It seems, there is an `OutOfMemoryError` in the build run.

rubenporras:
I retriggered the job and it has ended normally. 
Thanks for the PR!

PR #971, opened 2024-04-18
Add a preference foldingReconcilingStrategy.collapseImports to control

rubenporras:
if the imports are collapsed by the editor.


mickaelistria:
Just to make sure: did you really face an issue with imports being collapsed by default, or is it more that fact that they get collapsed late after editors open that is an issue? Because this doesn't fix the later.

PR #975, opened 2024-04-24
Fixed #973 The default label for multiple declarations is not useful.

### LINKED ISSUE #973: The default label for multiple declarations is not useful
The computed label for multiple declarations is the name and the path of the resource with is not really useful as an end user.

It would be better to display the line declaring the symbol so the used can select the relevant entry. This could be done at least for `IFile` from the workspace since we can have the encoding and the stream to read its content.

ylussaud:


mickaelistria:

PR #985, opened 2024-05-03
Fix race condition in LanguageServerWrapper resulting in NullPointerException during LS initialization

sebthom:
When running the test suite on Windows and in Docker with Ubuntu, **a lot** of NPEs with the following stacktrace are thrown:
```py
java.lang.NullPointerException: Cannot invoke
"org.eclipse.core.runtime.IProgressMonitor.done()" because
"this.this$0.initializeFutureMonitor" is null
	at org.eclipse.lsp4e.LanguageServerWrapper$2.run(LanguageServerWrapper.java:407)
	at org.eclipse.core.internal.jobs.Worker.run(Worker.java:63)
```
and because of the NPE random tests seem to fail.

### DIFF
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServerWrapper.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServerWrapper.java
@@ -19,6 +19,7 @@ package org.eclipse.lsp4e;
 
 import java.io.File;
 import java.io.IOException;
+import java.io.UncheckedIOException;
 import java.net.URI;
 import java.util.ArrayList;
 import java.util.HashMap;
@@ -297,7 +298,7 @@ public class LanguageServerWrapper {
 				try {
 					lspStreamProvider.start();
 				} catch (IOException e) {
-					throw new RuntimeException(e);
+					throw new UncheckedIOException(e);
 				}
 				return null;
 			}).thenRun(() -> {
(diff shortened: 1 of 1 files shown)


mickaelistria:
The tests failures are all related?

sebthom:
No, we are getting OOMs for all builds on Jenkins. No idea why. I opened a ticket but they say it must be a config change in our build process.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServerWrapper.java
```diff
@@ -387,11 +389,12 @@ private void advanceinitializeFutureMonitor() {
 		}
 	}
 
-	private synchronized Job createInitializeLanguageServerJob() {
+	private Job createInitializeLanguageServerJob() {
 		return new Job(NLS.bind(Messages.initializeLanguageServer_job, serverDefinition.label)) {
 			@Override
 			protected IStatus run(IProgressMonitor monitor) {
-				initializeFutureMonitor = SubMonitor.convert(monitor, initializeFutureNumberOfStages);
```
rubenporras, line 394: Maybe it is a silly question, but do we need an `AtomicReference`? Could we not achieve the same by saving `initializeFutureMonitor` to a local variable in line 394 (and also on line 382) before saving it to the class field, and then mark the monitor as done using the local variable?

Is there a benefit of `AtomicReference` over the local variable?

sebthom: As far as I see it, at a minimum the initializeFutureMonitor needs to be made volatile because it is accessed from two different threads and the Job in createInitializeLanguageServerJob otherwise may never see the assigned value. The nice part of AtomicReference is that it offers compareAndSet() which can avoid another type of race conditions when the wrapper is stopped and restarted in quick succession.

rubenporras (approved):
thanks

PR #1003, opened 2024-05-17
Rename head branch to main and update documentation links

sebthom:
This PR changes all references from "master" to "main". Besides the merge of this PR renaming the branch to `main` and setting it as default branch is required.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -104,7 +104,7 @@ LSP4E only accepts contributions via GitHub Pull Requests against [https://githu
 
 Before sending us a pull request, please ensure that:
 
-1. You are working against the latest source on the **master** branch.
+1. You are working against the latest source on the **main** branch.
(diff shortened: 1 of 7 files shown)


mickaelistria:
I'm OK with that. @rubenporras what do you think?

rubenporras:
I am also OK with the change. @sebthom, would you do the rename of the branch, etc. yourself after merging?

sebthom:
@rubenporras I don't think I have permission to change the default branch.

rubenporras:
I also do not know if I have the rights. @mickaelistria , do you know if you have them?

mickaelistria:
No, I don't. but we can proceed anyway, and just hope GitHub makes the right choice. If it doesn't, then we'll have to open an helpdesk ticket. The EF staff will encourage use to have our own `eclipse-lsp4e` GitHub orgs which would give committers and leads more power.

PR #1006, opened 2024-05-21
Code cleanup

sebthom:


org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/references/URIMatch.java
```diff
@@ -9,22 +9,28 @@
...
 import org.eclipse.jface.text.BadLocationException;
+import org.eclipse.jface.text.IDocument;
 import org.eclipse.lsp4e.LSPEclipseUtils;
 import org.eclipse.lsp4j.Location;
 import org.eclipse.search.ui.text.Match;
 
 public class URIMatch extends Match {
 
-	public final @NonNull Location location;
+	public static URIMatch create(final Location location) throws BadLocationException, URISyntaxException {
```
mickaelistria, line 22: What's wrong with sticking to a constructor here?

sebthom: The problem is that the first invocation in a constructor needs to be a super() call, that's why the current implementation parses the URI three times. to pass several values to the parent constructor args. which is very inefficient.

mickaelistria: OK, thanks. Can't wait for https://openjdk.org/jeps/447 ;)

sebthom: > OK, thanks. Can't wait for [openjdk.org/jeps/447](https://openjdk.org/jeps/447) ;)

Me too!

Issue #1007, opened 2024-05-21
Setup PublishDiagnosticsCapabilities


zulus:
Latest typescript server require this to send diagnostics notification

mickaelistria:
Fixed with #1008 , thanks!

PR #1008, opened 2024-05-21
Register PublishDiagnosticsCapabilities

zulus:
Simple fix for #1007

### DIFF
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/internal/SupportedFeatures.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/internal/SupportedFeatures.java
@@ -41,6 +41,7 @@ import org.eclipse.lsp4j.HoverCapabilities;
 import org.eclipse.lsp4j.InlayHintCapabilities;
 import org.eclipse.lsp4j.InsertTextMode;
 import org.eclipse.lsp4j.MarkupKind;
+import org.eclipse.lsp4j.PublishDiagnosticsCapabilities;
 import org.eclipse.lsp4j.RangeFormattingCapabilities;
 import org.eclipse.lsp4j.ReferencesCapabilities;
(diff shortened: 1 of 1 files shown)


rubenporras:
thanks

akurtakov:
As this is prereq for Typescript update in WildWebDeveloper is there a chance for a quick release for this one?

rubenporras:
Yes, no problem

akurtakov:
Let me push the lsp4j and guava updates first.

rubenporras:
Thanks for pointing that out. I had plan to wait for https://github.com/eclipse/lsp4e/pull/1010. That addresses LSP4J and Guava, or is any other PR comming?

akurtakov:
#1010 addresses LSP4J. The Guava one is trickier as we use it from Orbit and if we update it that would lead to other deps (bouncycastle, icu4j, objectweb-asm) from 2024-06 to be used or we can point to the maven artifact directly (like done here). Please have a say which one of the 3 options below you prefere:
* stay with old guava
* use new guava with other deps from 2024-06 Orbit
* use new guava from maven directly

rubenporras:
To keep the target file simple, I would keep using the orbit repository, if that is possible. Maybe it is not such a big deal to just update the orbit repo?

rubenporras:
The other option is to remove Guava, it does not look like we use for much, and we had a PR in the past where we tried to do the removal.

akurtakov:
Here you are https://github.com/eclipse/lsp4e/pull/1011

zulus:
in general would be nice to check if all capabilities that are already supported by lsp4e are correctly registered in SupportedFeatures.java

mickaelistria:
FWIW, this capability is new and wasn't required to get `publishDiagnostics` working, the behavior of the typescript language server isn't really backward compatible here, although the spec is (and that's why it has been the only LS failing).
But anyway, I agree with you that review capabilities often could be useful.

PR #1012, opened 2024-05-22
Sync tycho-build with the version in main pom.

akurtakov:


akurtakov:
@rubenporras With this one in I have nothing else and would appreciate a release.

PR #1025, opened 2024-06-04
Avoid assertion failure

BoykoAlex:
Looks like checking for `null` edit is a proper thing to do as `TextChange` from LTK has an assert `null` for the previous edit. Assertion exception appears in e4.32 but looks like the PR above fixes it.

### DIFF
--- a/org.eclipse.lsp4e/META-INF/MANIFEST.MF
+++ b/org.eclipse.lsp4e/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: Language Server Protocol client for Eclipse IDE (Incubation)
 Bundle-SymbolicName: org.eclipse.lsp4e;singleton:=true
-Bundle-Version: 0.18.10.qualifier
+Bundle-Version: 0.18.11.qualifier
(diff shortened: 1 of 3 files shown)


rubenporras:
Under which conditions is the `getEdit()` not null? 

Is that expected or do we have some programming error? If so, should we log a warning when it is not null with the necessary information to track down the problem?

rubenporras:
> e4.32



> Looks like checking for `null` edit is a proper thing to do as `TextChange` from LTK has an assert `null` for the previous edit. Assertion exception appears in e4.32 but looks like the PR above fixes it.

Is this so? I could not find the git repo, but looking at the plugins in my target, the check is already part of eclipse e.4.31.

BoykoAlex:
Okay, I have verified that the same occurs in 4.31 too.

The changes are grouped by resources. Under resources are leafs - the line changes.

Screenshot 2024-06-05 at 11 22 24

Clicking on different nodes and leafs in the changes tree in the wizard results in acquired doc count going up from 0 and back to 0 when doc is released. The doc release doesn't change the edit - it is set already. However, the code attempts to set it again. Thus, I'd just guard against setting the edit if it is set already.

rubenporras:
Thanks

PR #1026, opened 2024-06-06
Do not return a Consumer<MouseEvent> if the code mining has no action.

rubenporras:
According to ICodeMining.getAction(), null should be returned if there is no action to be executed when the mining is clicked.

When several InlayHintLabelParts are used, we cannot be sure if there is an action or not without knowing the actual MouseEvent. However, if no InlayHint has an action, we can be sure there is none even without knowing the MouseEvent. For these cases, the code returns null now.


rubenporras:
@BoykoAlex : I will merge it now. Please, let me know later if you have time if you think there are improvements that should be done.

Issue #1028, opened 2024-06-11
Deadlock on UI thread from Semantic Tokens markup with recent TM4e


ahmedneilhussain:
Ironically I think this is caused by the fix for https://github.com/eclipse/tm4e/issues/721 (which was at the instigation of LSP4e). I am using eclipse 2023-09 but with tm4e 11 installed (I wanted some of the bug fixes in recent versions that haven't yet made it into a version of TM4e bundled with an eclipse release).

Please see the attached thread dump: you can see the main thread (also the UI thread) thinks the document is dirty and is waiting for stuff to be posted onto the queue at https://github.com/eclipse-platform/eclipse.platform.ui/blob/934ccc66e79042dccd94093376e403d9a932dda9/bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/reconciler/AbstractReconciler.java#L123

Meanwhile the semantic token reconciler is trying to get the current theme:
https://github.com/eclipse/lsp4e/blob/11b81cfa0477fa9f8b135d0c9b3f8b45beb88640/org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/semanticTokens/TokenTypeMapper.java#L45

and in tm4e after the fix for issue 721 above, this now will try and use the UI thread to find out if the current theme is dark.

Personally, I think it a bit dodgy for AbstractReconciler to have a method that blocks its caller, if its caller can be the UI thread, but if that's a thing that can happen, then I don't think reconciler threads should be calling `syncExec()`: it looks as though the reconciler mechanism is supposed to be purely headless background computation that just posts its results onto a queue for the UI to pick up, but I haven't investigated in detail.

This occurred every time in one of our `SWTBot`-driven integration tests once I switched over to using TM4e as well as LSP4e, but I think it might be very timing-sensitive. It was fine interactively, worked fine in CI on MacOS, but hung every time when I ran just that test locally on my development machine (also a Mac). Our CI is hanging on windows, which I haven't confirmed is the same issue, but I'm hoping that that's the case! Our test teardown logic calls 

```
    PlatformUI.getWorkbench()
        .getDisplay()
        .syncExec(
            () ->
                PlatformUI.getWorkbench()
                    .getActiveWorkbenchWindow()
                    .getActivePage()
                    .closeAllEditors(false));
```
[...]

PR #1029, opened 2024-06-17
Semantic Tokens: Do not use default theme

ahmedneilhussain:
Fix for https://github.com/eclipse/lsp4e/issues/1028 - we requested the default theme if no TM mapping for the editor, but this calls syncExec. Reconciler threads mustn't block on the UI thread as it is possible for the UI thread to be blocked waiting for reconciliation; this deadlock was indeed observed.


ahmedneilhussain:
Hi @rubenporras without the fallback a test failed. I added a custom content type with a real TM grammar mapping and tweaked the expected tokens to match. Is that OK? Felt a bit heavyweight, but I couldn't see an easy way of mocking/injecting a TM mapping in the test setup. I thought it best to create a new content type mapping so that this change was segregated from the other tests.

rubenporras:
hi @ahmedneilhussain ,

some tests are still failing. Are these some flaky tests or does the removal of the fallback break those tests as well?

Regards

ahmedneilhussain:
> hi @ahmedneilhussain ,
> 
> some tests are still failing. Are these some flaky tests or does the removal of the fallback break those tests as well?
> 
> Regards

HI according to this https://ci.eclipse.org/lsp4e/job/lsp4e-github/job/PR-1029/2/testReport/ all the tests are passing. Am I missing something? Some of the CI runs are complaining because of the eclipse version bump check - I'm not sure if I'm supposed to do something about that or whether it will get bumped as part of prep for next release.

There are definitely some flaky tests but I think the fallback should not interfere with the majority of tests: the mock language server won't have anything in place for semantic tokens other than where we are explicitly testing that functionality, and we don't configure tm4e other than where I just introduced it (for a new content type used only for this one case). I could be wrong, but the test run above and the fact they ran clean (but flaky!) locally suggests not...

Cheers,
Ahmed

rubenporras:
Sorry, I have not noticed that. You need to manually bump the version of org.eclipse.lsp4e.test, if I recall right in the manifest.mf and in the pom.xml.

rubenporras (approved):
Thanks

PR #1034, opened 2024-07-03
lsp4e: Patch so we have access to the rename operations

ahmedneilhussain:
I'm in the process of getting us off a custom (shaded) version of lsp4e. We use some of the rename classes for a custom rename command initiated via a code lens, but this won't compile with production lsp4e as the package isn't exposed. Any objection to exposing it in the same way as some of the other operations (exposed but marked as private so you get a warning)?


rubenporras:
Not from my side. Any objections, @mickaelistria ?

rubenporras:
@mickaelistria , since there are no objections I merge the PR. If you have objections, we can take a look at them nevertheless.

PR #1044, opened 2024-07-24
fix: propagate LanguagServerWrapper startup exception correctly

ava-fred:
If there is an exception in the startup of the LanguageServerWrapper, consumers get a cancellation exception rather than the actual exception that occurred. This commit fixes the problem.


org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServerWrapper.java
```diff
@@ -532,7 +536,9 @@ public synchronized void stop() {
 		}
 
 		if (this.initializeFuture != null) {
-			this.initializeFuture.cancel(true);
+			if (doCancel) {
+				this.initializeFuture.cancel(true);
```
ghentschke, line 540: If the intializeFuture won't be canceled if an exception occurs e.g. during start of the LS binary, caused by an invalid path to the LS binary, the user gets annyed by many dialog boxes:
image
They pop up by every call on `org.eclipse.lsp4e.LanguageServerWrapper.getInitializedServer()` and the wrapper wants to start the server again, because `initializeFuture` is null and if fails again and and a new pop-up will be displayed. The `.getInitializedServer()` method will be called by every server call the editor does, and there are many....

ghentschke (changes requested):

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServerWrapper.java
ava-fred: I think I have found a solution, please re-review.

ava-fred: I discussed this with @rubenporras , and while we agree that this change makes sense, we realise that it is a significant change in behavior. I will open a discussion for this to collect opinions from the community.

rubenporras (approved):
@ghentschke , this makes sense now to me. We have opened a discussion in https://github.com/eclipse/lsp4e/discussions/1093 just in case you think it is not.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServerWrapper.java
```diff
@@ -486,6 +486,13 @@ public synchronized boolean isActive() {
 		return this.launcherFuture != null && !this.launcherFuture.isDone() && !this.launcherFuture.isCancelled();
 	}
 
+	/**
+	 * @return whether the last startup attempt has failed
+	 */
+	public synchronized boolean startupFailed() {
```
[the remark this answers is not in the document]
ava-fred: It is to determine whether a startup attempt on this LanguageServerWrapper object failed and a call to restart() is needed. We will be using it in our project.

ghentschke (approved):

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServerWrapper.java
ghentschke: I am fine with it now

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServerWrapper.java
```diff
@@ -241,8 +241,8 @@ private List<WorkspaceFolder> getRelevantWorkspaceFolders() {
 	}
 
 	/**
-	 * Starts a language server and triggers initialization. If language server is
-	 * started and active, does nothing. If language server is inactive, restart it.
+	 * Starts a language server and triggers initialization. If language server has been started
+	 * before, does nothing.
 	 */
 	public synchronized void start() {
 		start(false);
```
ghentschke, line 248: If an exception has been occurred during start. The LS cannot be started again when the cause has been fixed (e.g. wrong path do LS binary).

ghentschke (changes requested):

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServerWrapper.java
rubenporras: I think it could we worth explaining as well that "an Eclipse-base product can check `startupFailed` and call restart if the problem has been programmatically handled", and that otherwise "once the problem is fixed a restart of eclipse is needed". That would clarify as well the need for the `startupFailed` method for future readers of this class. My wording can most probably be improved :)

ghentschke: ..but I am okay with it. This can/should be handled by the LS provider

ghentschke (approved):

rubenporras:
Thanks for the improved javadoc

rubenporras:
It looks like there is a need for investigation of view the CI Jenkins job fails more or less consistently. Other PRs seem to be fine in the same job, so it does not look like a infrastructure glitch.

rubenporras (approved):
It looks like it was temporary glitch on the infrastructure.

Issue #693, opened 2023-06-12
Open Outline View with one level expanded


gerdleon:
Currently the outline view is collapsed when opening a source. See below:
![image](https://github.com/eclipse/lsp4e/assets/4208332/451f74b2-75ad-4e6e-bf94-161232590b13)

By default it should expand the first level as below:
![image](https://github.com/eclipse/lsp4e/assets/4208332/69dc9e53-e9d0-4203-9129-8067a745e51d)

PR #1048, opened 2024-08-12
Expand root outline view in outline view by default

joaodinissf:
Before this commit, the outline view is drawn with all elements collapsed by default. This often results in plenty of wasted space in the UI and low information density.

Instead, as discussed in #693, it would be preferable to draw the outline view with its first level expanded by default (i.e., displaying the root node and its immediate children by default).

Issue: #693


rubenporras (approved):

rubenporras:
Thanks

Issue #254, opened 2022-10-06
Filtering in the Outline View


gerdleon:
Similar to other Outline Views it would be useful to be able to filter the elements shown in the outline. For a  minimal viable feature it is likely the SymbolKind that is used for the filtering.

For a user it would be nice to be able to do such filtering but I am unsure about it in the context of LSP since it could differ based on the content-type. Meaning in one content-type hiding fields will be a reasonable filter but for another it would filter out nearly everything.

Do you have an opinion? Are the others seeing a need for such filtering?

mickaelistria:
Can you please give an example of 1 actual user-story where filtering would be helpful (eg in Wild Web Developer, or Corrosion, or ShellWax...)?

gerdleon:
I do not know enough about Corrosion, etc.

I think it is natural to offer this - which is also why VS Code has this implemented in their Outline/Breadcrumbs (per content-type https://www.youtube.com/watch?v=eyJS7rgRSYw).
Not all elements returned from the LS make always sense to be shown
A similar functionality is implemented in Eclipse JDT. There it is possible to filter the elements shown.

mickaelistria:
> Not all elements returned from the LS make always sense to be shown

Then it seems to be something to discuss with the Language Server.

gerdleon:
I looked at a few tickets opened for WildWebDeveloper. There is no such concrete ticket but I see one usability ticket that goes in a similar direction: 
https://github.com/eclipse/wildwebdeveloper/issues/404 --> here the request is to group constants to allow folding (which looks difficult to implement, since the group would need its own SymbolKind).
My interpretation is that the user wants to be able to see this information but it is not always useful to show it in the outline.

gerdleon:
Would you be open to a contribution of this feature?

gerdleon:
I do not think it is correct to say that it is a bug for a workaround in a language server. The language server will return as many elements as make sense - that is: if there is a constant, method, function, enum, enumKind, etc. it should be in the response even if there are many of them. 

How the result is visualized is up to the client. 

Taking Java as an example, there I would expect that all methods, all fields, all constants, etc. are returned from the LS.
But if a class contains 100 constants, it might be the fault of the developer of the class but it is not the fault of the LS.

Given there is limited space on a screen, it makes sense to filter out certain elements in the Outline View based on the users needs. I am not aware of specific support in LSP to configure the result returned from the LS.

Please remember that established products already implemented this functionality:
[...]

PR #1049, opened 2024-08-13
Add content filtering to outline view based on SymbolKind (fix issue #254)

travkin79:
Add generic implementation of content filtering in the outline view based on an element's SymbolKind and corresponding preferences, which make the user settings persistent.

### DIFF
--- a/org.eclipse.lsp4e/plugin.xml
+++ b/org.eclipse.lsp4e/plugin.xml
@@ -283,6 +283,27 @@
          </command>
       </menuContribution>
       <menuContribution
+            allPopups="false"
+            locationURI="menu:org.eclipse.ui.views.ContentOutline">
+         <menu
+               id="org.eclipse.lsp4e.outline.hideMenu"
+               label="Hide...">
+            <visibleWhen>
+               <reference
+                     definitionId="org.eclipse.lsp4e.activePartHasCNFOutlinePage">
+               </reference>
+            </visibleWhen>
+         </menu>
+      </menuContribution>
+      <menuContribution
+            allPopups="false"
+            locationURI="menu:org.eclipse.lsp4e.outline.hideMenu">
+         <dynamic
+               class="org.eclipse.lsp4e.outline.OutlineViewFilterMenuContributor"
+               id="org.eclipse.lsp4e.filters.dynamic">
+         </dynamic>
+      </menuContribution>
+      <menuContribution
             allPopups="true"
             locationURI="popup:org.eclipse.ui.genericeditor.source.menu?after=additions">
          <!-- "Source/Format" menu entry in editor popup menu -->
--- a/org.eclipse.lsp4e/plugin.xml
+++ b/org.eclipse.lsp4e/plugin.xml
@@ -299,7 +299,7 @@
             allPopups="false"
             locationURI="menu:org.eclipse.lsp4e.outline.hideMenu">
          <dynamic
-               class="org.eclipse.lsp4e.outline.OutlineViewFilterMenuContributor"
+               class="org.eclipse.lsp4e.outline.OutlineViewHideSymbolKindMenuContributor"
                id="org.eclipse.lsp4e.filters.dynamic">
          </dynamic>
       </menuContribution>
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/outline/LSSymbolsContentProvider.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/outline/LSSymbolsContentProvider.java
@@ -14,7 +14,7 @@
  *******************************************************************************/
 package org.eclipse.lsp4e.outline;
 
-import static org.eclipse.lsp4e.internal.NullSafetyHelper.*;
+import static org.eclipse.lsp4e.internal.NullSafetyHelper.lateNonNull;
 
 import java.net.URI;
 import java.net.URISyntaxException;
@@ -267,7 +267,7 @@ public class LSSymbolsContentProvider implements ICommonContentProvider, ITreeCo
 
 	public LSSymbolsContentProvider(boolean refreshOnResourceChanged) {
 		this.refreshOnResourceChanged = refreshOnResourceChanged;
-		this.preferencesDependantOutlineUpdater = new PreferencesChangedOutlineUpdater();
+		preferencesDependantOutlineUpdater = new PreferencesChangedOutlineUpdater();
 	}
 
 	@Override
@@ -318,10 +318,10 @@ public class LSSymbolsContentProvider implements ICommonContentProvider, ITreeCo
 
 	@Override
 	public Object[] getElements(@Nullable Object inputElement) {
-		if (this.symbols != null && !this.symbols.isDone()) {
+		if (symbols != null && !symbols.isDone()) {
 			return new Object[] { new PendingUpdateAdapter() };
 		}
-		if (this.lastError != null && symbolsModel.getElements().length == 0) {
+		if (lastError != null && symbolsModel.getElements().length == 0) {
 			return new Object[] { "An error occured, see log for details" }; //$NON-NLS-1$
 		}
 		return Arrays.stream(symbolsModel.getElements())
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/outline/LSSymbolsContentProvider.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/outline/LSSymbolsContentProvider.java
@@ -31,6 +31,9 @@ import org.eclipse.core.resources.IResourceChangeListener;
 import org.eclipse.core.resources.IResourceDelta;
 import org.eclipse.core.runtime.CoreException;
 import org.eclipse.core.runtime.IPath;
+import org.eclipse.core.runtime.preferences.IEclipsePreferences;
+import org.eclipse.core.runtime.preferences.IEclipsePreferences.IPreferenceChangeListener;
+import org.eclipse.core.runtime.preferences.IEclipsePreferences.PreferenceChangeEvent;
 import org.eclipse.core.runtime.preferences.InstanceScope;
 import org.eclipse.jdt.annotation.NonNullByDefault;
 import org.eclipse.jdt.annotation.Nullable;
@@ -53,10 +56,12 @@ import org.eclipse.lsp4e.LSPEclipseUtils;
 import org.eclipse.lsp4e.LanguageServerPlugin;
 import org.eclipse.lsp4e.LanguageServerWrapper;
 import org.eclipse.lsp4e.internal.CancellationUtil;
+import org.eclipse.lsp4e.outline.SymbolsModel.DocumentSymbolWithURI;
 import org.eclipse.lsp4e.ui.UI;
 import org.eclipse.lsp4j.DocumentSymbol;
 import org.eclipse.lsp4j.DocumentSymbolParams;
 import org.eclipse.lsp4j.SymbolInformation;
+import org.eclipse.lsp4j.SymbolKind;
 import org.eclipse.lsp4j.TextDocumentIdentifier;
 import org.eclipse.lsp4j.jsonrpc.messages.Either;
 import org.eclipse.ui.IMemento;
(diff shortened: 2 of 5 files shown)


org.eclipse.lsp4e/src/org/eclipse/lsp4e/outline/LSSymbolsContentProvider.java
```diff
@@ -144,6 +150,39 @@ public void documentAboutToBeChanged(DocumentEvent event) {
...
+			preferences.removePreferenceChangeListener(this);
+		}
+
+		@Override
+		public void preferenceChange(PreferenceChangeEvent event) {
+			if (viewer == null || outlineViewerInput == null) {
+				return;
+			}
+
+			if (event.getKey().startsWith(CNFOutlinePage.HIDE_DOCUMENT_SYMBOL_KIND_PREFERENCE_PREFIX)) {
```
rubenporras, line 176: Since `outlineViewerInput` is not used in this method, could we remove lines 172-175 and just add `viewer != null` to the `if`?

Note: I think that to avoid NullPointer warnings, you need assign `viewer` to a local such as `theViewer` and use that local in the rest of the method.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/outline/LSSymbolsContentProvider.java
```diff
@@ -292,12 +334,45 @@ public Object[] getElements(Object inputElement) {
...
+
+		if (element instanceof DocumentSymbol) {
+			kind = ((DocumentSymbol) element).getKind();
+		} else if (element instanceof DocumentSymbolWithURI) {
+			kind = ((DocumentSymbolWithURI) element).symbol.getKind();
+		} else if (element instanceof SymbolInformation) {
+			kind = ((SymbolInformation) element).getKind();
+		}
+
+		if (OutlineViewHideSymbolKindMenuContributor.isHideSymbolKind(kind)) {
```
rubenporras, line 371:
```suggestion
		return OutlineViewHideSymbolKindMenuContributor.isHideSymbolKind(kind);
```
and remove the rest of the method.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/outline/LSSymbolsContentProvider.java
```diff
@@ -292,12 +334,45 @@ public Object[] getElements(Object inputElement) {
...
+	 * are to be hidden in the outline view.
+	 * Sub-classes may override this method to add language-specific filtering.
+	 *
+	 * @param element an outline view contents element to be checked
+	 * @return <code>true</code> if the given element is not to be shown in the outline view
+	 */
+	protected boolean hideElement(Object element) {
+		SymbolKind kind = null;
+
+		if (element instanceof DocumentSymbol) {
```
rubenporras, line 363: Could we avoid the casts in this if/else if battery by using https://openjdk.org/jeps/394?

org.eclipse.lsp4e/src/org/eclipse/lsp4e/outline/OutlineViewHideSymbolKindMenuContributor.java
```diff
@@ -0,0 +1,87 @@
...
+public class OutlineViewHideSymbolKindMenuContributor extends CompoundContributionItem {
+
+	@Override
+	protected IContributionItem[] getContributionItems() {
+		return Arrays.stream(SymbolKind.values())
+			.sorted(new Comparator<SymbolKind>() {
+
+				@Override
+				public int compare(SymbolKind sk1, SymbolKind sk2) {
+					return sk1.name().compareToIgnoreCase(sk2.name());
```
rubenporras, line 39:
```suggestion
					return sk1.name().compareTo(sk2.name());
```
should be enough since the names in the Enum will be constant

org.eclipse.lsp4e/plugin.xml
```diff
@@ -282,6 +282,27 @@
             </visibleWhen>
          </command>
       </menuContribution>
+      <menuContribution
+            allPopups="false"
+            locationURI="menu:org.eclipse.ui.views.ContentOutline">
+         <menu
+               id="org.eclipse.lsp4e.outline.hideMenu"
+               label="Hide...">
```
rubenporras, line 290: If we can make it consistent with the JDT, I think that would be good. That would mean having a label such as

```suggestion
               label="Filter Elements...">
```

and the in the page a header such as "Select the elements to exclude from the view:". Do you think that would be a small improvement?

org.eclipse.lsp4e/src/org/eclipse/lsp4e/outline/LSSymbolsContentProvider.java
```diff
@@ -292,12 +334,45 @@ public Object[] getElements(Object inputElement) {
...
+			.toArray(Object[]::new);
+	}
+
+	/**
+	 * Decide whether to hide the given element or not.
+	 * Default implementation hides elements depending on a preference
+	 * that determines which {@link SymbolKind}s
+	 * (property in {@link DocumentSymbol}, {@link DocumentSymbolWithURI}s, and {@link SymbolInformation})
+	 * are to be hidden in the outline view.
+	 * Sub-classes may override this method to add language-specific filtering.
```
rubenporras, line 355: I think we we do not currently support such sub-classing. I would remove the comment and make the method private.

rubenporras (commented):
Hi @travkin79 ,

Thanks a lot for the contribution.

I have some minor comments on your PR. If you take a look at them and rebase your PR (it has conflicts now), I would be happy to merge it.

Regards

org.eclipse.lsp4e/src/org/eclipse/lsp4e/outline/LSSymbolsContentProvider.java
travkin79: Good point. I'll happily improve that. I often have to work with older Java versions and forgot that we can use the new Java features in LSP4E. Thank you for the hint.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/outline/LSSymbolsContentProvider.java
travkin79: I'll change that as suggested. Originally I used that variable here, but changed the code later and forgot to remove that dependency. Thanks.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/outline/LSSymbolsContentProvider.java
travkin79: Ok. I was not sure if LSSymbolContentProvider is meant to be extended. I'll make the method private and remove the comment.

org.eclipse.lsp4e/plugin.xml
travkin79: I'm not sure with this point. First, I thought, we should the same kind of filtering as in JDT, but due to limitations in the LSP specification that will not work (e.g. we cannot hide static or non-public members since `DocumentSymbol` from LSP4J and LSP spec provides no such details). Since our filters are a lot simpler than those from JDT, I thought, it would be a good idea ti simplify the dialog compared to JDT.

Besides, it would be more effort to implement the same dialog as in JDT, but IMHO it would not offer any improvement to the user.

If someday LSP gets extended to provide more details about access modifiers and similar details (see https://github.com/microsoft/language-server-protocol/issues/98), then we could change the dialog to look the same as in JDT and also add toolbar icons, e.g. for hiding non-public members, etc.

Do you agree?

rubenporras: Yes, I agree

travkin79:
Strange. All CI workflows / actions fail for some reason. There seems to be a connection problem.

image

org.eclipse.lsp4e/src/org/eclipse/lsp4e/outline/LSSymbolsContentProvider.java
```diff
@@ -139,6 +145,36 @@ public void documentAboutToBeChanged(DocumentEvent event) {
...
+			IEclipsePreferences preferences = InstanceScope.INSTANCE.getNode(LanguageServerPlugin.PLUGIN_ID);
+			preferences.removePreferenceChangeListener(this);
+		}
+
+		@Override
+		public void preferenceChange(PreferenceChangeEvent event) {
+			if (viewer != null
+					&& event.getKey().startsWith(CNFOutlinePage.HIDE_DOCUMENT_SYMBOL_KIND_PREFERENCE_PREFIX)) {
+				viewer.getControl().getDisplay().asyncExec(() -> {
+					if(viewer.getTree().isDisposed()) {
```
rubenporras, line 170: Since there are connection problems, what do you think of simplifying it further and remove the return as in

```
					if(!viewer.getTree().isDisposed()) {
						viewer.refresh();
					}
```
?

travkin79: Yes, I'll change that, too.

Since rebasing on main branch, I get a lot of Java problems like the following. Do I have to change some configuration. I'm using JDK 21 with source compatibility set to 17.

image

rubenporras: @sebthom , could you help with the errors? I suspect they are a follow-up of the last refactor.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/outline/OutlineViewHideSymbolKindMenuContributor.java
```diff
@@ -0,0 +1,87 @@
...
+		return !oldValue;
+	}
+
+	private static class HideSymbolKindAction extends Action {
+		private final SymbolKind kind;
+
+		HideSymbolKindAction(SymbolKind kind) {
+			super(kind.name(), IAction.AS_CHECK_BOX);
+			this.kind = kind;
+			this.setChecked(isHideSymbolKind(kind));
```
rubenporras, line 71:
```suggestion
			setChecked(isHideSymbolKind(kind));
```

org.eclipse.lsp4e/src/org/eclipse/lsp4e/outline/OutlineViewHideSymbolKindMenuContributor.java
```diff
@@ -0,0 +1,87 @@
...
+
+			Image img = LSPImages.imageFromSymbolKind(kind);
+			if (img != null) {
+				this.setImageDescriptor(ImageDescriptor.createFromImage(img));
```
rubenporras, line 75:
```suggestion
				setImageDescriptor(ImageDescriptor.createFromImage(img));
```

org.eclipse.lsp4e/src/org/eclipse/lsp4e/outline/LSSymbolsContentProvider.java
```diff
@@ -287,12 +324,33 @@ public Object[] getElements(@Nullable Object inputElement) {
 		if (this.lastError != null && symbolsModel.getElements().length == 0) {
```
rubenporras, line 324:
```suggestion
		if (lastError != null && symbolsModel.getElements().length == 0) {
```

org.eclipse.lsp4e/src/org/eclipse/lsp4e/outline/LSSymbolsContentProvider.java
```diff
@@ -225,17 +259,20 @@ public void resourceChanged(IResourceChangeEvent event) {
...
 	private @Nullable IOutlineUpdater outlineUpdater;
+	private IOutlineUpdater preferencesDependantOutlineUpdater;
 
 	public LSSymbolsContentProvider() {
 		this(false);
 	}
 
 	public LSSymbolsContentProvider(boolean refreshOnResourceChanged) {
 		this.refreshOnResourceChanged = refreshOnResourceChanged;
+		this.preferencesDependantOutlineUpdater = new PreferencesChangedOutlineUpdater();
```
rubenporras, line 270:
```suggestion
		preferencesDependantOutlineUpdater = new PreferencesChangedOutlineUpdater();
```

rubenporras (approved):

rubenporras:
thanks.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/outline/LSSymbolsContentProvider.java
travkin79: It seems, I had only to install the Maven Integration plug-ins that my IDE was missing. Maybe that was due to some pom-derived compile settings that were introduced with the latest commits in the main branch.

Issue #1051, opened 2024-08-13
DSPDebugTarget may leave dummy child process alive


FlorianKroiss:
When the DSPDebugTarget receives output before a process was started, it creates a dummy process so that it can write the output to a console. When the debugger reports another process later on, the direct reference to this dummy process (DSPDebugTarget.process) is overwritten. Later on, when terminate is called on DSPDebugTarget, it only tries to terminate the last reported process, leaving the dummy process alive.

I noticed that this happens when debugging Python with debugpy, which sends telemetry output before starting a process.
For my specific case it would be enough to defer creation of a dummy process until we actually have something to write to the console, so something that isn't telemetry data.

Steps to reproduce:

1. Install Python
2. Install [debugpy](https://pypi.org/project/debugpy/)
3. Create a new project in Eclipse and add a file called test.py containing a simple program, e.g., `print("foo")`
4. Create a new "Debug Adapter Launcher" launch configuration
5. Select "Launch a Debug Server using ..."
6. Set Command to your python executable
7. Set Arguments to `-m debugpy.adapter`
8. Set Launch Parameters to `{"request": "launch", "program": "path/to/your/test.py"}`
9. Press Debug to start Launch Configuration
10. See that an extra process is shown in the debug view, which is not terminated and prevents the launch from being removed
11. Performing step 9 multiple times causes Launches to pile up in Debug view

A workaround is to manually terminate the dummy process via the debug view, which allows to cleanup the launch

sebthom:
Your PR was merged and a new LSP4e version was released. Thanks for your contribution!

PR #1052, opened 2024-08-13
Defer creation of dummy process until we actually have something to output (#1051)

FlorianKroiss:
This partially fixes the problem described in #1051


rubenporras (approved):

rubenporras:
Thanks

PR #1053, opened 2024-08-14
Use the URI as the human readable name of the change instead of the path

rubenporras:
in the filesystem to support files which are only in the Eclipse filesystem (not in the IO filesystem)


rubenporras:
@Boy

rubenporras:
@BoykoAlex , as always, let me know if you think the PR is not good.

PR #1054, opened 2024-08-14
Add a protected method for easy configuration of retrieval of ConfigurationItem

rubenporras:
Configuration Items

### DIFF
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageClientImpl.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageClientImpl.java
@@ -13,6 +13,7 @@
  *******************************************************************************/
 package org.eclipse.lsp4e;
 
+import static org.eclipse.lsp4e.internal.NullSafetyHelper.castNonNull;
 import static org.eclipse.lsp4e.internal.NullSafetyHelper.lateNonNull;
 
 import java.util.ArrayList;
@@ -33,6 +34,7 @@ import org.eclipse.lsp4e.ui.Messages;
 import org.eclipse.lsp4e.ui.UI;
 import org.eclipse.lsp4j.ApplyWorkspaceEditParams;
 import org.eclipse.lsp4j.ApplyWorkspaceEditResponse;
+import org.eclipse.lsp4j.ConfigurationItem;
 import org.eclipse.lsp4j.ConfigurationParams;
 import org.eclipse.lsp4j.Location;
 import org.eclipse.lsp4j.MessageActionItem;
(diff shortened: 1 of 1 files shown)


rubenporras:
Hi @sebthom ,

I could not find a way to override the method `CompletableFuture<List<@Nullable Object>> configuration(ConfigurationParams configurationParams)` after the new annotations, as the definition in `LanguageClientImpl` and `LanguageClient` do not have the same annotations. 

To me this is a bug in the JDT, but the new protected method is in any case a better API for subclasses than the previous one.

What do you think of this PR?

sebthom:
Hi @rubenporras I don't understand why you can't find a way to override it. It seems to work fine for me:
```java
package org.haxe4e.langserver;

import java.util.List;

import org.eclipse.jdt.annotation.*;
import org.eclipse.lsp4e.LanguageClientImpl;
import org.eclipse.lsp4j.*;
import org.haxe4e.util.TreeBuilder;

@SuppressWarnings("restriction") // https://bugs.eclipse.org/bugs/show_bug.cgi?id=536215
public final class HaxeLangServerClientImpl extends LanguageClientImpl implements HaxeLangServerClient {

   private volatile boolean isInitTriggered = false;

   @Override
   public CompletableFuture<List<@Nullable Object>> configuration(final ConfigurationParams configurationParams) {
      return super.configuration(configurationParams);
   }

   @Override
   public @NonNullByDefault({}) CompletableFuture<Void> registerCapability(final RegistrationParams params) {
      if (!isInitTriggered) {
         // workaround for https://github.com/vshaxe/vshaxe/issues/501
         final var event = new DidChangeConfigurationParams(new TreeBuilder<String>() //
            .put("haxe", new TreeBuilder<String>() //
               .getMap() //
            ).getMap());

         getLanguageServer().getWorkspaceService().didChangeConfiguration(event);
         isInitTriggered = true;
      }

      return super.registerCapability(params);
   }
}
```

What is the error or warning you are seeing?

rubenporras:
Hi @sebthom,

I do not know why, but changing the configuration item `org.eclipse.jdt.core.compiler.annotation.inheritNullAnnotations` from `disabled` to `enabled` and restarting eclipse seems to solve the problem. I have tried changing the configuration yesterday already but it looks like the build that is triggered afterwards is not enough, and a restart is needed.

Does it makes sense to you?

sebthom:
do you still have the issue now that you merged the no-npe 1.0.4 version?
I made a change to the EEA of LanguageClientImpl at https://github.com/vegardit/no-npe/commit/ec54909964f93ac9284383ae4ad34cfe731d1f29 which might solve the issue.

rubenporras:
The problem is solved for me now. Thanks.

sebthom:
Nice!

PR #1060, opened 2024-08-15
refact: enable NonNullByDefault for lsp4e.debug and lsp4e.jdt plugins

sebthom:


sebthom:
As part of this PR I had to copy parts of the org.eclipse.lsp4e.internal.NullSafetyHelper to the lsp4e.debug plugin.
Copying code has been done before too where org.eclipse.lsp4e.debug.breakpoints.DocumentUtils only contains code copied from org.eclipse.lsp4e.LSPEclipseUtils.

@mickaelistria is there any real-world practical reason why lsp4e.debug should not depend on the lsp4e plugin to reuse these classes?

mickaelistria:
> is there any real-world practical reason why lsp4e.debug should not depend on the lsp4e plugin to reuse these classes?

IIRC, at the beginning, the goal was to get the Debug Adapter working for CDT without having LSP-based edition installed. It's the only case I'm aware of and I don't know if this still stands. @jonahgraham may know better.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/command/CommandExecutor.java
```diff
@@ -90,7 +90,7 @@ public class CommandExecutor {
 		return CompletableFuture.completedFuture(null);
 	}
 
-	@SuppressWarnings("unused") // ECJ compiler handlerService cannot be null because getService is declared as
+	// ECJ compiler handlerService cannot be null because getService is declared as
```
rubenporras, line 93: This comment can now be removed, right?

org.eclipse.lsp4e/src/org/eclipse/lsp4e/command/CommandExecutor.java
```diff
@@ -128,7 +125,7 @@ public class CommandExecutor {
 		return null;
 	}
 
-	@SuppressWarnings("unused") // ECJ compiler thinks commandService cannot be null (see above)
+	// ECJ compiler thinks commandService cannot be null (see above)
```
rubenporras, line 128: Same, this comment can now be removed, right?

sebthom:
@rubenporras thanks for the quick review!

jonahgraham:
> IIRC, at the beginning, the goal was to get the Debug Adapter working for CDT without having LSP-based edition installed. It's the only case I'm aware of and I don't know if this still stands. @jonahgraham may know better.

Going forward there is no reason to keep them separate. CDT is moving to use LSP (https://github.com/eclipse-cdt/cdt-lsp) and there probably isn't an install anywhere that has lsp4e.debug and not main lsp4e.

PR #1068, opened 2024-08-16
perf: improve StreamProxyInputStream

sebthom:
the current implementation is sub optimal as it creates a new queue element for each byte read from the process.
the new implementation queues arrays of bytes instead resulting in fewer queue elements and in avoiding `byte -> Byte -> byte` conversions. this also reduces GC pressure in the Eclipse IDE process.

I also implemented `read(final byte[] buf, final int off, final int len) ` which also improves bulk read performance.


mickaelistria:
Thank you for this interesting improvement!

PR #1069, opened 2024-08-19
fix: DocumentInputStream does not handle surrogate pairs correctly

sebthom:
This PR replaces the current DocumentInputStream implementation which does not handle high/low surrogate character pairs.

### DIFF
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/internal/CharsInputStream.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/internal/CharsInputStream.java
@@ -1,0 +1,295 @@
+/*******************************************************************************
+ * Copyright (c) 2024 Sebastian Thomschke and others.
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ * Sebastian Thomschke - initial implementation
+ *******************************************************************************/
+package org.eclipse.lsp4e.internal;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.nio.ByteBuffer;
+import java.nio.CharBuffer;
+import java.nio.charset.CharacterCodingException;
+import java.nio.charset.Charset;
+import java.nio.charset.CharsetEncoder;
+import java.nio.charset.CoderResult;
+import java.util.Objects;
+import java.util.function.IntSupplier;
+
+public class CharsInputStream extends InputStream {
+
+	/**
+	 * Functional interface for supplying characters at a specified index.
+	 * Implementations can define how characters are fetched.
+	 */
+	@FunctionalInterface
+	public interface CharsSupplier {
+		char charAt(int index) throws Exception;
+	}
+
+	private enum EncoderState {
+		/**
+		 * The {@link #encoder} is actively encoding characters into bytes. This is the
+		 * initial state of the encoder.
+		 */
+		ENCODING, //
+
+		/**
+		 * The {@link #encoder} has finished processing all characters and is now
+		 * flushing any remaining bytes in its internal buffer.
+		 */
+		FLUSHING, //
+
+		/**
+		 * The {@link #encoder} has completed both the encoding and flushing processes.
+		 * No more data is left to be read from the encoder.
+		 */
+		DONE
+	}
+
+	public static final char UNICODE_REPLACEMENT_CHAR = '\uFFFD';
+
+	/** 512 surrogate character pairs */
+	private static final int DEFAULT_BUFFER_SIZE = 512;
+	private static final int EOF = -1;
+
+	private final int bufferSize;
+	private final CharBuffer charBuffer;
+	private final ByteBuffer byteBuffer;
+	private final CharsetEncoder encoder;
+	private EncoderState encoderState = EncoderState.ENCODING;
+
+	private int charIndex = 0;
+	private final CharsSupplier chars;
+	private final IntSupplier charsLength;
+
+	public CharsInputStream(final CharSequence chars) {
+		this(chars, Charset.defaultCharset());
+	}
+
+	public CharsInputStream(final CharSequence chars, final Charset charset) {
+		this(chars, charset, DEFAULT_BUFFER_SIZE);
+	}
+
+	public CharsInputStream(final CharSequence chars, final Charset charset, final int bufferSize) {
+		this(chars::charAt, chars::length, charset, bufferSize);
+	}
+
+	public CharsInputStream(final CharsSupplier chars, final IntSupplier charsLength) {
+		this(chars, charsLength, Charset.defaultCharset());
+	}
+
+	/**
+	 * @param chars
+	 *            function to access indexed chars.
+	 * @param charsLength
+	 *            function to get the number of indexed chars provided by the
+	 *            <code>chars</code> parameter.
+	 */
+	public CharsInputStream(final CharsSupplier chars, final IntSupplier charsLength, final Charset charset) {
+		this(chars, charsLength, charset, DEFAULT_BUFFER_SIZE);
+	}
+
+	/**
+	 * @param chars
+	 *            function to access indexed chars.
+	 * @param charsLength
+	 *            function to get the number of indexed chars provided by the
+	 *            <code>chars</code> parameter.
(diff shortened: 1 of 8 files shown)


org.eclipse.lsp4e/src/org/eclipse/lsp4e/internal/CharsInputStream.java
```diff
@@ -0,0 +1,224 @@
...
+			encoderState = EncoderState.FLUSHING;
+		}
+
+		// flush
+		byteBuffer.clear();
+		final CoderResult result = encoder.flush(byteBuffer);
+		byteBuffer.flip();
+
+		if (result.isOverflow()) // byteBuffer too small
+			return true;
```
rubenporras, line 121: why not `result.throwException()`?

org.eclipse.lsp4e/src/org/eclipse/lsp4e/internal/CharsInputStream.java
```diff
@@ -0,0 +1,224 @@
...
+
+	public CharsInputStream(final CharSequence chars, final @Nullable Charset charset) {
+		this(chars, charset, DEFAULT_BUFFER_SIZE);
+	}
+
+	public CharsInputStream(final CharSequence chars, final @Nullable Charset charset, final int bufferSize) {
+		this(chars::charAt, chars::length, charset, bufferSize);
+	}
+
+	public CharsInputStream(final CharsSupplier chars, final IntSupplier charsLength) {
```
rubenporras, line 65: This method is unused, should we remove it?

org.eclipse.lsp4e/src/org/eclipse/lsp4e/internal/DocumentInputStream.java
```diff
@@ -0,0 +1,45 @@
...
+import java.nio.charset.Charset;
+
+import org.eclipse.core.filebuffers.FileBuffers;
+import org.eclipse.core.filebuffers.ITextFileBuffer;
+import org.eclipse.core.filebuffers.ITextFileBufferManager;
+import org.eclipse.jdt.annotation.Nullable;
+import org.eclipse.jface.text.IDocument;
+import org.eclipse.lsp4e.LanguageServerPlugin;
+
+public final class DocumentInputStream extends CharsInputStream {
```
rubenporras, line 23: Can we split the PR in two commits so that we do not lose the git history of the file?

org.eclipse.lsp4e/src/org/eclipse/lsp4e/internal/CharsInputStream.java
sebthom: because we can handle it. it just means that the bytebuffer we use for char processing is to small in the flushing phase when the encoder may return (based on the charset) other additional bytes.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/internal/CharsInputStream.java
```diff
@@ -0,0 +1,224 @@
...
+public class CharsInputStream extends InputStream {
+	@FunctionalInterface
+	public interface CharsSupplier {
+		char charAt(int index) throws Exception;
+	}
+
+	private enum EncoderState {
+		ENCODING,
+		FLUSHING,
+		DONE
```
rubenporras, line 36:
```suggestion
		FLUSHED
```

sebthom: DONE in this context means finished, i.e. the encoder will not be used to refill buffers. Flushed could mean the encoder switches to encoding again and flushes again.

rubenporras: Thanks. I am having some trouble understanding some parts of the logic, as you see :) 

What about adding a comment explaining what the DONE state means?

sebthom: I added some javadoc. I hope it helps :-)

org.eclipse.lsp4e/src/org/eclipse/lsp4e/internal/CharsInputStream.java
```diff
@@ -0,0 +1,224 @@
...
+	public int available() {
+		final int remaining = byteBuffer.remaining();
+		return remaining == 0 ? charsLength.getAsInt() - charIndex : remaining;
+	}
+
+	public Charset getCharset() {
+		return encoder.charset();
+	}
+
+	private boolean flushEncoder() throws IOException {
```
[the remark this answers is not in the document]
sebthom: I added some javadoc.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/internal/CharsInputStream.java
```diff
@@ -0,0 +1,276 @@
...
+
+		if (encoderState == EncoderState.FLUSHING)
+			return flushEncoder();
+
+		final int charsLen = charsLength.getAsInt();
+
+		// if EOF is reached transition to flushing
+		if (charIndex >= charsLen) {
+			// finalize encoding before switching to flushing
+			byteBuffer.clear();
```
rubenporras, line 228: what about adding a method such as (maybe choose a better name if you would know a better name)
```
	private void encodeByteBuffer(CharBuffer in, boolean endOfInput) throws CharacterCodingException {
		byteBuffer.clear();
		final CoderResult result = encoder.encode(in, byteBuffer, endOfInput /* signal EOF */);
		byteBuffer.flip();
		if (result.isError()) {
			result.throwException();
		}
	}
```
and call here just `encodeByteBuffer(CharBuffer.allocate(0),  true);` and below `encodeByteBuffer(CharBuffer.allocate(0),  false);`?

org.eclipse.lsp4e/src/org/eclipse/lsp4e/internal/CharsInputStream.java
```diff
@@ -0,0 +1,276 @@
...
+				if (Character.isHighSurrogate(nextChar)) { // handle surrogate pairs
+					if (charIndex < charsLen) {
+						final char lowSurrogate = chars.charAt(charIndex);
+						if (Character.isLowSurrogate(lowSurrogate)) {
+							charIndex++;
+							charBuffer.put(nextChar);
+							charBuffer.put(lowSurrogate);
+						} else {
+							// missing low surrogate - fallback to replacement character
+							charBuffer.put('\uFFFD');
```
rubenporras, line 250: extract `'\uFFFD'` to a constant?

org.eclipse.lsp4e/src/org/eclipse/lsp4e/internal/CharsInputStream.java
```diff
@@ -0,0 +1,276 @@
...
+			final CoderResult result = encoder.encode(CharBuffer.allocate(0), byteBuffer, true /* signal EOF */);
+			byteBuffer.flip();
+			if (result.isError()) {
+				result.throwException();
+			}
+			return flushEncoder();
+		}
+
+		try {
+			charBuffer.clear();
```
rubenporras, line 238: should we not do the `charBuffer.clear` rather at the end of the method in a `finally` block so that we do not hold data that will never be used until the next call to refill?

rubenporras:
@sebthom , I do not have more comments.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/internal/CharsInputStream.java
sebthom: the charBuffer contains a fixed size char[] array. I don't see any advantage clearing the buffer at the end. I prefer clearing the buffer just before I starting refilling and using it.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/internal/CharsInputStream.java
sebthom: good idea. done

org.eclipse.lsp4e/src/org/eclipse/lsp4e/internal/CharsInputStream.java
sebthom: I added a method called encodeChars(). Thanks for the suggestion

sebthom:
> @sebthom , I do not have more comments.

oh man, you are thorough!

PR #1074, opened 2024-08-21
fix: don't eagerly set temporary launcherFuture

sebthom:
Setting a temporary launcherFuture before the actual initialization results in LanguageServerWrapper#isActive to return true despite the connection to the language server process has not yet been established.


sebthom:
@mickaelistria can you please have a look at this? the current behavior is problematic because is the temporarily assigned future is accessible but gets replaced during initialization.

sebthom:
@rubenporras I think you wanted to cut another release soon. would be great if we could sort out this issue before

rubenporras:
I have not concrete plans at the moment. It is good to be able to do periodic releases on when there is enough fixes.

So at the moment no hurry on this one from my side.

rubenporras:
In a attempt to move the review forward. @sebthom , have you checked [the PR where the code was introduced](https://github.com/eclipse/lsp4e/pull/611)? As by that PR, it was preventing an infinite start/stop loop.

Do you think this not happen any longer because some other code change? Otherwise, should not have a more elaborate fix that replaces the "naive and simple fix" (quoting the first comment on #611)?

sebthom:
Thanks for pointing that out. I will do some more investigation.

rubenporras:
I would like to let you know that I would like to do release at the beginning of the next week.

sebthom:
OK. This won't make it. I already went into the rabbit hole of concurrency and future handling, there is more to be done...

PR #1075, opened 2024-08-22
Add hide fields toolbar button to outline view (fix #1073)

### LINKED ISSUE #1073: Add "hide fields" filter to outline view's toolbar
I'd like to add a toolbar action to the outline view for hiding fields. This filter is very common and is available, e.g. for Java (JDT) and C++ (CDT). It would look like very much like in JDT (selected button in the following screenshot):

image

Are there any objections on implementing that filter in LSP4E (for all language servers)?

This is a follow-up issue to #254.

travkin79:
Hi @rubenporras,
I'm working on the "hide fields" toolbar entry in the outline view and it works already.

Nevertheless, I think, I need to improve the implementation. I'd like to improve how the states of the menu entry "Hide..." -> "Field" and the toolbar entry "Hide Fields" are synchronized in the outline view. Currently I let the Hide Fields handler register for preference changes and do update the command's state. Do you have a better solution for that?

Besides, I observe some strange behavior in the outline view. As soon as the outline looses focus (by clicking into the CDT LSP editor in my case), the sort and hide fields buttons disappear. I observed that already before preparing my PR with the sort toolbar entry. I'm not sure if this is going wrong in LSP4E or in CDT LSP. Can you observe the same issue?

Another minor issue is that the toolbar buttons icon look different to the icon used in the menu entry "Hide..." > "Field" (and other places). But I could live with that for now.

Thanks and best regards.


org.eclipse.lsp4e/src/org/eclipse/lsp4e/outline/ToggleHideFieldsOutlineHandler.java
```diff
@@ -0,0 +1,74 @@
...
+
+	@Override
+	public void dispose() {
+		super.dispose();
+		preferences.removePreferenceChangeListener(this);
+	}
+
+	@Override
+	public void preferenceChange(PreferenceChangeEvent event) {
+		if (toggleHideFieldsCommand.isPresent()
```
rubenporras, line 64: what about using ifPresent as in
```
  if (event.getKey().startsWith(CNFOutlinePage.HIDE_DOCUMENT_SYMBOL_KIND_PREFERENCE_PREFIX)
      && event.getKey().endsWith(SymbolKind.Field.name())) {
    toggleHideFieldsCommand.ifPresent(command ->
      try {
        HandlerUtil.toggleCommandState(command);
      } catch (ExecutionException e) {
      }
    );
  }
```?

Comment (travkin79) [reply org.eclipse.lsp4e/src/org/eclipse/lsp4e/outline/ToggleHideFieldsOutlineHandler.java:64]: I'll changed that. Thanks. What about the other points? Any idea?

Comment (travkin79) [reply org.eclipse.lsp4e/src/org/eclipse/lsp4e/outline/ToggleHideFieldsOutlineHandler.java:64]: What about something like the following?

```java
if (toggleHideFieldsCommand.isEmpty()) {
			return;
		}

		if (event.getKey().startsWith(CNFOutlinePage.HIDE_DOCUMENT_SYMBOL_KIND_PREFERENCE_PREFIX)
				&& event.getKey().endsWith(SymbolKind.Field.name())) {
			try {
				HandlerUtil.toggleCommandState(toggleHideFieldsCommand.get());
			} catch (ExecutionException e) {
			}
		}
```

rubenporras:
@travkin79 , shall I merge it now once the checks pass?

travkin79:
Thank you @rubenporras

PR #1076, opened 2024-08-23
refact: cleanup SupportedFeatures class

sebthom:
This PR cleans up the SupportedFeatures class:
1. Format code for better readability
1. use primitive `true` instead of verbose Boolean.TRUE
1. use `SymbolKind.values()` instead of naming each symbol kind separately

### DIFF
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/internal/SupportedFeatures.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/internal/SupportedFeatures.java
@@ -63,10 +63,14 @@ public class SupportedFeatures {
 
 	public static TextDocumentClientCapabilities getTextDocumentClientCapabilities() {
 		final var textDocumentClientCapabilities = new TextDocumentClientCapabilities();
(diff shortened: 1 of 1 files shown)


sebthom:
@rubenporras I saw that we are currently not announcing the capability `CodeAction.SourceFixAll`. Is that omitted on purpose?

rubenporras:
> @rubenporras I saw that we are currently not announcing the capability `CodeAction.SourceFixAll`. Is that omitted on purpose?

Not that I am aware of, but I came to LSP4E when most of the code was already in place. Maybe @mickaelistria knows?

rubenporras (approved):

mickaelistria:
> I saw that we are currently not announcing the capability CodeAction.SourceFixAll. Is that omitted on purpose?

`CodeAction.SourceFixAll` was introduced relatively recently. We forgot to declare it when it was added to the LSP spec and LSP4J, and since it has caused no issue so far, it remained forgotten.

sebthom:
@mickaelistria is it as simple as adding it to the list of supported features or do we need to do anything else?

mickaelistria:
I don't know for this particular case. Some new LSP features require some development in LSP4E, some do not and can just be declared as supported...

PR #1079, opened 2024-08-23
refact: improve handling of arrays

sebthom:
This PR improves working with arrays, by avoiding unnecessary temporary array allocations and providing specialized functions for filtering/iterating arrays which improves readability and avoids duplicate code.


sebthom:
@rubenporras I did not want to name the empty arrays as verbose as `EMPTY_BYTE_ARRAY` so I went with `ZERO_BYTES` but would I actually prefer to name them even shorter like `NO_BYTES`, `NO_OBJECTS` but I am unsure if this could result in misunderstandings when reading the code. WDYT?

rubenporras:
> @rubenporras I did not want to name the empty arrays as verbose as `EMPTY_BYTE_ARRAY` so I went with `ZERO_BYTES` but would I actually prefer to name them even shorter like `NO_BYTES`, `NO_OBJECTS` but I am unsure if this could result in misunderstandings when reading the code. WDYT?

I think that if you know that it is beneficial to reuse empty arrays to reduce creation/garbage collection, all names are fine, and otherwise you will not understand it.

What about choosing the names you like and commenting the class `ArrayUtil` with the explanation of the pattern?

PR #1080, opened 2024-08-23
perf: improve DSPLaunchDelegate debug adapter monitor performance

sebthom:
This commit replaces `Collections.synchronizedList(new LinkedList<Byte>())` with `new ConcurrentLinkedQueue<Byte>()` and only determines the console encoding once and not on every text output. 

If the console encoding was not specified in the launch config it now falls back to the JVM's default encoding instead of throwing an exception.

### DIFF
--- a/org.eclipse.lsp4e.debug/src/org/eclipse/lsp4e/debug/launcher/DSPLaunchDelegate.java
+++ b/org.eclipse.lsp4e.debug/src/org/eclipse/lsp4e/debug/launcher/DSPLaunchDelegate.java
@@ -13,12 +13,12 @@ import static org.eclipse.lsp4e.internal.NullSafetyHelper.castNonNull;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
+import java.nio.charset.Charset;
 import java.util.ArrayList;
-import java.util.Collections;
 import java.util.HashMap;
-import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.ConcurrentLinkedQueue;
 import java.util.function.Supplier;
 
 import org.eclipse.core.runtime.CoreException;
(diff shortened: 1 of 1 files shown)


org.eclipse.lsp4e.debug/src/org/eclipse/lsp4e/debug/launcher/DSPLaunchDelegate.java
```diff
@@ -323,16 +324,14 @@ public int read() throws IOException {
 							return -1;
 						}
 					};
-					DSPLaunchDelegateLaunchBuilder finalBuilder = builder;
+					final var consoleEncoding = builder.launch.getAttribute(DebugPlugin.ATTR_CONSOLE_ENCODING);
+					final var consoleCharset = consoleEncoding == null //
```
rubenporras, line 328: could you add to the PR description why are you adding this two calls to Charset instead of just passing `consoleEncoding`?

sebthom: consoleEncoding might return null or what do you mean? I moved the instruction out of the inner loop in the listener

rubenporras: Previously the value returned by `builder.launch.getAttribute(DebugPlugin.ATTR_CONSOLE_ENCODING)` was passed to `text.getBytes` as is (just casting to `NonNull`).
Now if the value is null we map it to a default charset. Is that an intentional change? If so I think it would be worth putting in the commit and PR messages.

sebthom: Because getBytes(String) would first lookup the CharSet object. If we obtain the Charset before and directly pass it to getBytes() we save that lookup on each listener event.

rubenporras: Thanks for the explanation. To be equivalent as before, we should throw a NPE if `builder.launch.getAttribute(DebugPlugin.ATTR_CONSOLE_ENCODING)`, otherwise we are changing the behavior, are we not?

I am fine changing the behaviour, but I think it should be added to the commit an PR messages.

sebthom: I actually think falling back to the default charset is more sensible than throwing an exception if the encoding is not overwritten in the launch config.

rubenporras: Yes, me too. 

Could you amend the commit message and add "If the console encoding was not specified in the launch config it now falls back to the JVM's default encoding instead of throwing an exception." there as well?

Then I am good

PR #1082, opened 2024-08-26
refactor: improve code reuse

rubenporras:


org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/completion/LSContentAssistProcessor.java
```diff
@@ -341,6 +327,16 @@ private static IContextInformation toContextInformation(SignatureInformation inf
 		return new ContextInformation(information.getLabel(), signature.toString());
 	}
 
+	private void cancelFuture(@Nullable CompletableFuture<List<Void>> future) {
```
sebthom, line 330: how about adding it as `public static void cancelFuture(@Nullable Future<?> future)` to `org.eclipse.lsp4e.internal.CancellationUtil` and reuse it in other places too?

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/completion/LSContentAssistProcessor.java
```diff
@@ -216,40 +214,28 @@ private String createErrorMessage(int offset, Exception ex) {
...
 						}
 						return CompletableFuture.completedFuture(null);
 					});
-			this.contextInformationLanguageServersFuture = LanguageServers.forDocument(document)
+			contextInformationLanguageServersFuture = LanguageServers.forDocument(document)
 					.withFilter(capabilities -> capabilities.getSignatureHelpProvider() != null) //
 					.collectAll((w, ls) -> {
-						SignatureHelpOptions provider = castNonNull(w.getServerCapabilities()).getSignatureHelpProvider();
+						List<String> triggerChars = castNonNull(w.getServerCapabilities()).getSignatureHelpProvider().getTriggerCharacters();
 						synchronized (contextTriggerCharsSemaphore) {
```
sebthom, line 237: I know, not part of your PR, but I think these synchronized blocks can be avoided too. mergeTriggers is a static method so no need for synchronization and contextTriggerChars can be assigned atomicly. I.e. declaring contextTriggerChars volatile should be enough and actually better.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/completion/LSContentAssistProcessor.java
rubenporras: Following up on your comment, I cannot find any other place in the codebase that uses this idiom. I have them read the documentation of `CompletableFuture.cancel(boolean)` and checked the implementation, and I think the exception will never be thrown. So we can just remove the catch block. 

Do you agree?

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/completion/LSContentAssistProcessor.java
rubenporras: Great, I will do that after I have an answer to the other thread in this review.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/completion/LSContentAssistProcessor.java
sebthom: Yes. Reading the JavaDoc and checking the implementation it looks like it is never thrown. So removing the try catch should be ok

sebthom: I still would suggest adding this as `public static void cancelFuture(@Nullable Future<?> future)` to `org.eclipse.lsp4e.internal.CancellationUtil` and reuse it in other places. There are a few more places where a future is first checked for null and then canceled and if we have a method here already why not use it everywhere else.

sebthom (approved):

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/completion/LSContentAssistProcessor.java
rubenporras: I will take a look.

sebthom (approved):

Issue #899, opened 2024-01-24
BadLocationException when editing a target file


laeubi:
I edit an PDE target file (latest i build) and getting the exception on save:

```
org.eclipse.jface.text.BadLocationException
	at org.eclipse.jface.text.TreeLineTracker.fail(TreeLineTracker.java:1054)
	at org.eclipse.jface.text.TreeLineTracker.nodeByLine(TreeLineTracker.java:302)
	at org.eclipse.jface.text.TreeLineTracker.getLineLength(TreeLineTracker.java:1116)
	at org.eclipse.jface.text.AbstractLineTracker.getLineLength(AbstractLineTracker.java:240)
	at org.eclipse.wst.sse.core.internal.text.BasicStructuredDocument.getLineLength(BasicStructuredDocument.java:1486)
	at org.eclipse.lsp4e.LSPEclipseUtils.getLineLength(LSPEclipseUtils.java:211)
	at org.eclipse.lsp4e.LSPEclipseUtils.toOffset(LSPEclipseUtils.java:204)
	at org.eclipse.lsp4e.operations.diagnostics.LSPDiagnosticsToMarkers.getExistingMarkerFor(LSPDiagnosticsToMarkers.java:214)
	at org.eclipse.lsp4e.operations.diagnostics.LSPDiagnosticsToMarkers$1.runInWorkspace(LSPDiagnosticsToMarkers.java:158)
	at org.eclipse.core.internal.resources.InternalWorkspaceJob.run(InternalWorkspaceJob.java:43)
	at org.eclipse.core.internal.jobs.Worker.run(Worker.java:63)
```

system is linux.

sebthom:
Thanks for reporting. The issue cannot be fully avoided because of the asynchronous nature of LSP. However you should see way less exceptions in the log after #1083. I am closing the ticket for now. Feel free to reopen if required.

Issue #203, opened 2022-08-09
BadLocationException SymbolsLabelProvider


jukzi:
eclipse sometimes logs this - do not know how how exactly to reproduce:
```
org.eclipse.jface.text.BadLocationException
	at org.eclipse.jface.text.TreeLineTracker.fail(TreeLineTracker.java:1054)
	at org.eclipse.jface.text.TreeLineTracker.getLineInformation(TreeLineTracker.java:1163)
	at org.eclipse.jface.text.AbstractLineTracker.getLineInformation(AbstractLineTracker.java:137)
	at org.eclipse.jface.text.AbstractDocument.getLineInformation(AbstractDocument.java:882)
	at org.eclipse.core.internal.filebuffers.SynchronizableDocument.getLineInformation(SynchronizableDocument.java:290)
	at org.eclipse.lsp4e.LSPEclipseUtils.toOffset(LSPEclipseUtils.java:164)
	at org.eclipse.lsp4e.outline.SymbolsLabelProvider.getMaxSeverity(SymbolsLabelProvider.java:184)
	at org.eclipse.lsp4e.outline.SymbolsLabelProvider.getImage(SymbolsLabelProvider.java:161)
	at org.eclipse.ui.internal.navigator.NavigatorContentServiceLabelProvider.findImage(NavigatorContentServiceLabelProvider.java:204)
	at org.eclipse.ui.internal.navigator.NavigatorContentServiceLabelProvider.getColumnImage(NavigatorContentServiceLabelProvider.java:112)
	at org.eclipse.ui.internal.navigator.NavigatorContentServiceLabelProvider.getImage(NavigatorContentServiceLabelProvider.java:104)
	at org.eclipse.ui.internal.navigator.NavigatorDecoratingLabelProvider$StyledLabelProviderAdapter.getImage(NavigatorDecoratingLabelProvider.java:63)
	at org.eclipse.jface.viewers.DelegatingStyledCellLabelProvider.getImage(DelegatingStyledCellLabelProvider.java:198)
	at org.eclipse.jface.viewers.DecoratingStyledCellLabelProvider.getImage(DecoratingStyledCellLabelProvider.java:171)
	at org.eclipse.jface.viewers.DelegatingStyledCellLabelProvider.update(DelegatingStyledCellLabelProvider.java:124)
	at org.eclipse.jface.viewers.DecoratingStyledCellLabelProvider.update(DecoratingStyledCellLabelProvider.java:134)
[...]

PR #1083, opened 2024-08-27
fix: don't log same BadLocationException excessively

sebthom:
The goal of this PR is to address https://github.com/eclipse/lsp4e/issues/899 , https://github.com/eclipse/lsp4e/issues/536 and https://github.com/eclipse/lsp4e/issues/203 by logging BadLocationExceptions in LSPDiagnosticsToMarkers and SymbolsLabelProvider only when `Platform.getDebugOption("org.eclipse.lsp4e/debug")` is set to true.

### DIFF
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServerPlugin.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServerPlugin.java
@@ -11,17 +11,33 @@
  *******************************************************************************/
 package org.eclipse.lsp4e;
 
+import static java.nio.charset.StandardCharsets.UTF_8;
+
+import java.util.concurrent.ConcurrentHashMap;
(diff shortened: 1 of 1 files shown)


rubenporras:
I prefer to have the exceptions in the log if the error happens. 

Otherwise the user will still miss some feature and will not know why.

Is the error such a problem that we would like to hide it?

sebthom:
I guess the problem is the amount of exceptions being logged repetitively

rubenporras:
What about counting exceptions and only log the first three unless in debug mode?

I think that is a good compromise between informing the user and not bothering him

sebthom:
> What about counting exceptions and only log the first three unless in debug mode?
> 
> I think that is a good compromise between informing the user and not bothering him

@rubenporras ok, please check my update

rubenporras (approved):
Thanks

PR #1084, opened 2024-08-27
refactor: remove duplicate null check

rubenporras:
Remove a duplicated null check (already checked at the very beginning of the method)

### DIFF
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/codeactions/CodeActionMarkerResolution.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/codeactions/CodeActionMarkerResolution.java
@@ -81,12 +81,10 @@ public class CodeActionMarkerResolution extends WorkbenchMarkerResolution implem
 				}
 			}
 			if (wrapper != null) {
-				if (codeAction.getEdit() == null) {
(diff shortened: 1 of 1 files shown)


org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/codeactions/CodeActionMarkerResolution.java
```diff
@@ -81,12 +81,10 @@ public void run(IMarker marker) {
...
-						if (resolvedCodeAction != null) {
-							codeAction = resolvedCodeAction;
-						}
+				if (CodeActionCompletionProposal.isCodeActionResolveSupported(wrapper.getServerCapabilities())) {
+					CodeAction resolvedCodeAction = wrapper.execute(ls -> ls.getTextDocumentService().resolveCodeAction(codeAction)).get(2, TimeUnit.SECONDS);
+					if (resolvedCodeAction != null) {
+						codeAction = resolvedCodeAction;
 					}
 				}
 				if (codeAction.getEdit() != null) {
```
sebthom, line 90: Because of the check if (codeAction.getEdit() != null) at the beginning of the method with an early exit. This condition here can never be true meaning this whole block is redundant

sebthom (approved):

PR #1086, opened 2024-08-28
refact: improve handling of futures

sebthom:
This PR moves generic future utility functions out of the LanguageServers class into a separate internal FutureUtil class.

It also renames the `addAll()` function to `join()` and changes the implementation in the way that not the result list of the first future is extended with entries of a secondary future (which makes the unsafe assumption the first result list is modifiable) but adds all items into a new list.

### DIFF
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/internal/FutureUtil.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/internal/FutureUtil.java
@@ -9,9 +9,11 @@
  * Contributors:
  *  Ahmed Hussain (Cocotec Ltd) - initial implementation
  *  Sebastian Thomschke (Vegard IT GmbH) - extracted code from LanguageServers class
+ *  Sebastian Thomschke (Vegard IT GmbH) - changed addAll() method to join()
  *******************************************************************************/
 package org.eclipse.lsp4e.internal;
 
+import java.util.ArrayList;
 import java.util.List;
 import java.util.concurrent.CancellationException;
 import java.util.concurrent.CompletableFuture;
@@ -20,25 +22,20 @@ import java.util.function.Function;
 public class FutureUtil {
 
 	/**
-	 * Combines two async lists of results into a single list by adding all the
-	 * elements of the second list to the first one.
+	 * Combines two async lists of results into a new single list containing all
+	 * elements of the first and the second result.
 	 *
-	 * @param <T>
-	 *            Result type
-	 * @param accumulator
-	 *            One async result
-	 * @param another
-	 *            Another async result
 	 * @return Async combined result
 	 */
-	public static <T> CompletableFuture<List<T>> addAll(CompletableFuture<List<T>> accumulator,
-			CompletableFuture<List<T>> another) {
-		CompletableFuture<List<T>> res = accumulator.thenCombine(another, (a, b) -> {
-			a.addAll(b);
-			return a;
+	public static <T> CompletableFuture<List<T>> join(final CompletableFuture<List<T>> firstResult,
+			final CompletableFuture<List<T>> secondResult) {
+		final CompletableFuture<List<T>> result = firstResult.thenCombine(secondResult, (firstList, secondList) -> {
+			final List<T> combinedList = new ArrayList<>(firstList);
+			combinedList.addAll(secondList);
+			return combinedList;
 		});
-		forwardCancellation(res, accumulator, another);
-		return res;
+		forwardCancellation(result, firstResult, secondResult);
+		return result;
 	}
 
 	@SuppressWarnings("null")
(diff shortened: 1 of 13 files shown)


rubenporras:
@sebthom: Regarding the fourth commit on the PR, should not rather use CancellationSupport in order to have one pattern and not two? I think that would beneficial and worthwhile if we are refactoring.

As an example, I have refactored only one method (https://github.com/eclipse/lsp4e/pull/1088). I can do the rest if you agree.

sebthom:
To be honest I don't see the point in the CancellationSupport, it looks like classical over engineering to me. I would even say that the method `execute` as misleadingly named.

rubenporras:
I think it makes sense where there is need to keep track of a list of futures that need to be cancelled together. I agree that whenever there is only ever one that can be cancelled, it is probably overkill.

I have also wondered about the name of `execute`.

I will close my PR and take a look again at your refactor tomorrow.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/internal/FutureUtil.java
```diff
@@ -20,25 +22,20 @@
...
 	 */
-	public static <T> CompletableFuture<List<T>> addAll(CompletableFuture<List<T>> accumulator,
-			CompletableFuture<List<T>> another) {
-		CompletableFuture<List<T>> res = accumulator.thenCombine(another, (a, b) -> {
-			a.addAll(b);
-			return a;
+	public static <T> CompletableFuture<List<T>> join(final CompletableFuture<List<T>> firstResult,
+			final CompletableFuture<List<T>> secondResult) {
+		final CompletableFuture<List<T>> result = firstResult.thenCombine(secondResult, (firstList, secondList) -> {
+			final List<T> combinedList = new ArrayList<>(firstList);
```
rubenporras, line 50: I am unsure about this. The new implementation is safer, but it is also going to make unnecessary copies in many cases.

Could we not check before creating a new list if the old one is modifiable?

I know it is a bit ugly but it is probably worth to make a copy only if the list is a Guava ImmutableList or an instance of java.util.ImmutableCollections?

Otherwise I think I would be happier assuming the lists used for the join are not immutable.

rubenporras, line 50: should we not create the list with the size firstList.size() + 1 to avoid a potential copy of the internal array?

org.eclipse.lsp4e/src/org/eclipse/lsp4e/internal/FutureUtil.java
```diff
@@ -17,10 +17,27 @@
...
 import java.util.concurrent.CancellationException;
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.Future;
 import java.util.function.Function;
 
+import org.eclipse.jdt.annotation.Nullable;
+
 public class FutureUtil {
 
+	public static void cancel(final @Nullable Future<?> futureToCancel) {
```
rubenporras, line 27: add javadoc for both methods?

org.eclipse.lsp4e/src/org/eclipse/lsp4e/internal/FutureUtil.java
```diff
@@ -0,0 +1,66 @@
...
+ *  Sebastian Thomschke (Vegard IT GmbH) - extracted code from LanguageServers class
+ *******************************************************************************/
+package org.eclipse.lsp4e.internal;
+
+import java.util.List;
+import java.util.concurrent.CancellationException;
+import java.util.concurrent.CompletableFuture;
+import java.util.function.Function;
+
+public class FutureUtil {
```
rubenporras, line 25: add javadoc?

rubenporras (commented):
Thanks for splitting the PR in several commit. It makes the review much easier.

rubenporras:
> It also renames the `addAll()` function to `join()` and changes the implementation in the way that not the result list of the first future is extended with entries of a secondary future (which makes the unsafe assumption the first result list is modifiable) but adds all items into a new list.

Maybe we should not rename it? I can see how by renaming one could think that the method makes the unsafe assumption the first result list is modifiable, but if the name is as of now `addAll()` one could argue that that is expected and logical.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/internal/FutureUtil.java
sebthom: I think it is any case not good practice to modify the list in-place that is held by a future. If the first future is for example accessed through another code path too it suddenly contains elements from unrelated futures. I think the way `addAll` is implemented it is so specific and with side effects that it should not be exposed at all and only privately declared and inside where needed.

PR #1092, opened 2024-08-29
Fix disappearing sort button in outline view (fix #1091)

### LINKED ISSUE #1091: Sort toolbar button disappears in outline view if an LSP editor is activated
The outline view provides a sort button. It is available as long as the outline view is active. As soon as instead an LSP editor becomes active, the button disappears, although the LSP editor provides the outline page that is still visible.

Examples:

Outline view is active, sort button is visible in the toolbar:
image

LSP editor is active, the sort button disappeared:
image

I found this issue while working on PR #1075.

travkin79:
This PR should be merged before merging PR #1075.

### DIFF
--- a/org.eclipse.lsp4e/plugin.xml
+++ b/org.eclipse.lsp4e/plugin.xml
@@ -338,7 +338,7 @@
       <menuContribution allPopups="false" locationURI="toolbar:org.eclipse.ui.views.ContentOutline?after=additions">
          <command commandId="org.eclipse.lsp4e.toggleSortOutline" style="toggle" label="%command.toggle.outline.sort.label">
             <visibleWhen>
-               <reference definitionId="org.eclipse.lsp4e.activePartHasCNFOutlinePage" />
+               <reference definitionId="org.eclipse.lsp4e.activeEditorHasCNFOutlinePage" />
             </visibleWhen>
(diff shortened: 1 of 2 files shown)


rubenporras:
Could you explain why the fix works in the commit (and PR) message? I am not very skilled on this area of Eclipse, and just from looking at the PR, I cannot tell why it was broken before and why it is fixed now.

travkin79:
Yes. Previously, the sort button was only active if the active part was the outline view. As soon as an editor became active, the `visibleWhen` expression became `false` and the sort button disappeared. The check was done by the expression definition named `activePartHasCNFOutlinePage`.

I added a new expression definition named `activeEditorHasCNFOutlinePage`. I changed the `visibleWhen` expression to only check if the (last) active editor (that does not necessarily has to be in focus) has an outline page of type `CNFOutlinePage`. The remaining checks using the expression `activePartHasCNFOutlinePage` are left unchanged.

rubenporras:
Thanks for the explanation. I understand now. Can you amend the commit message with the explanation? then I would be happy to merge.

rubenporras (approved):
Thanks

travkin79:
Thank you, @rubenporras . 😄

Issue #628, opened 2023-05-04
Using LanguageServerWrapper#getServerCapabilities() can be dangerous and provides a lot of freeze


angelozerr:
I create this issue because I'm porting LSP4E in IJ and the call of this method can freeze the IJ and I have an idea how to avoid doing that but I would like to know your feedback.

The main problem when you call LanguageServerWrapper#getServerCapabilities() is when:

 * you are closing the last file mapped with the language server which reports diagnostics and a component is calling LanguageServerWrapper#getServerCapabilities(). We have this usecase in IJ because we need to check the code action capability when we reports diagnostics. If we close  IJ and there are diagnostics which are reported when language server is stopping, we call LanguageServerWrapper#getServerCapabilities()  which restarts the language server which is bad. Perhaps in LSP4E there is this kind of usecase, I don't know.
 * the language server cannot be started and in this case there are a lot of freeze (because we wait for 10 seconds) but more we could wait for 2 * 10 seconds. For instance this code https://github.com/eclipse/lsp4e/blob/71fdc020de59658be6a29205f24d4f2c56ea2eaa/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServiceAccessor.java#L300 will wait for 20 seconds because we call twice the getServerCapabilities.

My idea is to cache the serverCapabilities instance in a LanguageServerDefinition. The idea is to wait for 10 seconds just one time (if the language server cannot be started for instance) and after that we use the cached serverCapabilities. It will give the capability to check server capability even if the language server is not started the second time.

What do you think about this idea?

mickaelistria:
I don't think that capabilities should belong to the server definition, they're a per-instance piece of data.
Before changling LSP4E, can you please provide steps to reproduce this issue? The fact that it fails in IJ isn't per se a strong enough reason to change LSP4E.

angelozerr:
> I don't think that capabilities should belong to the server definition, they're a per-instance piece of data.

Why? All language server wrapper instance will return the same server capabilities for a given server definition, no?

> Before changling LSP4E, can you please provide steps to reproduce this issue? The fact that it fails in IJ isn't per se a strong enough reason to change LSP4E.

It is very difficult to provide a usecase which have some freeze, because getServerCapabilities can be consume in a lot of context.

mickaelistria:
> All language server wrapper instance will return the same server capabilities for a given server definition, no?

[...]

PR #1098, opened 2024-09-04
feat: introduce LanguageServerWrapper#getServerCapabilitiesAsync

sebthom:
`getServerCapabilities()` can block for up to 10 seconds. In code locations where the result is a future anyways it is useful to have a `getServerCapabilitiesAsync()`

I am planning to use `getServerCapabilitiesAsync()` more in future PRs where it will play nicely with the then suggested changes.

Addresses https://github.com/eclipse/lsp4e/issues/628

### DIFF
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/inlayhint/LSPLineContentCodeMining.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/inlayhint/LSPLineContentCodeMining.java
@@ -92,7 +92,13 @@ public class LSPLineContentCodeMining extends LineContentCodeMining {
 
 	@Override
 	protected CompletableFuture<@Nullable Void> doResolve(ITextViewer viewer, IProgressMonitor monitor) {
-		if (wrapper.isActive() && canResolveInlayHint(wrapper.getServerCapabilities())) {
+		if (!wrapper.isActive()) // TODO is this check required? if so is it missing in LSPCodeMining.doResolve()?
+			return CompletableFuture.completedFuture(null);
+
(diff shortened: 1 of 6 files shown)


org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/inlayhint/LSPLineContentCodeMining.java
```diff
@@ -92,16 +92,18 @@ public void setLabel(final @Nullable String label) {
 
 	@Override
 	protected CompletableFuture<@Nullable Void> doResolve(ITextViewer viewer, IProgressMonitor monitor) {
-		if (wrapper.isActive() && canResolveInlayHint(wrapper.getServerCapabilities())) {
```
rubenporras, line 95: why do we remove the `wrapper.isActive()` condition?

rubenporras:
Other than my comment, the PR is a very nice idea.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/inlayhint/LSPLineContentCodeMining.java
sebthom: for me the question why is it there. this pattern is not used consistently. in `LSPCodeMining.doResolve` for example this check is not performed. so either it's obsolete or missing in the other location

sebthom: Ok I re-added the check and added a comment about it.

rubenporras (approved):
thanks

PR #1105, opened 2024-09-11
F3 jumps between definitions and declarations - Adapt order of definitions and declarations (fix #1104)

### LINKED ISSUE #1104: Use F3 to switch between "Open Definition" and "Open Declaration", e.g. for switching between header and cpp files in C++
I'd like to slightly change how LSP4E collects and uses definitions, declarations, implementations and type definitions for a cursor position or for a selection in an LSP-based text editor. My goal is the following: For C++ editors (in CDT LSP), by pressing F3 a user should jump to the definition or to the declaration of the element under cursor and by pressing F3 multiple times just jump from a method's declaration (usually in a header file) to its definition (usually in a cpp file) and vice versa. In general, F3 should not just pick the first element from the list of all definitions, declarations, type definitions, and implementations, but at least select one of these that the cursor is not already placed upon.

At the moment, pressing F3 always opens the first definition (or declaration or type definition or implementation) found, even if the cursor is already placed on that definition.

My proposal is to detect the element that the cursor is placed upon and place that element (a definition or declaration) to the end of the list of all definitions and declarations, then followed by the type definitions and implementations.

I.e., the only thing that would change is the order of the definitions, declarations, type definitions and implementations in the list that is used to pick the element to jump to (the first element from that list is used as target). The same list is used when using Ctrl + click to select an element to jump to. Since we wouldn't remove any element from that list, that change should work for both use cases.

This issue might be related to issue #332.

travkin79:


mickaelistria:
As mentioned on the issue, I have some concerns that this might not be the expected or desired behavior for most users.

akurtakov:
What is the status here?

travkin79:
Thanks for asking, @akurtakov,
I'd like to have (I'm waiting for) an answer to my last solution proposals from [this comment](https://github.com/eclipse-lsp4e/lsp4e/issues/1104#issuecomment-2343963424).
I would really like to find a solution that solves the issue without compromising the existing concepts or conventions.

travkin79:
FYI: @akurtakov and @mickaelistria,
It seems, the issue is already solved. We wanted F3 to jump between declaration and definition in CDT LSP (which uses LSP4E). Now, in the latest CDT LSP version, F3 seems to behave exactly as we want it to. Thus, I'll consider this issue as solved.

PR #1107, opened 2024-09-17
Do not crash on re-registration

mickaelistria:
Some language servers such as OmniSharp may (erroneously but not fatally) re-register the same capability multiple times. Instead of crashing in such case, just log.

### DIFF
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServerWrapper.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServerWrapper.java
@@ -59,6 +59,7 @@ import org.eclipse.core.resources.WorkspaceJob;
 import org.eclipse.core.runtime.Adapters;
 import org.eclipse.core.runtime.Assert;
 import org.eclipse.core.runtime.CoreException;
+import org.eclipse.core.runtime.ILog;
 import org.eclipse.core.runtime.IPath;
 import org.eclipse.core.runtime.IProduct;
 import org.eclipse.core.runtime.IProgressMonitor;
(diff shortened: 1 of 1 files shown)


mickaelistria:
@akurtakov Can you confirm this would help aCute?

akurtakov:
I know see:
```
!ENTRY org.eclipse.lsp4e 2 0 2024-09-17 14:06:39.120
!MESSAGE A registration with id 49a06df1-0048-4f91-9cb3-e55ca8781e46 already exists. Unregistering may not fully work in this case.


!ENTRY org.eclipse.lsp4e 2 0 2024-09-17 14:06:39.121
!MESSAGE A registration with id f113b054-d549-442a-af7d-2faceaf5831e already exists. Unregistering may not fully work in this case.


!ENTRY org.eclipse.lsp4e 2 0 2024-09-17 14:06:39.122
!MESSAGE A registration with id 2c1bfa22-1b23-4382-8d3e-e1216cc0dbc3 already exists. Unregistering may not fully work in this case.


!ENTRY org.eclipse.lsp4e 2 0 2024-09-17 14:06:39.123
!MESSAGE A registration with id b4a7234e-b299-491a-b349-b827c4bc9b0b already exists. Unregistering may not fully work in this case.
```
Unfortunately, Omnisharp is not working with it too. If you believe it's the proper thing to do please merge. I'll have to fight more with Omnisharp to figure what is going on.

angelozerr:
I had a similar problem with LSP4IJ and you need to store in a map the registration id because I had a case with a ls which registers 2 fileWatchers and unregister the last.

In this case you need to keep the first and remove the second.

mickaelistria:
Failures are not related, let's merge.

PR #1116, opened 2024-09-20
Render markdown with commonmark instead of wikitext

mickaelistria:
commonmark is the library used internally by the JDK and by JDT for markdown rendering. It's more standard and leaner than wikitext.

### DIFF
--- a/target-platforms/target-platform-latest/target-platform-latest.target
+++ b/target-platforms/target-platform-latest/target-platform-latest.target
@@ -3,12 +3,6 @@
 <target name="LSP4E Target-platform" sequenceNumber="1473749310">
 	<locations>
 		<location includeAllPlatforms="false" includeConfigurePhase="true" includeMode="planner" includeSource="true" type="InstallableUnit">
-			<unit id="org.eclipse.mylyn.wikitext" version="0.0.0"/>
-			<unit id="org.eclipse.mylyn.wikitext.markdown" version="0.0.0"/>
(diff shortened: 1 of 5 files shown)


target-platforms/target-platform-latest/target-platform-latest.target
```diff
@@ -21,6 +15,11 @@
 			<unit id="org.eclipse.sdk.ide" version="0.0.0"/>
 			<repository location="https://download.eclipse.org/eclipse/updates/latest/"/>
 		</location>
+		<location type="InstallableUnit" includeConfigurePhase="true" includeMode="planner" includeSource="true">
+			<repository location="https://download.eclipse.org/eclipse/updates/4.33-P-builds/"/>
```
akurtakov, line 19: Please don't use 4.33-Pbuilds as they will be gone soon. Rather add the bundles to the orbit location with latest https://download.eclipse.org/tools/orbit/simrel/orbit-aggregation/milestone/S202409180633 as used in https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/master/eclipse.platform.releng.prereqs.sdk/eclipse-sdk-prereqs.target

mickaelistria: OK, done.

mickaelistria:
I'm going to merge it right now as we're in the beginning of a new stream for the IDE, so that leaves some time to find and fix issues if necessary. Removing the dependency on wikitext and all in favor of the standard-ish commonmark is quite likely to bring many small benefits and doesn't seem to have any obvious pitfall at the moment.

PR #1124, opened 2024-09-27
set sort label for quick access elements to avoid symbols get removed when they have the same label

martinlippert:
When you have multiple workspace symbols with the same label, only one shows up in the workspace quick access list. The reason is that the filtering and sorting mechanism of the quick access dialog uses a comparator for a set that uses the `getSortLabel` method of the elements - which refers to `getLabel` by default.

This PR overrides the default implementation and makes sure (by adding the random extension ID to make symbols unique) that  `getSortLabel` returns different values for different symbols, even if they have the exact same label.

As a follow-up here, we should think about showing more information for a symbol on the dialog to allow users to distinguish those symbols with the same label somehow, but that is a task for a separate PR or issue.


mickaelistria (approved):
All good to me. Feel free to merge as soon as Jenkins build completes successfully.

rubenporras (approved):

martinlippert:
Thanks @mickaelistria and @rubenporras for taking a look, much appreciated. Merged now.

Issue #1103, opened 2024-09-10
OOM and Process Hang on Windows 10 in LanguageServerWrapperTest


sebthom:
As of PR https://github.com/eclipse/lsp4e/pull/1044 (@ava-fred) running integration tests reproducible results in a OOMs and process hangs on at least Windows 10. I tried with different JDKs - same effect. Reverting the commit from https://github.com/eclipse/lsp4e/pull/1044 solves the issue.

The issue appears when the LanguageServerWrapperTest.testStopAndActivate() is executed.

Running `mvn verify -Dtest=LanguageServerWrapperTest#testStopAndActivate` immediately results in OOM within a few seconds and process hangs.

I am attaching a thread dump where one can see that thousands of threads are created.
[threaddump-1725956087712.txt](https://github.com/user-attachments/files/16941993/threaddump-1725956087712.txt)

Because of the LSPs async natures I am having difficulties tracking down the root cause.

Can anyone else reproduce that issue? @rubenporras @mickaelistria

Even though this only appears in that one integration test so prominently it probably has an underlying cause that may affect end-users during normal usage.

ava-fred:
I finally managed to reproduce the OOM exception on my Windows 10 machine, but I had to change the test to make it run longer. I guess my Windows 10 machine is much better at counting to 10_000_000 than yours :->.

As you mentioned, it is hard to track what is going on because of the async nature of things, but I'll see what I can find.

sebthom:
I switched to a Ryzen CPU some time ago. Maybe they behave somewhat different making the underlying issue more prominently visible.

rubenporras:
Hi @sebthom , thanks for looking at this. Since @ava-fred is looking into it, I will not look into this problem myself (Note: in case you did not notice from our activity in the project, we both work for the same company).

Regards

sebthom:
This issue now also affects most of the Jenkins builds see https://ci.eclipse.org/lsp4e/job/lsp4e-github/job/main/75/console

ava-fred:
My analysis shows that LanguageServerWrapperTest.testStopAndActivate() leaks threads executing ConcurrentMessageProcessor.run() because the thread started in MockConnectionProviderMultiRootFolders.start() is never shut down. This was so before #1044, the only thing that apparently changed with this PR is that the balance between the 2 loops in the test has changed slightly so that the thread leak now causes actual problems.  
[...]

PR #1126, opened 2024-10-01
fix: improve startup and shutdown

ava-fred:
In #1103, it was noted that
LanguageServerWrapperTest#testStopAndActivate causes OOM errors. On investigation, it was discovered that there are multiple causes for this.

1) The MockConnectionProviderMultiRootFolder used in the test created a message processor thread in LSP4J on each call to start() but did not shut them down.
2) The termination of the start/stop loop in the test was wrong: the loop ran for as long as the VM running tests was alive. 3) The "already stopping" logic in LanguageServerWrapper#shutdown was wrong: while the shutdown of a LanguageServerWrapper was processed, further calls to shutdown were ignored.
4) There was a race between the initializationFuture of the LanguageServerWrapper and the future used for shutdown.

With this commit, the test is rewritten to avoid crashes and to properly test that calling stop / start repeatedly on LanguageServerWrapper does not leave any connection providers running. The LanguageServerWrapper class is refactored to make the test pass.

### DIFF
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServerWrapper.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServerWrapper.java
@@ -39,7 +39,6 @@ import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
-import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.Consumer;
 import java.util.function.Function;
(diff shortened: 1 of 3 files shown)


org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServerWrapper.java
```diff
@@ -152,27 +151,33 @@ public void dirtyStateChanged(IFileBuffer buffer, boolean isDirty) {
 
 	};
 
+	private static class Context {
+		boolean cancelled= false;
```
rubenporras, line 155:
```suggestion
		boolean cancelled = false;
```

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServerWrapper.java
```diff
@@ -286,79 +291,88 @@ private synchronized void start(boolean forceRestart) {
...
-			.thenAccept(res -> {
-				advanceInitializeFutureMonitor();
-				serverCapabilities = res.getCapabilities();
-				this.initiallySupportsWorkspaceFolders = supportsWorkspaceFolders(serverCapabilities);
+
+					UnaryOperator<MessageConsumer> wrapper = consumer -> (message -> {
+						logMessage(message);
+						consumer.consume(message);
+						final var lspStreamProvider = workingContext.lspStreamProvider;
+						if (lspStreamProvider != null && isActive() && workingContext.languageServer != null) {
```
rubenporras, line 331: extract `workingContext.languageServer` also a local final?

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServerWrapper.java
```diff
@@ -286,79 +291,88 @@ private synchronized void start(boolean forceRestart) {
...
-			.thenCompose(unused -> {
-				advanceInitializeFutureMonitor();
-				return initServer(rootURI);
-			})
-			.thenAccept(res -> {
-				advanceInitializeFutureMonitor();
-				serverCapabilities = res.getCapabilities();
-				this.initiallySupportsWorkspaceFolders = supportsWorkspaceFolders(serverCapabilities);
+
+					UnaryOperator<MessageConsumer> wrapper = consumer -> (message -> {
```
rubenporras, line 327:
```suggestion
					UnaryOperator<MessageConsumer> wrapper = consumer -> message -> {
```

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServerWrapper.java
```diff
@@ -286,79 +291,88 @@ private synchronized void start(boolean forceRestart) {
...
+						logMessage(message);
+						consumer.consume(message);
+						final var lspStreamProvider = workingContext.lspStreamProvider;
+						if (lspStreamProvider != null && isActive() && workingContext.languageServer != null) {
+							lspStreamProvider.handleMessage(message, workingContext.languageServer, rootURI);
+						}
+					});
```
rubenporras, line 334:
```suggestion
					};
```

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServerWrapper.java
```diff
@@ -286,79 +291,88 @@ private synchronized void start(boolean forceRestart) {
...
+					initParams.setWorkspaceFolders(getRelevantWorkspaceFolders());
+					Launcher<LanguageServer> launcher = serverDefinition.createLauncherBuilder() //
+							.setLocalService(languageClient)//
+							.setRemoteInterface(serverDefinition.getServerInterface())//
+							.setInput(castNonNull(workingContext.lspStreamProvider).getInputStream())//
```
rubenporras, line 339: Extract `castNonNull(workingContext.lspStreamProvider)` to a local field?

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServerWrapper.java
```diff
@@ -553,46 +575,39 @@ private void shutdown() {
...
+						LanguageServerPlugin.logError(ex.getClass().getSimpleName() + " occurred during shutdown of " //$NON-NLS-1$
+								+ workingContext.languageServer, ex);
+					}
 				}
-			}
 
-			if (serverFuture != null) {
-				serverFuture.cancel(true);
-			}
+				if (workingContext.launcherFuture != null) {
```
rubenporras, line 594: Now that we have an object that represents a working context, could we maybe refactor this bit by adding a method `close` (or similar name) and move the `cancel` of the future, the `exit` on the server and the `stop` of the streamProvider into it?

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServerWrapper.java
```diff
@@ -152,27 +151,33 @@ public void dirtyStateChanged(IFileBuffer buffer, boolean isDirty) {
 
 	};
 
+	private static class Context {
```
rubenporras, line 154: Could `LanguageServerContext` be a better name?

If I am not mistaken, all the information in the class refers to the LanguageServer and the related communication channels.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServerWrapper.java
ava-fred: done

rubenporras (approved):

Issue #1127, opened 2024-10-01
Formatting selected code doesn't work anymore, the whole file is formatted instead


travkin79:
**Issue**
If you select some text in an LSP-based text editor and then you use the shortcut for formatting code (Ctrl + Shift + F on PCs), then the whole file is formatted instead of formatting the selected text only.

travkin79:
The following commit seems to have introduced that behavior:
https://github.com/eclipse/lsp4e/commit/cc4b1ba047709a9186393ff15a8cae314aa8ba6f

See https://github.com/eclipse/lsp4e/blob/cc4b1ba047709a9186393ff15a8cae314aa8ba6f/org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/format/LSPFormatter.java#L59-L63

According to this code, the range parameters (the given text selection) are only used if the LS does not support document formatting and if the text selection has length 0. That does not make sense to me.

sebthom:
Fixed in LSP4E 0.26.6

PR #1128, opened 2024-10-01
Fix selective formatting (issue #1127)

travkin79:
This PR fixes selective formatting.

### DIFF
--- a/org.eclipse.lsp4e.test/META-INF/MANIFEST.MF
+++ b/org.eclipse.lsp4e.test/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: Tests for language server bundle (Incubation)
 Bundle-SymbolicName: org.eclipse.lsp4e.test;singleton:=true
-Bundle-Version: 0.15.18.qualifier
+Bundle-Version: 0.15.19.qualifier
 Fragment-Host: org.eclipse.lsp4e
 Bundle-Vendor: Eclipse.org
 Bundle-RequiredExecutionEnvironment: JavaSE-17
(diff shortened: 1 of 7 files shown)


travkin79:
I think, my fix is ready, but I'd like to add an automated test for that case, too.

rubenporras:
Thanks for the fix. It looks good to me but I will wait for the test case before merging.

travkin79:
I think, now, the PR is ready to be merged. But please take a look at the following line. Is that the right thing to be returned in this case?

https://github.com/travkin79/lsp4e/blob/15350a9bff093771bc306cf06ffea2f25d9106c7/org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/format/LSPFormatter.java#L66

travkin79:
Hi @rubenporras,
Since this bugfix is important and urgent for my team, when do you plan to make the next LSP4E release?

rubenporras:
> I think, now, the PR is ready to be merged. But please take a look at the following line. Is that the right thing to be returned in this case?
> 
> https://github.com/travkin79/lsp4e/blob/15350a9bff093771bc306cf06ffea2f25d9106c7/org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/format/LSPFormatter.java#L66

I am also not sure and you implementation seems sensible.

In any case, I think that if the client pays attention to the capabilities of the client, we should never reach this code.

rubenporras:
> Hi @rubenporras, Since this bugfix is important and urgent for my team, when do you plan to make the next LSP4E release?

In that case, I will prepare a new release today or tomorrow.

rubenporras:
Hi @travkin79 ,

you have a new release ;)

Cheers.

travkin79:
> you have a new release ;)
Hi @rubenporras,
That's awesome! Thank you very much. :-)

PR #1143, opened 2024-11-15
Set the vendor of all plugins to Eclipse LSP4E

FlorianKroiss:
The vendor for org.eclipse.lsp4e.debug was missing. Some of the other plugins used Eclipse.org as vendor.

### DIFF
--- a/org.eclipse.lsp4e.debug/META-INF/MANIFEST.MF
+++ b/org.eclipse.lsp4e.debug/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: Debug Adapter client for Eclipse IDE (Incubation)
 Bundle-SymbolicName: org.eclipse.lsp4e.debug;singleton:=true
-Bundle-Vendor: Eclipse.org
+Bundle-Vendor: Eclipse LSP4E
 Bundle-Version: 0.15.12.qualifier
 Bundle-Activator: org.eclipse.lsp4e.debug.DSPPlugin
 Require-Bundle: org.eclipse.ui,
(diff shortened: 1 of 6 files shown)


FlorianKroiss:
I noticed this while going through the list of installed plugins in the About dialog

org.eclipse.lsp4e.debug/META-INF/MANIFEST.MF
```diff
@@ -2,7 +2,8 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: Debug Adapter client for Eclipse IDE (Incubation)
 Bundle-SymbolicName: org.eclipse.lsp4e.debug;singleton:=true
-Bundle-Version: 0.15.11.qualifier
+Bundle-Vendor: Eclipse.org
```
mickaelistria, line 5: https://www.eclipse.org/projects/handbook/#code states that

> Provider information is set to the project’s formal name: e.g. the Bundle-Vendor entry set to "Eclipse Foo" in OSGi Bundles;

So it should be
```suggestion
Bundle-Vendor: Eclipse LSP4E
```

mickaelistria (commented):
See comment inline

mickaelistria (changes requested):
See comment inline

FlorianKroiss:
I guess in that case, some of the other plugins also need to be changed to "Eclipse LSP4E"?
image
If so, I can fix that too

mickaelistria:
> If so, I can fix that too

that would be welcome!

FlorianKroiss:
Ok, I will also fix the other plugins.

Maybe on an unrelated note: The rule mentioned in the handbook does not seem to be widely followed?
image

mickaelistria:
> Ok, I will also fix the other plugins.

Thanks.

> The rule mentioned in the handbook does not seem to be widely followed?

IIRC the recommendation used to be "Eclipse.org" something like ~12 years ago. And many projects didn't change, and new projects took those older projects as example repeating the same anti-pattern.
But it's not a big deal as long as no-one complains.

mickaelistria (approved):

mickaelistria:
Thank you!

PR #1148, opened 2024-11-21
fix: merge instead of replacing style from semantic highlighting

gregdyke:
Semantic highlighting might be applied on top of other highlighting (e.g. diff highlighting in a compare view editor) -> merge instead of replacing


mickaelistria:
Thank you!

PR #1153, opened 2024-12-04
Allow semantic highlighting to unset bold and italic

gregdyke:
Semantic highlighting needs to be able to override the styles set by TM4E. Simply not setting the styles in TM4E would introduce flickering in the delay waiting for semantic highlighting.

If unsetting of bold and/or italic is not desired, disable using semanticHighlightReconciler.ignoreBoldNormal and
semanticHighlightReconciler.ignoreItalicNormal

----

When applying semantic highlighting, use
TextPresentation#mergeStyleRanges to create new style ranges and merge as "usual". Then traverse resulting style ranges and unset bold/italic where it is not set in semantic highlighting.

The double traversal algorithm is simplified because after calling TextPresentation#mergeStyleRanges we are guaranteed that each style range from semantic highlighting has exact overlap with 1 or more style ranges in the textPresentation.

----

Because textmate uses css, style {font-weight:"normal"} should be expected to modify the font-weight, whereas {font-weight:"inherit"} should not. However, TM4E parses the style into a an SWT StyleRange which prevents the differentiation between "normal"=unset bold if it is set and "inherit"=don't unset bold if it is set.

A more long term solution would use the css to determine whether an unset bold bit is overriding or not.

For the medium term we simply say "semantic highlighting has ownership of bold/italic and the style sheet must set them explicitly"

This addresses the issue discussed in https://github.com/eclipse-platform/eclipse.platform.ui/issues/2543


mickaelistria (approved):

rubenporras:
@mickaelistria , do you know what is the problem with the License Check? It looks like it is completely broken.

mickaelistria:
ClearlyDefined is fragile, it's probably a connection issue.

mickaelistria:
license check is OK now.

rubenporras:
@akurtakov , @mickaelistria :

The test org.eclipse.lsp4e.test.hover.HoverTest.testIntroUrlLink() is failing due to I believe Eclipse 2024-12. There is an error message "!MESSAGE Failed to execute IntroURL: http://org.eclipse.ui.intro/execute?command=org.eclipse.ui.file.close
"

Do you know what is the URL now of that page? I could not find it.

akurtakov:
Sorry I have no idea nor I can look into it now.

rubenporras:
Thanks for the reply, I have marked the test as ignore for the moment.

rubenporras:
There is one test that is failing on Mac OS unrelated to this PR, so I am merging it.

mickaelistria:
> Do you know what is the URL now of that page?

The `http://org.eclipse.ui.intro` is a special interpretation by the eclipse intro framework to process the following parts of the URL and convert them into an action invocation. This was one way to allow triggering IDE actions from the LSP, I think before `commands` were properly specified in LSP.
I don't mind if we ignore it at the moment, I don't expect any existing serious LS to actually leverage this Eclipse-specific feature.

PR #1155, opened 2024-12-05
Remove assertion that makes the test fail

rubenporras:


rubenporras:
I think the test starts failing after Eclipse 2024-12 has been released. I do not see any harm in removing the assertion that you introduced. Do you?

rubenporras:
@sebthom, after this commit the test is passing on windows but not on Mac OS. I do not have access to any Mac OS where I could test locally. Since you have worked on this test before (and added the assertion), do you think you can take a look why the assertion started failing after the upgrade from Eclipse 2024-09 to Eclipse 2024-12?

PR #1156, opened 2024-12-05
Mark test HoverTest as ignored until it can be fixed

rubenporras:


rubenporras:
@BoykoAlex , are you able to take a look why the test I have marked as ignore fails with Eclipse 2024-12 but did not fail before?

BoykoAlex:
@rubenporras active part at the time of the command execution is not the editor for some reason. I just made the active part the editor via #1160

PR #1160, opened 2024-12-10
Fix failing in Eclipse 4.34 hover test

BoykoAlex:
The ignored hover has an issue that active part becomes package explorer all of a sudden when hover shell is brought up.
For the test to pass I simply decided to activate the expected editor part manually. After all the test is to verify that intro URL's clicked on the hover are working.


rubenporras (commented):
Thanks for looking at it. 

I think you need to update the version of the test plug-in.

BoykoAlex:
@rubenporras yep, just did that :-)

PR #1163, opened 2024-12-19
fix: GHA builds on MacOS fail

sebthom:


rubenporras:
Thanks for looking into it.

PR #1170, opened 2025-01-06
feat: cancel willSaveWaitUntil if the result will no longer be used

rubenporras:


org.eclipse.lsp4e/src/org/eclipse/lsp4e/DocumentContentSynchronizer.java
```diff
@@ -259,18 +259,13 @@ public void documentAboutToBeSaved() {
 
 		// Use @link{TextDocumentSaveReason.Manual} as the platform does not give enough information to be accurate
 		final var params = new WillSaveTextDocumentParams(identifier, TextDocumentSaveReason.Manual);
-
+		CompletableFuture<@Nullable List<TextEdit>> edits = languageServerWrapper.executeImpl(ls -> ls.getTextDocumentService().willSaveWaitUntil(params));
```
mickaelistria, line 262: I don't remember whether the return value of "executeImpl" is actually cancellable.
The best way to check that is to enable the log, get into a case where cancellation happens, and then check is the `$cancel` request was send from LSP4E to the LS. Did you verify it that way?
IIRC there is also a test case somewhere about cancellation that intercept those cancellation request between client and LS to verify them.

rubenporras: I have not though about that. Thanks for the hint. I will check.

PR #1171, opened 2025-01-11
feat: support auto folding of comments and add folding preferences page

sebthom:
This PR implements auto folding for comments and regions and adds this new settings page:
image

This is a follow-up PR to https://github.com/eclipse-lsp4e/lsp4e/pull/971 and solves https://github.com/eclipse-lsp4e/lsp4e/issues/895 and https://github.com/eclipse-lsp4e/lsp4e/issues/927


org.eclipse.lsp4e/src/org/eclipse/lsp4e/ui/FoldingPreferencePage.java
```diff
@@ -0,0 +1,82 @@
...
+import org.eclipse.ui.IWorkbenchPreferencePage;
+
+public class FoldingPreferencePage extends FieldEditorPreferencePage implements IWorkbenchPreferencePage {
+
+	public static final String PREF_FOLDING_ENABLED = "foldingReconcilingStrategy.enabled"; //$NON-NLS-1$
+	public static final String PREF_AUTOFOLD_COMMENTS = "foldingReconcilingStrategy.collapseComments"; //$NON-NLS-1$
+	public static final String PREF_AUTOFOLD_REGIONS = "foldingReconcilingStrategy.collapseRegions"; //$NON-NLS-1$
+	public static final String PREF_AUTOFOLD_IMPORT_STATEMENTS = "foldingReconcilingStrategy.collapseImports"; //$NON-NLS-1$
+
+	public static final class PreferenceInitializer extends AbstractPreferenceInitializer {
```
rubenporras, line 32: Is initializing the preferences to some value the preferred pattern? I think org.eclipse.lsp4e.ui.LoggingPreferencePage does not initialize its preferences, rather it relies on the defaults of IPreferenceStore.getBoolean. I think this is better because one can change the defaults in LSP4E any time and they will be changed for users as well unless they have set the values themselves. Do you agree with that or do you think it is better to initialize the values? Why then?

sebthom: I don't know what the preferred way is TBH. I used it because one pref key needs a default value of true and with the initializer I have a single place to define it.

rubenporras:
@mickaelistria , do you know if there is a preferred way? (see my comment above)

mickaelistria:
One can set preference in a preference initializer, that's the right pattern.
RCP providers can still override it in their plugin_customization.ini files, this one will take priority over the PreferenceInitializer.

rubenporras (approved):
Thanks

ghentschke:
Thanks for implementing!

Issue #907, opened 2024-02-05
slow IO access in Ui Thread


jukzi:
during sampling full build of eclipse platform workspace i saw this:
![image](https://github.com/eclipse/lsp4e/assets/51790620/1cf69e77-6444-4306-9cb3-a3fea9fd01df)
callstack:
![image](https://github.com/eclipse/lsp4e/assets/51790620/2e4fdffc-6665-4535-b76d-4f3121854c5e)
![image](https://github.com/eclipse/lsp4e/assets/51790620/dd0bac48-7202-4f7c-8797-d1277b3f2df5)

at best UI thread would not  perform I/O at all, but at east it should not call super slow java.io.File.getCanonicalPath ()

jukzi:
![image](https://github.com/eclipse/lsp4e/assets/51790620/e9d45b1e-3cde-4497-8420-368a65f07c60)

mickaelistria:
findResourceFor is a basic workspace query, that doesn't deal with IO, so I've retitled the issue. Moreover, this particular method cannot be optimized as about all of its execution time is spent in Workspace.findFilesForLocationURI. 
This stack most likely come from refreshing the Outline view; the resource is necessary in order to compute markers for the given symbol. However, the `findResourceFor` operation can probably be cached somehow in the label provider: IIRC, it's a stateful class that lives as long as LSP4E provides an outline, so caching the URI->resolution result in it might work and save most calls to `findResourceFor`.

jcompagner:
i have the same thing, i just need to have the outline view closed because with it it is very hard to work inside editors..

ghentschke:
The next step would be to refactor the `org.eclipse.lsp4e.outline.LSSymbolsContentProvider` to support partial results as described [here](https://github.com/eclipse-lsp4j/lsp4j/issues/815#issuecomment-1975121777)
and use something like `org.eclipse.jface.databinding.viewers.ObservableListTreeContentProvider<E>` to support lazy loading.

sebthom:
Caching has been implemented.

PR #1172, opened 2025-01-12
perf: add findResourceFor(URI) cache to SymbolsLabelProvider #907

sebthom:
This caches the repeated findResourceFor(URI) calls in SymbolsLabelProvider.getImage() that can result in performance issues as described in #907

### DIFF
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/outline/SymbolsLabelProvider.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/outline/SymbolsLabelProvider.java
@@ -11,6 +11,7 @@
  *******************************************************************************/
 package org.eclipse.lsp4e.outline;
 
+import static org.eclipse.lsp4e.LSPEclipseUtils.findResourceFor;
 import static org.eclipse.lsp4e.internal.NullSafetyHelper.castNullable;
 
 import java.net.URI;
@@ -91,6 +92,7 @@ public class SymbolsLabelProvider extends LabelProvider
 	 * value: array of images decorated with marker for severity (index + 1)
 	 */
 	private final Map<Image, Image[]> overlays = new HashMap<>();
+	private final Map<Object /*URI|String*/, IResource> resourceCache = new HashMap<>();
 
 	private final boolean showLocation;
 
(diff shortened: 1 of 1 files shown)


org.eclipse.lsp4e/src/org/eclipse/lsp4e/outline/SymbolsLabelProvider.java
```diff
@@ -64,6 +66,9 @@
 import org.eclipse.ui.navigator.ICommonLabelProvider;
 import org.eclipse.ui.progress.PendingUpdateAdapter;
 
+import com.google.common.cache.CacheBuilder;
+import com.google.common.cache.CacheLoader;
+import com.google.common.cache.LoadingCache;
```
mickaelistria, line 71: Can't we deal with a simple map here?

org.eclipse.lsp4e/src/org/eclipse/lsp4e/outline/SymbolsLabelProvider.java
```diff
@@ -117,9 +122,21 @@ public void dispose() {
 		super.dispose();
 	}
 
+	private final LoadingCache<Object /*URI | String*/, Optional<IResource>> resourceCache = CacheBuilder.newBuilder() //
+			.maximumSize(200) //
```
mickaelistria, line 126: I don't think we'll have more than 200 symbol types anyway.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/outline/SymbolsLabelProvider.java
```diff
@@ -117,9 +122,21 @@ public void dispose() {
...
+			.expireAfterAccess(30, TimeUnit.SECONDS) //
```
mickaelistria, line 127: why 30 seconds?
Why not caching for the lifecycle of SymbolsLabelProvider? Also shouldn't the cache be static?

rubenporras (approved):
thanks

rubenporras (changes requested):
I have seen that while I was reviewing posted some questions. I think it makes sense to take back my approval so that we do not have conflicting reviews.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/outline/SymbolsLabelProvider.java
rubenporras: I was also wondering about the caching strategy and why we chose 30 seconds and how that would compare to use weak keys. I had approved it without questions because I guess there might not be a best strategy for this case.

mickaelistria (approved):

PR #1174, opened 2025-01-13
fix: use human-friendly sorting in Outline

sebthom:
This addresses https://github.com/eclipse-wildwebdeveloper/wildwebdeveloper/issues/1616

Currently sorting does not sort numbers numerical.

|before PR|after PR|
|-|-|
|image|image|


mickaelistria:
That's a good change.
IIRC there is a similar pattern somewhere in Platform. Did you check whether there isn't such a comparator already existing that we could easily reuse?

rubenporras:
> That's a good change. IIRC there is a similar pattern somewhere in Platform. Did you check whether there isn't such a comparator already existing that we could easily reuse?

I was thinking the same. This seems a very common problem. Could we use some code in the platform or in Guava (to mention one library we already use) to achieve it?

sebthom:
I haven't found one. I naively thought that Comparator.naturalOrder() does the trick but it doesn't.

mickaelistria:
ok, thanks!

Issue #1173, opened 2025-01-13
Path matching in TextSelectionToIVariable does not work on Windows


FlorianKroiss:
The following code does not seem to work on Windows:

https://github.com/eclipse-lsp4e/lsp4e/blob/308d1670430ec8eedad61fcaefa26a2195fa13c7/org.eclipse.lsp4e.debug/src/org/eclipse/lsp4e/debug/breakpoints/TextSelectionToIVariable.java#L124-L129

For example, when debugging a file at `D:\dev\foo.py`, the returned `uri` is `file:///D:/dev/foo.py` and `uri.getPath()` is `/D:/dev/foo.py`. On the other hand, `sourceElement` is `D:\dev\foo.py`.

The following string comparison then fails, because

1. `uri.getPath()` is prefixed with `/`
2. `sourceElement` uses `\` while `uri.getPath()` uses `/` as a separator.

Both of which are only a problem on Windows machines.
My proposed solution is to do something like this:
```java
var sourceUri = Paths.get(sourceElement).toUri();
return Objects.equals(uri.getPath(), sourceUri.getPath()); 
```
But I'm not sure if `sourceElement` is necessarily always a valid file path, so this might require an additional check for `InvalidPathException`?

PR #1175, opened 2025-01-13
fix: Path matching in TextSelectionToIVariable does not work on Windows

FlorianKroiss:
+ Convert sourceElement to URI before comparing paths
+ Bump point version of org.eclipse.lsp4e.debug


FlorianKroiss:
Closes #1173

org.eclipse.lsp4e.debug/src/org/eclipse/lsp4e/debug/breakpoints/TextSelectionToIVariable.java
```diff
@@ -125,7 +128,15 @@ private boolean match(IDocument document, DSPStackFrame frame) {
 			final var uri = LSPEclipseUtils.toUri(document);
 			if (uri == null)
 				return false;
-			return Objects.equals(uri.getPath(), sourceElement);
+
```
mickaelistria, line 131: Please add a comment here just to remember the case of \ conversion

PR #1179, opened 2025-01-14
fix: Prevent LanguageServers.forProject().anyMatching() from blocking

sebthom:
This PR addresses https://github.com/eclipse-lsp4e/lsp4e/issues/894

`LanguageServers.forProject(...).anyMatching()` may unexpectedly block for up to 10 seconds per matching language server.

The issue occurs because `LanguageServers.forProject(...).anyMatching()` internally calls `LanguageServiceAccessor.capabilitiesComply()`, which in turn invokes `LanguageServerWrapper.getServerCapabilities()`. This method waits up to 10 seconds for the server to initialize, causing potential delays.

This fix ensures non-blocking behavior by addressing the synchronous invocation chain.

---

The offending call chain is:
```py
org.eclipse.jface.text.TextViewerHoverManager.doPresentInformation() or
org.eclipse.ui.internal.views.markers.MarkersTreeViewer.doUpdateItem()
-> LSPCodeActionMarkerResolution.checkMarkerResolution()
   -> LanguageServerProjectExecutor.withCapability(ServerCapabilities::getCodeActionProvider)
        .anyMatching()
      -> LanguageServerProjectExecutor.getServers()
         -> LanguageServiceAccessor.getStartedWrappers()
            -> ...capabilitiesComply(ServerCapabilities::getCodeActionProvider)
               -> LanguageServerWrapper.getServerCapabilities()
                  -> ...getInitializedServer().get(10, TimeUnit.SECONDS);
```

### DIFF
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServiceAccessor.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServiceAccessor.java
@@ -23,6 +23,7 @@ import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
+import java.util.LinkedHashMap;
 import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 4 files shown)


org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServiceAccessor.java
```diff
@@ -204,6 +206,27 @@ private static boolean capabilitiesComply(LanguageServerWrapper wrapper,
...
+	 *            predicate testing the capabilities of {@code wrapper}.
+	 * @return The result of applying the capabilities of {@code wrapper} to
+	 *         {@code capabilitiesPredicate}, or {@code false} if
+	 *         {@code capabilitiesPredicate == null} or
+	 *         {@code wrapper.getServerCapabilities() == null}
+	 */
+	private static CompletableFuture<Boolean> capabilitiesComplyAsync(final LanguageServerWrapper wrapper,
+			final @Nullable Predicate<ServerCapabilities> capabilitiesPredicate) {
+		return capabilitiesPredicate == null //
+				? CompletableFuture.completedFuture(false) //
```
rubenporras, line 226: Am I misreading the existing method `capabilitiesComply`? Is that method not returning true if the predicate is null? If so, should we not change the documentation to match the implementation and return `CompletableFuture.completedFuture(true)`?

sebthom: Good catch. fixed.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServiceAccessor.java
```diff
@@ -204,6 +206,27 @@ private static boolean capabilitiesComply(LanguageServerWrapper wrapper,
...
+	 * Asynchronously checks if the given {@code wrapper}'s capabilities comply with the given
+	 * {@code capabilitiesPredicate}.
+	 *
+	 * @param wrapper
+	 *            the server that's capabilities are tested with
+	 *            {@code capabilitiesPredicate}
+	 * @param capabilitiesPredicate
+	 *            predicate testing the capabilities of {@code wrapper}.
+	 * @return The result of applying the capabilities of {@code wrapper} to
+	 *         {@code capabilitiesPredicate}, or {@code false} if
```
rubenporras, line 219: can you change the javadoc to reflect the change in the PR? Could you also change the javadoc of the method right before, which seems to be wrong as well?

sebthom:
> @sebthom , I was planning on doing a release early next week. Since I see you are now working very actively on LSP4E, should I wait for some PR that you may have planned?

Feel free to release as planned. Thanks for checking!

PR #1182, opened 2025-01-15
Improve the performance of LSPFoldingReconcilingStrategy.applyFolding

merks:
- Avoid so many calls ProjectionViewer.postCatchupRequest by computing a map from Position to FoldingAnnotation for existing annotations.
- Use that map to determine if a new annotation needs to be created.
- LSPFoldingReconcilingStrategy.updateAnnotations is no longer required/used so delete it.
- Make markInvalidAnnotationsForDeletion private because the signature has changed.

https://github.com/eclipse-lsp4e/lsp4e/issues/1180

### DIFF
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/folding/LSPFoldingReconcilingStrategy.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/folding/LSPFoldingReconcilingStrategy.java
@@ -180,7 +180,7 @@ public class LSPFoldingReconcilingStrategy
 		// these are what are passed off to the annotation model to
 		// actually create and maintain the annotations
 		final var deletions = new ArrayList<FoldingAnnotation>();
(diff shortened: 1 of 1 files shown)


mickaelistria:
Since the position may change (move 1 character left or right), would it be even more interesting to use Guava's RangeMap to find any overlapping good range to change?

merks:
You should run it under debug control.  The position is computed from the document, it is not recorded in the annotation itself, so when you add some character, everything shifts and basically all this logic typically appears to be a big no-op...

mickaelistria:
OK, thanks!

mickaelistria:
The general idea seems good; and I don't have the opportunity to go for detailed testing of the PR in a near future. However, I'm using snapshots build of everything all the time.
So I suggest we go for merging, and we'll see the outcome upon build ;)

rubenporras:
Okay, let us do that

merks:
You guys are really responsive. I really appreciate that!

sebthom:
Thanks for the prompt fix!

PR #1184, opened 2025-01-15
feat: support initial folding of license header

sebthom:
This PR addresses https://github.com/eclipse-lsp4e/lsp4e/issues/884 by adding a new folding option to automatically fold license headers on initial opening of an editor.

image

If the first foldable comment of a file a license header is determined using a regex `(?i)(copyright|licensed under|all rights reserved|SPDX-License-Identifier)`

The behaviour is different to Eclipse JDT which has an option to generally fold all header comments:
image

### DIFF
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/folding/LSPFoldingReconcilingStrategy.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/folding/LSPFoldingReconcilingStrategy.java
@@ -21,6 +21,7 @@ import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.CompletableFuture;
+import java.util.regex.Pattern;
 
 import org.eclipse.core.runtime.IProgressMonitor;
 import org.eclipse.jdt.annotation.Nullable;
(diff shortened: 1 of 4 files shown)


org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/folding/LSPFoldingReconcilingStrategy.java
```diff
@@ -63,6 +64,9 @@ public class LSPFoldingReconcilingStrategy
 
 	private static final Annotation[] NO_ANNOTATIONS = new Annotation[0];
 
+	private static final Pattern LICENSE_KEYWORDS = Pattern
+			.compile("(?i)(copyright|licensed under|all rights reserved|SPDX-License-Identifier)"); //$NON-NLS-1$
```
rubenporras, line 68: Is this kind of folding not something that should be determined by de server, which can know more about the language?

mickaelistria: Yes, probably, but the LSP does specify such information at the moment; so we cannot use anything like it so far.

sebthom: This was discussed/denied here https://github.com/microsoft/language-server-protocol/pull/1819

rubenporras (approved):

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/folding/LSPFoldingReconcilingStrategy.java
rubenporras: thanks

PR #1185, opened 2025-01-15
feat: add "Copy Value" context menu item to LanguageServers view

sebthom:
This PR adds a context menu item to the Language Servers view to copy the selected cell as text to the clipboard. This is esp. handy for troubleshooting to copy the effective command line or PID of an LS.

image

### DIFF
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/ui/LanguageServersView.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/ui/LanguageServersView.java
@@ -35,10 +35,20 @@ import org.eclipse.jface.viewers.ViewerComparator;
 import org.eclipse.lsp4e.LanguageServerWrapper;
 import org.eclipse.lsp4e.LanguageServiceAccessor;
 import org.eclipse.swt.SWT;
+import org.eclipse.swt.custom.TableCursor;
 import org.eclipse.swt.custom.TableEditor;
+import org.eclipse.swt.dnd.Clipboard;
+import org.eclipse.swt.dnd.TextTransfer;
+import org.eclipse.swt.dnd.Transfer;
+import org.eclipse.swt.events.KeyAdapter;
+import org.eclipse.swt.events.KeyEvent;
 import org.eclipse.swt.events.SelectionAdapter;
 import org.eclipse.swt.events.SelectionEvent;
+import org.eclipse.swt.graphics.Point;
 import org.eclipse.swt.widgets.Composite;
+import org.eclipse.swt.widgets.Menu;
+import org.eclipse.swt.widgets.MenuItem;
+import org.eclipse.swt.widgets.Table;
 import org.eclipse.swt.widgets.TableItem;
 import org.eclipse.swt.widgets.ToolBar;
 import org.eclipse.swt.widgets.ToolItem;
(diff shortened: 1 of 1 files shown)


rubenporras:
Should we not add the possibility to copy with the keyboard as well to make it behave as other menu items?

sebthom:
> Should we not add the possibility to copy with the keyboard as well to make it behave as other menu items?
The view has full row selection. So when you press CTRL+C we don't know which value the user wants.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/ui/LanguageServersView.java
```diff
@@ -213,9 +219,45 @@ public String getText(Object element) {
...
+		// Avoiding the use of MenuManager, table.setMenu(contextMenu), and
+		// SWT.MenuDetect because the table's SWT.FULL_SELECTION interferes with
+		// right-click detection, causing inconsistent behavior and preventing the
+		// context menu from being displayed reliably.
+
+		final var table = viewer.getTable();
+
+		final Menu contextMenu = new Menu(table);
+		final var copyValue = new MenuItem(contextMenu, SWT.NONE);
+		copyValue.setText("Copy Value"); //$NON-NLS-1$
```
rubenporras, line 237:
```suggestion
		copyValue.setText("Copy Cell Value"); //$NON-NLS-1$
```

sebthom:
@rubenporras  I changed the implementation. You can now select a cell value and also use CTRL+C to copy the value

org.eclipse.lsp4e/src/org/eclipse/lsp4e/ui/LanguageServersView.java
```diff
@@ -224,14 +258,81 @@ public void dispose() {
...
+		// SWT.MenuDetect because the table's SWT.FULL_SELECTION interferes with
+		// right-click detection, causing inconsistent behavior and preventing the
+		// context menu from being displayed reliably.
+		final Table table = viewer.getTable();
+		final var contextMenu = new Menu(table);
+		final var copyValueMenuItem = new MenuItem(contextMenu, SWT.NONE);
+		copyValueMenuItem.setText("Copy Cell Value"); //$NON-NLS-1$
+		copyValueMenuItem.addListener(SWT.Selection, event -> copyCurrentCellToClipboard());
+		table.addListener(SWT.MouseDown, event -> {
+			if (event.button == 3 /* Right-click? */
```
rubenporras, line 272: very surprising that there is no better way to find if it is a right-click that hardcoding the number 3, but I could not find anything better in SWT; the documentation only says the button can be 1, 2, 3.

rubenporras (approved):
Thanks for taking up my comment on Ctrl-C.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/ui/LanguageServersView.java
sebthom: Yes, I was surprised by this as well. Ideally, I would use SWT.MenuDetect as the listener event, which would make this check unnecessary. However, for some reason, SWT.MenuDetect becomes unreliable when SWT.FULL_SELECTION is enabled on the table. Then it only triggers intermittently after several right mouse clicks.

PR #1186, opened 2025-01-15
Ignoring failures when checking for hyperlinks

will-lauer:
Adding LanguageServers.addAllSuccessful as an alternative to LanguageServers.addAll so that exceptions from one of the requests being combined can be ignored. Then using this new method in OpenDeclarationHyperlinkDetector when combining the results from definitions, declarations, typeDefinitions, and implementions because some LSPs return errors instead of empty lists for some of these calls. Instead, we combine only the successful results when building the list of hyperlinks, ignoring the errors.

This should resolve https://github.com/eclipse-lsp4e/lsp4e/issues/1169

### DIFF
--- a/org.eclipse.lsp4e/.project
+++ b/org.eclipse.lsp4e/.project
@@ -30,10 +30,16 @@
 			<arguments>
 			</arguments>
 		</buildCommand>
+		<buildCommand>
+			<name>org.whitesource.eclipse.plugin.WSbuilder</name>
+			<arguments>
+			</arguments>
+		</buildCommand>
 	</buildSpec>
 	<natures>
 		<nature>org.eclipse.m2e.core.maven2Nature</nature>
 		<nature>org.eclipse.pde.PluginNature</nature>
 		<nature>org.eclipse.jdt.core.javanature</nature>
+		<nature>org.whitesource.eclipse.plugin.WSNature</nature>
 	</natures>
 </projectDescription>
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServers.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServers.java
@@ -12,6 +12,16 @@
  *******************************************************************************/
 package org.eclipse.lsp4e;
 
+import org.eclipse.core.resources.IProject;
+import org.eclipse.core.runtime.Assert;
+import org.eclipse.jdt.annotation.Nullable;
+import org.eclipse.jface.text.IDocument;
+import org.eclipse.lsp4e.LanguageServersRegistry.LanguageServerDefinition;
+import org.eclipse.lsp4e.internal.ArrayUtil;
+import org.eclipse.lsp4j.ServerCapabilities;
+import org.eclipse.lsp4j.jsonrpc.messages.Either;
+import org.eclipse.lsp4j.services.LanguageServer;
+
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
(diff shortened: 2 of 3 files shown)


org.eclipse.lsp4e/.project
```diff
@@ -30,10 +30,16 @@
 			<arguments>
 			</arguments>
 		</buildCommand>
+		<buildCommand>
```
rubenporras, line 33: this looks unrelated to the ticket you are resolving. Could you move the change to a new PR so that it can be looked at independently?

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServers.java
```diff
@@ -12,6 +12,16 @@
  *******************************************************************************/
 package org.eclipse.lsp4e;
 
+import org.eclipse.core.resources.IProject;
```
rubenporras, line 15: it looks like organize imports is configured different than by default. Could you change it so that the PR does not change the order of the import in all the modified files?

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServers.java
```diff
@@ -410,6 +410,30 @@ public static <T> CompletableFuture<List<T>> addAll(CompletableFuture<List<T>> a
 		return res;
 	}
 
+	/**
+	 * Combines two async lists of results into a single list by adding all the elements of the second list to the first one,
```
rubenporras, line 414:
```suggestion
	 * Combines async lists of results into a single list by adding all the elements of each lists to a new list
```

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServers.java
```diff
@@ -410,6 +410,30 @@ public static <T> CompletableFuture<List<T>> addAll(CompletableFuture<List<T>> a
...
+	 * ignoring abnormal results and only including succeses.
```
rubenporras, line 415:
```suggestion
	 * ignoring any CompletableFuture that completed exceptionally.
```

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServers.java
```diff
@@ -410,6 +410,30 @@ public static <T> CompletableFuture<List<T>> addAll(CompletableFuture<List<T>> a
...
+	 * @param <T> Result type
+	 * @param futures Async results
+	 * @return Async combined result
+	 */
+	public static <T> CompletableFuture<List<T>> addAllSuccessful(CompletableFuture<List<T>>... futures) {
+		CompletableFuture<List<T>> res = CompletableFuture.allOf(futures)
+				.thenApply( v -> {
+					List<T> results = new ArrayList<>(futures.length);
+					for(CompletableFuture<List<T>> f : futures) {
+						if (!f.isCompletedExceptionally()) {
```
rubenporras, line 426: Instead of just ignoring exceptions, I think we should log them using `LanguageServerPlugin.logError`

rubenporras (changes requested):
Thanks for the PR. I like the idea but I think we should improve a little bit the changes.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServers.java
will-lauer: I disagree with logging these as errors, as I don't think this is an error in the plugin. Instead, this is handling and acceptable response from the LSP more correctly. The "errors" that are coming back are effectively saying "no matching items found". This seems like something the plugin needs to handle and not clutter the log with error messages.

rubenporras: I agree the error did not happen in the client. If not logging them as an error, how do you propose that the user can realize that he did not get some results because the server returned errors?

In your particular example, I would expect that the server does not come back with an exception saying "no matching items found", but rather an empty list or null.

akurtakov:
@will-lauer Would you please react to review comments?

PR #1191, opened 2025-01-17
add custom gson config

yanshudan:


org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServersRegistry.java
```diff
@@ -208,6 +215,24 @@ public Class<? extends LanguageServer> getServerInterface() {
 			return super.getServerInterface();
 		}
 
+		@SuppressWarnings("unchecked")
+		@Override
+		public Consumer<GsonBuilder> getConfigureGson() {
```
rubenporras, line 220: I think you should use the method `createLauncherBuilder` as a template

rubenporras: Indeed, I think you can create a helper method that can be reused by both methods, just passing `CONFIGURE_GSON_ATTRIBUTE` or `LAUNCHER_BUILDER_ATTRIBUTE`

PR #1193, opened 2025-01-21
feat: Use DSPValue::getValueString for detail computation

FlorianKroiss:


FlorianKroiss:
I noticed that the detail area is not used when debugging with DAP. The following UI elements make use of the detail computation. Before this patch, they were left empty:
image
image
image

I find the details especially helpful in the hover dialog, where they allow you to directly copy the value.

Issue #1199, opened 2025-01-27
[DAP] Editing an expression after a debug session completed results in an "Unexpected null value present!" error


agarciadom:
Within the Eclipse Epsilon project, we're using LSP4E's DAP support for our own debugging. One of our users has reported getting spurious "Unexpected null value present!" errors while using the Expressions view, like this one:

![Image](https://github.com/user-attachments/assets/7fbeb176-d97a-4235-a967-e67ce209c2bb)

The specific stack trace is this:

```
java.lang.IllegalStateException: Unexpected null value present!
	at org.eclipse.lsp4e.internal.NullSafetyHelper.castNonNull(NullSafetyHelper.java:36)
	at org.eclipse.lsp4e.debug.presentation.DAPWatchExpression.evaluateExpression(DAPWatchExpression.java:35)
	at org.eclipse.debug.internal.core.WatchExpression.evaluate(WatchExpression.java:78)
	at org.eclipse.debug.internal.core.WatchExpression.setExpressionText(WatchExpression.java:257)
	at org.eclipse.debug.internal.ui.elements.adapters.WatchExpressionCellModifier.modify(WatchExpressionCellModifier.java:60)
	at org.eclipse.debug.internal.ui.viewers.model.InternalTreeModelViewer$CellModifierProxy.modify(InternalTreeModelViewer.java:223)
	at org.eclipse.jface.viewers.ColumnViewer$2.setValue(ColumnViewer.java:272)
	at org.eclipse.jface.viewers.EditingSupport.saveCellEditorValue(EditingSupport.java:115)
	at org.eclipse.jface.viewers.ColumnViewerEditor.saveEditorValue(ColumnViewerEditor.java:444)
	at org.eclipse.jface.viewers.ColumnViewerEditor.applyEditorValue(ColumnViewerEditor.java:312)
	at org.eclipse.jface.viewers.ColumnViewerEditor$1.applyEditorValue(ColumnViewerEditor.java:165)
	at org.eclipse.jface.viewers.CellEditor$1.run(CellEditor.java:335)
	at org.eclipse.core.runtime.SafeRunner.run(SafeRunner.java:47)
	at org.eclipse.jface.util.SafeRunnable.run(SafeRunnable.java:174)
	at org.eclipse.jface.viewers.CellEditor.fireApplyEditorValue(CellEditor.java:332)
	at org.eclipse.jface.viewers.TextCellEditor.handleDefaultSelection(TextCellEditor.java:286)
[...]

PR #1200, opened 2025-01-27
DAPWatchExpression: avoid 'Unexpected null value' error when editing expressions

agarciadom:
This is a proposed fix for #1199. I add a `null` check to the result of the `adapt` call, as this adaptation may return `null` if you try to edit or re-evaluate the expression after the program has finished running (e.g. between repeated debugging sessions of the same program).

### DIFF
--- a/org.eclipse.lsp4e.debug/src/org/eclipse/lsp4e/debug/presentation/DAPWatchExpression.java
+++ b/org.eclipse.lsp4e.debug/src/org/eclipse/lsp4e/debug/presentation/DAPWatchExpression.java
@@ -32,10 +32,19 @@ public class DAPWatchExpression implements IWatchExpressionDelegate {
 		if (context.getDebugTarget() instanceof DSPDebugTarget dapDebugger) {
 			final var args = new EvaluateArguments();
 			args.setExpression(expression);
-			DSPStackFrame frame = castNonNull(Adapters.adapt(context, DSPStackFrame.class));
-			args.setFrameId(frame.getFrameId());
-			dapDebugger.getDebugProtocolServer().evaluate(args).thenAccept(
-					res -> listener.watchEvaluationFinished(createWatchResult(dapDebugger, expression, res)));
+
+			@Nullable
+			DSPStackFrame dspStackFrame = Adapters.adapt(context, DSPStackFrame.class);
+			if (dspStackFrame != null) {
+				DSPStackFrame frame = castNonNull(dspStackFrame);
+				args.setFrameId(frame.getFrameId());
+				dapDebugger.getDebugProtocolServer().evaluate(args).thenAccept(
+						res -> listener.watchEvaluationFinished(createWatchResult(dapDebugger, expression, res)));
+			} else {
+				EvaluateResponse res = new EvaluateResponse();
+				res.setResult("");
+				listener.watchEvaluationFinished(createWatchResult(dapDebugger, expression, res));
+			}
 		}
 	}
 


org.eclipse.lsp4e.debug/src/org/eclipse/lsp4e/debug/presentation/DAPWatchExpression.java
```diff
@@ -32,10 +32,20 @@ public void evaluateExpression(String expression, IDebugElement context, IWatchE
...
+			if (dspStackFrame != null) {
+				DSPStackFrame frame = castNonNull(dspStackFrame);
+				args.setFrameId(frame.getFrameId());
+				dapDebugger.getDebugProtocolServer().evaluate(args).thenAccept(
+						res -> listener.watchEvaluationFinished(createWatchResult(dapDebugger, expression, res)));
+			} else {
+				// Program not running at the moment
+				EvaluateResponse res = new EvaluateResponse();
+				res.setResult("(pending)");
+				listener.watchEvaluationFinished(createWatchResult(dapDebugger, expression, res));
```
jonahgraham, line 47: I think this should just set the evaluation to `null` in this case. I am pretty sure that is what is done in other debuggers like  JDT ([link to line 86 of JavaWatchExpressionDelegate](https://github.com/eclipse-jdt/eclipse.jdt.debug/blob/6c5cadc5c3dc4a125d50f566e0e70d5cd9f89c32/org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/JavaWatchExpressionDelegate.java#L86)) when it can't evaluate.

```suggestion
				// Program not running at the moment
				listener.watchEvaluationFinished(null);
```

I think it depends if you want to show an error/warning in this case, or just leave the value field blank.

jonahgraham (commented):
@agarciadom thanks for the PR and I think you are on the right track with this change. I can't test/confirm my suggestion at the moment so if it seems incorrect let me know.

BTW if it helps (and I am recording this for my own future reference too), Eclipse JDT is showing no classes implementing `IWatchExpressionListener` or overriding `watchEvaluationFinished` which made it tricky to dive into this code. This is because of how the listener is installed [here](https://github.com/eclipse-platform/eclipse.platform/blob/e44cc1b8be5011ce3e8a459d0b41777ac6333869/debug/org.eclipse.debug.core/core/org/eclipse/debug/internal/core/WatchExpression.java#L74)

org.eclipse.lsp4e.debug/src/org/eclipse/lsp4e/debug/presentation/DAPWatchExpression.java
agarciadom: Thank you for the suggestion! I think it makes sense to follow the same approach as JDT, for consistency :-). I have updated my PR to just pass `null` when we can't adapt the context.

agarciadom: Just to add - I've tested this version that uses `null`, and it simply leaves the value field blank as one would expect.

agarciadom:
Oops: if I pass `null` as you suggested, the null checks during the build process fail. I'll try creating a watch result with `null` as the value.

agarciadom:
Using `createWatchResult(dapDebugger, expression, null)` produces a runtime error:

```
An internal error has occurred.
Cannot invoke "org.eclipse.lsp4j.debug.EvaluateResponse.getVariablesReference()" because "this.val$res" is null
```

I'll just use an empty string as the value of the `EvaluateResponse`. This would look the same as what JDT does (leaving the value field empty in the UI), but should also pass the null checks in the build process.

jonahgraham:
> Oops: if I pass `null` as you suggested, the null checks during the build process fail:

Perhaps not something for today, but I think that this is a build error that we should resolve and not write less correct code. I suspect it is happening because of non-null by default annotation and that the watchEvaluationFinished is missing a nullable annotation.

Perhaps one of the more active LSP4E contributors can shed some light on this. Does anyone in @eclipse-lsp4e/technology-lsp4e-committers have a thought on how to call Eclipse Platfom API that accepts a null but doesn't have the correct annotations?

If we can't easily resolve the build error, then doing an empty string seems like a reasonable workaround.

rubenporras:
Hi @jonahgraham ,

external libraries can be annotated locally. In LSP4E however we rely on a project https://github.com/vegardit/no-npe maintained by @sebthom, so the process would be to add the external annotation to no-npe, release it and upgrade the version in our target file.

To me, given the last comment of @agarciadom , it looks like the easiest to take the PR as it is now. What do you think?

jonahgraham (approved):

jonahgraham:
Thanks @agarciadom for fixing this error.

Thanks @rubenporras for the complete explanation.

I have approved and applied the changes.

PR #1202, opened 2025-01-31
Update to LSP4J 0.24.0

jonahgraham:

PR #1204, opened 2025-01-31
test: remove dependency bundle-version constrains from test modules

sebthom:
Specifying exact versions is redundant and leads to unnecessary version bumps of the test plugins. Dependency versions will be resolved via the target platform definition used during the build.


jonahgraham:
Thanks @sebthom. I'm glad to not have to touch those bundles next time I have to make an update like #1202 (and thanks for merging that too)

PR #1205, opened 2025-01-31
limit LS console text buffer to prevent OOM

ghentschke:


org.eclipse.lsp4e/src/org/eclipse/lsp4e/LoggingStreamConnectionProviderProxy.java
```diff
@@ -286,6 +286,7 @@ private MessageConsole findConsole() {
 		}
 		// no console found, so create a new one
 		final var myConsole = new MessageConsole(Messages.LSConsoleName, null);
+		myConsole.setWaterMarks(80000, 800000); // limit text buffer size to prevent OOM
```
sebthom, line 289: Could you use thousand separators for better readability?

PR #1208, opened 2025-02-05
[#1207] cache IResource in EnablementTester

### LINKED ISSUE #1207: [performance] EnablementTester.evaluate takes way to long in UI thread
Due to some performance issues in CDT-LSP I found out that the `EnablementTester.evaluate` method consumes a lot of time + it gets called very frequently (on ever user typing in the editor):
The profiling while editing a C-File returns me this behavior:

Image

My proposal would be to cash the IResource object found for the URI here, since `findResourceFor` seems to be the bottleneck:
The resource handle could cashed until its available.

Image

ghentschke:
Cache resource for URI because each call of
LSPEclipseUtils.findResourceFor(URI) takes ~300 microseconds. And it gets called a lot of times.

fixes #1207

### DIFF
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/ResourceForUriCache.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/ResourceForUriCache.java
@@ -12,52 +12,50 @@
 
 package org.eclipse.lsp4e;
 
-import static org.eclipse.lsp4e.internal.NullSafetyHelper.castNonNull;
-
 import java.net.URI;
 
 import org.eclipse.core.resources.IResource;
-import org.eclipse.core.resources.IResourceChangeEvent;
-import org.eclipse.core.resources.IResourceChangeListener;
-import org.eclipse.core.resources.IResourceDelta;
-import org.eclipse.core.runtime.CoreException;
-import org.eclipse.core.runtime.Platform;
 import org.eclipse.jdt.annotation.Nullable;
 
 import com.google.common.cache.Cache;
 import com.google.common.cache.CacheBuilder;
 
 /**
- * Caches the {@link LSPEclipseUtils#findResourceFor(URI)} call. <code>ResourceForUriCache.getInstance().get(URI)</code>
+ * <p>Caches the {@link LSPEclipseUtils#findResourceFor(URI)} call. <code>ResourceForUriCache.getInstance().get(URI)</code>
  * can an be used instead of a direct call to LSPEclipseUtils.findResourceFor(URI),
  * because the call is very time consuming.
  *
- * The cache is limited to 100 resources elements.
- * Resources are removed in case the URI has changed due to a resource move or delete operation.
+ * <p>NOTE: In case a resource has been moved or deleted the entry will not be removed automatically.
+ * It's up to the caller to check if the resource is accessible.
+ *
+ * <p>The cache is limited to 100 resource elements. It uses least-recently-used eviction if limit exceeds.
+ * The cache will try to evict entries that haven't been used recently.
+ * Therefore entries can be removed before the limit exceeds.
  */
-public final class ResourceForUriCache implements IResourceChangeListener {
-	private static final int URI_CHANGED = IResourceDelta.REPLACED |
-			IResourceDelta.MOVED_FROM |
-			IResourceDelta.MOVED_TO |
-			IResourceDelta.REMOVED |
-			IResourceDelta.REMOVED_PHANTOM;
-
+public final class ResourceForUriCache {
 	private static final Cache<URI, IResource> cache =  CacheBuilder.newBuilder().maximumSize(100).build();
-	private static @Nullable ResourceForUriCache instance;
+	private static final ResourceForUriCache INSTANCE = new ResourceForUriCache();
 
 	private ResourceForUriCache() {
 		//  use getInstance()
 	}
 
-	public static synchronized ResourceForUriCache getInstance() {
-		if (instance == null) {
-			instance = new ResourceForUriCache();
-		}
-		return castNonNull(instance);
+	public static ResourceForUriCache getInstance() {
+		return INSTANCE;
 	}
 
+	/**
+	 * <p>Returns the cached IResource for the given URI. Tries to determine the IResource
+	 * if it's not already in the cache. Returns NULL if the IResource could not be determined,
+	 * e.g. the URI points to a file outside the workspace.
+	 *
+	 * <p>NOTE: In case a resource has been moved or deleted the entry will not be removed automatically.
+	 * It's up to the caller to check if the resource is accessible.
+	 * @param uri
+	 * @return IResource or NULL
+	 */
 	@Nullable
-	public synchronized IResource get(@Nullable URI uri) {
+	public IResource get(@Nullable URI uri) {
 		IResource resource = null;
 		if (uri != null) {
 			resource = cache.getIfPresent(uri);
@@ -71,26 +69,6 @@ public final class ResourceForUriCache implements IResourceChangeListener {
 		}
 		return resource;
 	}
-
-	@Override
-	public void resourceChanged(IResourceChangeEvent event) {
-		if (event.getDelta() != null) {
-			try {
-				event.getDelta().accept(delta -> {
-					if ((delta.getKind() & URI_CHANGED) != 0) {
-						var uri = delta.getResource().getLocationURI();
-						if (uri != null) {
-							cache.invalidate(uri);
-						}
-					}
-					return true;
-				});
-			} catch (CoreException e) {
-				Platform.getLog(getClass()).error(e.getMessage(), e);
-			}
-		}
-	}
-
 }
 
 
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/ResourceForUriCache.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/ResourceForUriCache.java
@@ -77,7 +77,7 @@ public final class ResourceForUriCache implements IResourceChangeListener {
 		if (event.getDelta() != null) {
 			try {
 				event.getDelta().accept(delta -> {
-					if ((delta.getKind() | URI_CHANGED) == URI_CHANGED) {
+					if ((delta.getKind() & URI_CHANGED) != 0) {
 						var uri = delta.getResource().getLocationURI();
 						if (uri != null) {
 							cache.invalidate(uri);
@@ -86,7 +86,7 @@ public final class ResourceForUriCache implements IResourceChangeListener {
 					return true;
 				});
 			} catch (CoreException e) {
-				Platform.getLog(getClass()).log(e.getStatus());
+				Platform.getLog(getClass()).error(e.getMessage(), e);
 			}
 		}
 	}
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/ResourceForUriCache.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/ResourceForUriCache.java
@@ -1,0 +1,96 @@
+/*******************************************************************************
+ * Copyright (c) 2025 Contributors to the Eclipse Foundation.
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *   See git history
+ *******************************************************************************/
+
+package org.eclipse.lsp4e;
+
+import static org.eclipse.lsp4e.internal.NullSafetyHelper.castNonNull;
+
+import java.net.URI;
+
+import org.eclipse.core.resources.IResource;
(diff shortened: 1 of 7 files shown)


org.eclipse.lsp4e/src/org/eclipse/lsp4e/enablement/EnablementTester.java
```diff
@@ -69,7 +72,7 @@ public boolean evaluate(@Nullable URI uri) {
 		IResource resource = null;
 		try {
 			IDocument document = null;
-			resource = LSPEclipseUtils.findResourceFor(uri);
```
rubenporras, line 72: I think there are several places where we map resources to URIs.  Should we not benefit by having the cache reused by all consumers of the method? If so, what about building the cache as an static map in `LSPEclipseUtils`?

org.eclipse.lsp4e/src/org/eclipse/lsp4e/enablement/EnablementTester.java
```diff
@@ -104,4 +107,25 @@ public boolean evaluate(@Nullable URI uri) {
...
+	 * @param uri
+	 * @return
+	 */
+	private @Nullable IResource getResourceFor(@Nullable URI uri) {
+		if (uri != null) {
+			var resource = cache.get(uri);
+			if (resource != null && resource.isAccessible()) {
+				return resource;
+			}
+			resource = LSPEclipseUtils.findResourceFor(uri);
```
rubenporras, line 122: could it happen that a resource is moved after being cached? Would that not invalidate the cache? If that could happen, should we not either detect to be able to invalidate the cache or have some eviction policy?

martinlippert:
One question: is there a mechanism to clean up the cache? It looks like it is an ever growing map that keeps references to all those IResource objects forever. Could that be a problem memory-wise over time?

mickaelistria:
> LSPEclipseUtils.findResourceFor(URI) takes ~300 microseconds. And it gets called a lot of times.

Shouldn't such a strategy be implemented directly in `LSPEclipseUtils.findResourceFor(uri)` instead? And isn't there some other possible improvement to find there, or even in Eclipse Platform?

org.eclipse.lsp4e/src/org/eclipse/lsp4e/enablement/EnablementTester.java
ghentschke: > could it happen that a resource is moved after being cached?

Yes, but  shouldn't it be handled by the `resource.isAccessible` call: 
> Returns whether this resource is accessible. For files and folders, this is equivalent to existing;

ghentschke:
> One question: is there a mechanism to clean up the cache? It looks like it is an ever growing map that keeps references to all those IResource objects forever. Could that be a problem memory-wise over time?

Yes, I'll add a clear mechanism which could be triggered when the `expression.evaluate` returns false for the given URI.

ghentschke:
> Shouldn't such a strategy be implemented directly in `LSPEclipseUtils.findResourceFor(uri)` instead? And isn't there some other possible improvement to find there, or even in Eclipse Platform?

Yes. I'll consider it.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/enablement/EnablementTester.java
ghentschke: If the resource has been moved, the uri will be changed as well. This leads to a new entry in the cache and to an onsolete entry.

rubenporras: Yes, you are correct that its URI will change, but the cache will still have an entry for the old URI and will return the resource, will it not?

ghentschke: IMO that's not a problem, if the old URI isn't cached anywhere.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/enablement/EnablementTester.java
```diff
@@ -40,6 +42,7 @@ public final class EnablementTester {
 	private final Expression expression;
 	private final String description;
 	private final Supplier<@Nullable IEvaluationContext> parent;
+	private final Map<URI, IResource> cache = new HashMap<>();
```
sebthom, line 45: How about using Guava Cache instead which supports expiring entries after read access and is thread safe. LSP4E already depends on guava so no new dependency needs to be added.

ghentschke: Thanks for the hint. I'll take a closer look at it.

jonahgraham: @sebthom The Javadocs for Guava's [says](https://guava.dev/releases/snapshot-jre/api/docs/com/google/common/cache/CacheBuilder.html#prefer-caffeine-over-guava-s-caching-api-heading) to prefer using [Caffeine](https://github.com/ben-manes/caffeine/wiki) instead. 

AFAICT nothing contributes caffeine to SimRel yet, but some projects do use it (e.g. [Mylyn](https://github.com/eclipse-mylyn/org.eclipse.mylyn/blob/d33888928631a359cf5f3820c7103d3b84e53a6e/org.eclipse.mylyn-site/category.xml#L179), [m2e's tests](https://github.com/eclipse-m2e/m2e-core/blob/c45423e5a1603887ea0e20d42e296c4e6bf675b6/org.eclipse.m2e.swtbot.tests/src/org/eclipse/m2e/pde/ui/MavenTargetDependencyEditorTest.java#L66)) and it is in [orbit aggregator for 2025-03 too](https://download.eclipse.org/tools/orbit/simrel/orbit-aggregation/2025-03)

Any thoughts on best way to proceed?

sebthom: I personally would be pragmatic and just use guava instead of introducing another dependency. I think it is good enough.

rubenporras: I think the same.

jonahgraham: Thanks @sebthom and @rubenporras for your input.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/enablement/EnablementTester.java
ghentschke: @rubenporras  Please review my current implementation. I use a `IResourceChangeListener` now to track modifications and to remove deleted/moved resources from the cache.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/ResourceForUriCache.java
```diff
@@ -0,0 +1,96 @@
...
+
+	public static synchronized ResourceForUriCache getInstance() {
+		if (instance == null) {
+			instance = new ResourceForUriCache();
+		}
+		return castNonNull(instance);
+	}
+
+	@Nullable
+	public synchronized IResource get(@Nullable URI uri) {
```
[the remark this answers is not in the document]
ghentschke: That was my first attempt. It seems that `LoadingCache` is not really appropriate for this use case, because the laod method has to return a non null value:
```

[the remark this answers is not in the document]
Comment (sebthom) [reply org.eclipse.lsp4e/src/org/eclipse/lsp4e/ResourceForUriCache.java:60]: Ah true, that only works with Caffeine. I personally then would just remove the synchronized keyword and take the risk of a race conditions over synchronizing the method. The worst case is that a URI is looked up twice at the same time. Or alternatively use `synchronized(cache) { ...  }` inside the `if (uri != null)` block. But I believe that isn't really necessary.

File: org.eclipse.lsp4e/src/org/eclipse/lsp4e/ResourceForUriCache.java [hunk 5]
```diff
@@ -0,0 +1,96 @@
...
+			IResourceDelta.REMOVED_PHANTOM;
+
+	private static final Cache<URI, IResource> cache =  CacheBuilder.newBuilder().maximumSize(100).build();
+	private static @Nullable ResourceForUriCache instance;
+
+	private ResourceForUriCache() {
+		//  use getInstance()
+	}
+
+	public static synchronized ResourceForUriCache getInstance() {
```
sebthom, line 52: how about instead of adding a synchronized getInstance method simply defining a `public static final ResourceForUriCache INSTANCE = new ResourceForUriCache();`

org.eclipse.lsp4e/src/org/eclipse/lsp4e/ResourceForUriCache.java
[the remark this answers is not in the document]
ghentschke: I removed the synchronization from the get method because current usage in `EnablementTester.evaluate(URI)` runs AFAIK in main thread only.
I removed the `IResourceChangeListener` as well because it won't worked as expected (see git commit comment) and the guave cache has already a mechanism to remove unused entries as well.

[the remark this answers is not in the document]
ghentschke: I've to withdraw my statement above: there are concurrent calls to `EnablementTester.evaluate(URI)`

sebthom (approved):

org.eclipse.lsp4e/src/org/eclipse/lsp4e/ResourceForUriCache.java
```diff
@@ -0,0 +1,74 @@
...
+	 */
+	@Nullable
+	public IResource get(@Nullable URI uri) {
+		IResource resource = null;
+		if (uri != null) {
+			resource = cache.getIfPresent(uri);
+			if (resource != null) {
+				return resource;
+			}
+			resource = LSPEclipseUtils.findResourceFor(uri);
```
mickaelistria, line 65: Should we move the logic of `findResourceFor` in this cache too, and then make `LSPEclipseUtils.findResourceFor(...)` reference `ResourceForUriCache.getInstance().get(...)` ?

Doing so will make all consumers of findResourceFor(...) save 300ms, not just en enablementTester

rubenporras: I also think that is a good idea

org.eclipse.lsp4e/src/org/eclipse/lsp4e/ResourceForUriCache.java
```diff
@@ -0,0 +1,74 @@
...
+ *
+ * <p>NOTE: In case a resource has been moved or deleted the entry will not be removed automatically.
+ * It's up to the caller to check if the resource is accessible.
+ *
+ * <p>The cache is limited to 100 resource elements. It uses least-recently-used eviction if limit exceeds.
+ * The cache will try to evict entries that haven't been used recently.
+ * Therefore entries can be removed before the limit exceeds.
+ */
+public final class ResourceForUriCache {
+	private static final Cache<URI, IResource> cache =  CacheBuilder.newBuilder().maximumSize(100).build();
```
rubenporras, line 36: could we pass a `CacheLoader` to the build method with a call to ` LSPEclipseUtils.findResourceFor`? I think that would save the double access to the cache (get & set) in case of cache miss.

ghentschke: The load method in `CacheLoader` does not allow a null to be returned See [here](https://guava.dev/releases/22.0/api/docs/com/google/common/cache/CacheLoader.html#load-K-):
```
Returns:
the value associated with key; must not be null
```
But it cannot be guaranteed that we find an IResource object for a given URI (e.g. when the file is outside the workspace)

ghentschke: I should add a comment to the method why we do it this way... (See also comment from @sebthom [here](https://github.com/eclipse-lsp4e/lsp4e/pull/1208#discussion_r1953124316))

rubenporras:
looks good to me

org.eclipse.lsp4e/src/org/eclipse/lsp4e/ResourceForUriCache.java
```diff
@@ -0,0 +1,104 @@
...
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *   See git history
+ *******************************************************************************/
+
+package org.eclipse.lsp4e;
```
sebthom, line 13: This should move it into the internal package. I don't think this should be part of the LSP4E API.

sebthom (changes requested):

sebthom (approved):

ghentschke:
Thanks for the reviews!

PR #1221, opened 2025-03-05
Consider viewer's visible region when preparing underline StyleRange[]. Fixes #1220

### LINKED ISSUE #1220: LSPDocumentLinkPresentationReconcilingStrategy does not consider viewer's visible region
When `LSPDocumentLinkPresentationReconcilingStrategy.underline(List<DocumentLink>)` is preparing `StyleRange[]` for `TextPresentation`, that it then passes to `viewer.changeTextPresentation()`, it gets current viewer's styles for target region, sets underline flag for them, and the sets them to text presentation.

However it does not take into consideration the fact that `viewer.getTextWidget().getStyleRanges()` returns style ranges with widget (renderer) cordinates and not document (model) coordinates. `TextPresentation` is expected to have style ranges with document (model) coordinates, since inside `viewer.changeTextPresentation()` the coordinates are translated to widget's coordinates.

When the viewer is not showing the document from the beggining (changed visible region by call to `setVisibleRegion()`), styles from wrong ranges are updated with underline resutling in messed up coloring on parts of thext that becomes underlined. Example:

Image

How to quickly reproduce:

In `LSPDocumentLinkPresentationReconcilingStrategy.setDocument(IDocument)` change the visible region of the viewer
```
@Override
public void setDocument(@Nullable IDocument document) {
	this.document = document;
	// line below is added to reproduce
	viewer.setVisibleRegion(viewer.getVisibleRegion().getOffset() + 500, viewer.getVisibleRegion().getLength() - 500);
}
```
and open some HTML document in GenericEditor (uses LSP4E) that has hrefs that gets underlined.

(I stubled acrros this when preparing/debugging GenericEditor contribution to new QuickSearch text viewer extension 
https://github.com/eclipse-platform/eclipse.platform.ui/issues/2010)

RedeemerSK:
Before
Image
After
image
(with quick-reproducer-code mentioned in #1220 )

(I haven't checked LSP4E repository for other possible places where widget's style ranges may be considered to have viewer's document (model) coordinates)

### DIFF
--- a/org.eclipse.lsp4e.test/src/org/eclipse/lsp4e/test/documentLink/DocumentLinkReconcilingTest.java
+++ b/org.eclipse.lsp4e.test/src/org/eclipse/lsp4e/test/documentLink/DocumentLinkReconcilingTest.java
@@ -1,0 +1,189 @@
+/*******************************************************************************
+ * Copyright (c) 2025 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * https://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *     Jozef Tomek - initial implementation
+ *******************************************************************************/
+
+package org.eclipse.lsp4e.test.documentLink;
+
+import static org.junit.Assert.assertEquals;
+
+import java.util.ArrayList;
+import java.util.List;
+
+import org.eclipse.jface.text.BadLocationException;
+import org.eclipse.jface.text.IDocument;
+import org.eclipse.jface.text.Region;
+import org.eclipse.jface.text.TextPresentation;
+import org.eclipse.jface.text.TextViewer;
+import org.eclipse.lsp4e.test.utils.AbstractTestWithProject;
+import org.eclipse.lsp4e.test.utils.TestUtils;
+import org.eclipse.lsp4e.tests.mock.MockLanguageServer;
+import org.eclipse.lsp4j.DocumentLink;
+import org.eclipse.lsp4j.Position;
+import org.eclipse.lsp4j.Range;
+import org.eclipse.swt.custom.StyleRange;
+import org.eclipse.swt.graphics.Color;
+import org.junit.Test;
+
+public class DocumentLinkReconcilingTest extends AbstractTestWithProject {
+	
+	private static final String CONTENT = """
+				1st_line #LINK1 1st_line
+				2nd_line #LINK2_START
+				#LINK2_END 3rd_line #LINK3 3rd_line
+				4th_line #LINK4 4th_line
+				5th_line #LINK5_START_#LINK5_END
+				6th_line #LINK6""";
+	
+	private static final List<DocumentLink> CONTENT_LINKS = List.of(
+			new DocumentLink(new Range(new Position(0, 9), new Position(0, 15)), "file://link1"),
+			new DocumentLink(new Range(new Position(1, 9), new Position(2, 10)), "file://link2"),
+			new DocumentLink(new Range(new Position(2, 20), new Position(2, 26)), "file://link3"),
+			new DocumentLink(new Range(new Position(3, 9), new Position(3, 15)), "file://link4"),
+			new DocumentLink(new Range(new Position(4, 9), new Position(4, 32)), "file://link5"),
+			new DocumentLink(new Range(new Position(5, 9), new Position(5, 15)), "file://link6"));
+	
+	public static final Color COLOR_1ST_LINE = new Color(255, 0, 0);
+	public static final Color COLOR_2ND_LINE = new Color(0, 255, 0);
+	public static final Color COLOR_3RD_LINE = new Color(0, 0, 255);
+	public static final Color COLOR_4TH_LINE = new Color(255, 255, 0);
+	public static final Color COLOR_5TH_LINE = new Color(0, 255, 255);
+	public static final Color COLOR_6TH_LINE = new Color(255, 0, 255);
+	
+	private List<TextPresentation> textPresentations = new ArrayList<>(4);
+	
+	@Test
+	public void testFullDocumentLinkReconciling() throws Exception {
+		MockLanguageServer.INSTANCE.setDocumentLinks(CONTENT_LINKS);
+
+		TextViewer viewer = (TextViewer) TestUtils.openTextViewer(TestUtils.createUniqueTestFile(project, CONTENT));
+		IDocument document = viewer.getDocument();
+		viewer.getTextWidget().setStyleRanges(new StyleRange[] {
+				textStyle(0, 25, COLOR_1ST_LINE), // whole 1st line
+				textStyle(25, 22, COLOR_2ND_LINE), // whole 2nd line
+				textStyle(47, 36, COLOR_3RD_LINE), // whole 3rd line
+				textStyle(83, 25, COLOR_4TH_LINE), // whole 4th line
+				textStyle(108, 33, COLOR_5TH_LINE), // whole 5th line
+				textStyle(141, 15, COLOR_6TH_LINE) // whole 6th line
+		});
+		viewer.addTextPresentationListener(this::textPresentationListener);
+		
+		TestUtils.waitForAndAssertCondition(1_000, () -> textPresentations.size() == 6);
+		
+		assertEquals(linkRegion(CONTENT_LINKS.get(0), document), textPresentations.get(0).getExtent());
+		assertEquals(linkRegion(CONTENT_LINKS.get(1), document), textPresentations.get(1).getExtent());
+		assertEquals(linkRegion(CONTENT_LINKS.get(2), document), textPresentations.get(2).getExtent());
+		assertEquals(linkRegion(CONTENT_LINKS.get(3), document), textPresentations.get(3).getExtent());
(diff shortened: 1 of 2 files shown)


RedeemerSK:
This should probably include update to existing / new test as well -.-

RedeemerSK:
> Thanks for the PR. Yes, it would be great if you could add / modify a test

Done. I think the builds are failing for reason not related to any of my changes.

Once that goes away I think it would make sense to see that the build of just the 1st commit (adding tests) fails and with 2nd commit new tests would pass - if anyone can trigger such 2 build manually, I wouldn't need to trick Jenkins by force-pushing just the 1st commit and then later 2nd commit.

org.eclipse.lsp4e.test/src/org/eclipse/lsp4e/test/documentLink/DocumentLinkReconcilingTest.java
```diff
@@ -0,0 +1,189 @@
...
+				thirdLineStart, // TextViewer would align start of visible region to start of the line anyway
+				untilMiddleOfLink5);
+		viewer.getTextWidget().setStyleRanges(new StyleRange[] {
+				textStyle(0, 36, COLOR_3RD_LINE), // whole 3rd line
+				textStyle(36, 25, COLOR_4TH_LINE), // whole 4th line
+				textStyle(61, 21, COLOR_5TH_LINE) // whole last displayed line - visible part of 5th line
+		});
+		viewer.addTextPresentationListener(this::textPresentationListener);
+		
+		TestUtils.waitForAndAssertCondition(1_000, () -> textPresentations.size() == 6);
```
RedeemerSK, line 141: Without the fix (2nd commit) it fails here since `LSPDocumentLinkPresentationReconcilingStrategy` would try to get style ranges for invalid coordinates (falling outside of visible region) from viewer's text widget which would cause `IllegalArgumentException` . Also it would end up looking like this

image

With fix:
image

mickaelistria:
Code and result and test look good. Can you please also share screenshots of an example involving folding ranges to make sure this is compatible?

rubenporras:
> Once that goes away I think it would make sense to see that the build of just the 1st commit (adding tests) fails and with 2nd commit new tests would pass - if anyone can trigger such 2 build manually, I wouldn't need to trick Jenkins by force-pushing just the 1st commit and then later 2nd commit.

Thanks, no need for that. We can check locally. I will try to find some time this week to review the PR.

RedeemerSK:
> Code and result and test look good. Can you please also share screenshots of an example involving folding ranges to make sure this is compatible?

Took me while to find the setup where I would have both folding and links in a document (did'n know there is Language Servers / Folding in preferences) and it does produce misaligned/broken syles with fodling. Luckily the fix has the potential to make the code even simpler since all that needs to be done is to use `TextViewer.modelOffset2WidgetOffset()` that considers both visible region and folding it seems.

I'll most probably do `instanceof` check against `TextViewer` rather than `ITextViewerExtension5` since I expect there may be more viewers extending `TextViewer` & not implementing `ITextViewerExtension5` than those implementing `ITextViewerExtension5` but not extending `TextViewer` - if it ever happens that they're not subclass of `ProjectionViewer` (thus both). Still I'm mentioning this here since I'm not completely sure that's the way to go.

RedeemerSK:
Now works with folding

testFullDocumentLinkReconciling:
testFullDocumentLinkReconciling
testDocumentWithFoldingLinkReconciling:
testDocumentWithFoldingLinkReconciling
testClippedDocumentLinkReconciling:
testClippedDocumentLinkReconciling

However in real usage scenario with GenericEditor and folding enabled it may be useless since if the timing is "right" it might happen that:
- editor is opened, whole document is displayed
- then `LSPDocumentLinkPresentationReconcilingStrategy` does it's job, seeing viewer's widget without any folding
- then `LSPFoldingReconcilingStrategy` does it's job, folding areas
result:  underline styles are lost

lsp_links_gone_on_folding

I see this - links underline styles disappearing - happening very often now that I pay attention. Only once I started testing my new quicksearch textviewer extension with generic editor I noticed there links getting underlined in documents. Previously I don't even remember seeing them at all.

But this is IMHO architectural issue - `LSPDocumentLinkPresentationReconcilingStrategy` is just not present in all "pipelines" where something is setting new styles to viewer's widget. Like eg. simple hover:
lsp_links_gone_on_hover
But that's another story.

Thanks in advance for another round of code review.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/documentLink/LSPDocumentLinkPresentationReconcilingStrategy.java
```diff
@@ -93,23 +94,70 @@ private void underline(@Nullable List<DocumentLink> links) {
...
+						}
+						if (visibleStart > linkRegion.getOffset() || visibleLength < linkRegion.getLength()) {
+							linkRegion = new Region(visibleStart, visibleLength);
+						}
+					} else {
+						// whole link is not visible - outside model coverage (visible region) or inside folded areas
+						continue;
+					}
+				} else {
+					styleRanges = viewer.getTextWidget().getStyleRanges(start, length);
```
RedeemerSK, line 153: Here we could still try to consider at least visible region by using `org.eclipse.jface.text.ITextViewer.getVisibleRegion()` (what I was doing in my 2nd commit before finding out folding breaks things) but I think there's very low chance that the viewer will not be `org.eclipse.jface.text.TextViewer` or its sublcass. Therefore I decided not to make this code even more complicated.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/documentLink/LSPDocumentLinkPresentationReconcilingStrategy.java
```diff
@@ -93,23 +94,70 @@ private void underline(@Nullable List<DocumentLink> links) {
...
+							visibleLength =
+								textViewer.widgetOffset2ModelOffset(widgetOffset + widgetRange.getLength() - 1)
+								- visibleStart;
+						}
+						if (visibleStart > linkRegion.getOffset() || visibleLength < linkRegion.getLength()) {
+							linkRegion = new Region(visibleStart, visibleLength);
+						}
+					} else {
+						// whole link is not visible - outside model coverage (visible region) or inside folded areas
+						continue;
```
RedeemerSK, line 150: This changes the current behavior that would always add link `StyleRange` to `TextPresentation` even though it will eventually be ignored inside TextViewer code (unles it gets modified before reaching that code). Which also means that all `ITextPresentationListener`s registered on viewer would find the link `StyleRange` there and possibly do who knows what to it (or whole text presentation).
I'm pointing this out to have this decision evaluated.

RedeemerSK: Re-evaluated this decision myself and made a new commit that brings back old behavior - now again `TextPresentation` is always sent even if it will get ignored inside `TextViewer` code because link completely falls out of the visible areas of the document.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/documentLink/LSPDocumentLinkPresentationReconcilingStrategy.java
```diff
@@ -101,32 +106,47 @@ private void underline(@Nullable List<DocumentLink> links) {
 				int length = end - start;
 				final var linkRegion = new Region(start, length);
 
-				// Update existing style range with underline or create a new style range with
-				// underline
-				StyleRange styleRange = null;
-				StyleRange[] styleRanges = viewer.getTextWidget().getStyleRanges(start, length);
-				if (styleRanges != null && styleRanges.length > 0) {
-					// It exists some styles for the range of document link, update just the
-					// underline style.
```
mickaelistria, line 110: Isn't the equivalent logic missing now?
Maybe `mergeStyleRange` should be called later instead of `add...` or `replace...` ?

RedeemerSK: This is now covered with 'else' segment starting @ line 138

rubenporras: If you do not mind, I will leave the review of this PR to @mickaelistria. @RedeemerSK , I think you need to rebase the PR for the builds to be successful.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/documentLink/LSPDocumentLinkPresentationReconcilingStrategy.java
```diff
@@ -101,32 +106,47 @@ private void underline(@Nullable List<DocumentLink> links) {
...
-					viewer.changeTextPresentation(presentation, false);
+					styleRanges = viewer.getTextWidget().getStyleRanges(start, length);
+					if (styleRanges != null && styleRanges.length > 0) {
+						// There are some styles for the range of document link, first update the underline style.
+						for (StyleRange s : styleRanges) {
+							s.underline = true;
+						}
+						// Then overlay on top of whole-region style range
+						presentation.replaceStyleRanges(styleRanges);
+					}
```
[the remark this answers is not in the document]
RedeemerSK: You're right. In between all other last minute changes to simplify the code I missed this rather obvious one.
At least the tests are now in place hopefully providing some  assurance it will still work in case anyone ever decides to do this change.

PR #1232, opened 2025-03-20
Fix #1231: Add missing separators to call hierarchy item labels

### LINKED ISSUE #1231: Fix call hierarchy labels: add missing spaces, separate method name from method container (e.g. class)
When displaying the call hierarchy, the tree entry labels are missing some spaces or other separator symbols between the method name and its container (e.g. a class). Here's an example:

Image

We should add some spaces or something like " - " between the method name and the container name.

travkin79:
Fix #1231 and add some test cases.


rubenporras (approved):

travkin79:
Hi @rubenporras,
Thank you for the review. Do you have an idea why the pr-merge check failed? There seems to be a time-out.

rubenporras:
> Hi @rubenporras, Thank you for the review. Do you have an idea why the pr-merge check failed? There seems to be a time-out.

I do not know but I have triggered it again and it passed.

PR #1234, opened 2025-03-28
refact: use TextPresentation#mergeStyles instead of hand-written merge

rubenporras:


rubenporras:
As suggested in https://github.com/eclipse-lsp4e/lsp4e/pull/1221#discussion_r2000434462

mickaelistria:
I didn't test it but code looks good enough to me. Assuming you have verified that the underline doesn't remove other styles, let's then merge.

rubenporras:
I did not verify it, I assumed that the tests where covering it, as this was part of the comments on the previous PR. I do not have time to do manual testing, so if you are not sure, I would rather drop the PR.

mickaelistria:
I'm not sure whether it's covered by tests or not. I think at this stage, let's keep the PR open until someone can verify it and whenever it's confirmed to be working as expected we'll merge.

rubenporras:
I am not getting to test this, so I close it to save resources.

PR #1253, opened 2025-04-23
refact: add LSPEclipseUtils.toRange()

sebthom:


rubenporras (approved):

rubenporras:
@mickaelistria , the dash license check has been failing for a while in different PRs, which looks like an infrastructure problem. Given that this PR (or others) do not modify any license or any license headers, may I merge it?

mickaelistria:
Indeed, Eclipse infra is facing its biggest and longest failure for a long time right now: https://www.eclipsestatus.io/
While I'm not too worried about the license check and would be comfortable merging despite this issue; I'm more worrying about the "continuous-integration/jenkins" build currently unable to validate the change.
Note that if you merge right now, the CI would still not be working and wouldn't be able to produce a newer snapshot. So there is not much time saved in merging now rather than later. However, if you're really willing to merge, as other CI builds (GitHub ones) seem to validate the PR, I don't think it would be a bad thing either. My current policy is to wait for jenkins build; but YMMV and as long a risk is assessed to be low enough, it's acceptable to merge.
To sum it up, as you wish ;)

rubenporras:
Thanks for the explanation, I will wait then.

sebthom:
We should not merge before Jenkins is back, as we currently don't get snapshot builds with the merged changes.

Issue #626, opened 2023-05-02
LanguageClientImpl is required but is not available


vgcpge:
Extension point `org.eclipse.lsp4e.languageServer` requires a class given in the field `clientImpl` to extends `org.eclipse.lsp4e.LanguageClientImpl`. However `org.eclipse.lsp4e` package is internal and an attempt to extend the class leads to warnings.

Thus, extensions for lsp4e are unable to customize the client behavior or schema in a legitimate manner.

There are following solutions available:
1. Require `clientImpl` to implement `org.eclipse.lsp4j.services.LanguageClient` interface instead.
1.1 (Optional) Provide another basic implementation of  `org.eclipse.lsp4j.services.LanguageClient` for extensions to reuse.
2. Make `org.eclipse.lsp4e` an API package (too many classes in that package to expose them as API).
3. Move `org.eclipse.lsp4e.LanguageClientImpl` to another package (backward incompatible change).

Context: I've been trying to implement [partial result progress with data transfer](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#partialResults) in an extension and default client does  partial progress reporting, but does not expose the data collected.

sebthom:
@rubenporras @mickaelistria any thoughts on this. I am faced with the same issue and so far simply ignore the fact that it is an internal/restricted class.

mickaelistria:
The overall proposal seems fine; here his some nitpicking bits of mine (which of course are not necessarily best; as it's about impossible to define what "best" mean when designing APIs; as one cannot really know all the possible use-cases already in place)

* HasLanguageServerPropertyTester is not expected to be referenced directly from Java code but is supposed to be used with the property tester in plugin.xml.
* LanguageServiceAccessor is kind of internal, as `LanguageServers` is now a better entry point to get the right LS
* `LaunchConfigurationStreamProvider` should IMO be internal as it's not really meant to be extensible; it's more an internal shortcut for quickly testing a LS.
* LoggingStreamConnectionProviderProxy should also be internal as it's about the internal Logging capabilities of LSP4E that are not supposed to be API
* LanguageServersRegistry and ContentTypeTo* classes are not meant to be used by extenders AFAIK

sebthom:
[...]

PR #1254, opened 2025-04-26
feat: move LanguageClientImpl to a package which is exported without restrictions

rubenporras:
for #626


rubenporras:
If you agree with this, I will investigate why git does consider the move a deletion and addition, or if you know, you could tell me ;)

mickaelistria:
The schema/languageServer.exds needs to be updated.
Maybe we should consider renaming to something like `DefaultLanguageClient` or something that sounds fine to extend.

rubenporras:
> The schema/languageServer.exds needs to be updated. Maybe we should consider renaming to something like `DefaultLanguageClient` or something that sounds fine to extend.

Done both.

I have also created two commits so that the first one can register the move properly and keep the git history.

rubenporras:
@mickaelistria , @sebthom , any further comments? Otherwise I would merge it.

FYI: I will be on holidays until next Wednesday

sebthom:
> @mickaelistria , @sebthom , any further comments? Otherwise I would merge it.

@rubenporras there are two pending comments from me.

PR #1255, opened 2025-04-27
Basic support for onTypeFormatting

mickaelistria:


org.eclipse.lsp4e/src/org/eclipse/lsp4e/format/DocumentFormatOnTypeAutoEditStrategy.java
```diff
@@ -0,0 +1,67 @@
...
+						try {
+							int fromOffset = LSPEclipseUtils.toOffset(edit.getRange().getStart(), document);
+							int toOffset = LSPEclipseUtils.toOffset(edit.getRange().getEnd(), document);
+							if (fromOffset == command.offset) {
+								command.text += edit.getNewText();
+							} else {
+								command.addCommand(fromOffset, toOffset - fromOffset, edit.getNewText(), null);
+							}
+						} catch (BadLocationException ex) {
+							ILog.get().error(ex.getMessage(), ex);
```
rubenporras, line 58: I think in other places in LSP4E we use `LanguageServerPlugin.logError(ex)`, would it make sense to do the same?

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServerWrapper.java
```diff
@@ -1040,6 +1041,12 @@ void registerCapability(RegistrationParams params) {
 				serverCapabilities.setTypeHierarchyProvider(Boolean.TRUE);
 				addRegistration(reg, () -> serverCapabilities.setTypeHierarchyProvider(typeHierarchyBeforeRegistration));
 				break;
+			case "textDocument/onTypeFormatting": //$NON-NLS-1$
+				final var onTypeFormattingBeforeRegistration = serverCapabilities.getDocumentOnTypeFormattingProvider();
```
rubenporras, line 1045: given that we are using var to avoid the types, and o to have a short name for an object, I think the variables does not give much value, what about removing them and using
```
				serverCapabilities.setDocumentOnTypeFormattingProvider(reg.getRegisterOptions() instanceof DocumentOnTypeFormattingOptions opts ? opts : null);
				addRegistration(reg, () -> serverCapabilities.setDocumentOnTypeFormattingProvider(serverCapabilities.getDocumentOnTypeFormattingProvider()));
```
?

rubenporras (approved):

PR #1258, opened 2025-05-03
skip linked editing update on selection (size) changes

danthe1st:
Fixes eclipse-wildwebdeveloper/wildwebdeveloper#1844

When selecting a range of characters spanning different linked editing ranges with the keyboard, it resets the selection moving the cursor to the end of the selection.

This PR fixes that issue by skipping that processing of linked editing ranges when the size of the selection changes as these don't seem to be necessary for the linked editing feature.

I don't know whether this is the best way of solving this issue (for example, it might be possible to somehow change the linked editing logic to not modify the selection and then reset it or to explicitly restore the cursor/caret position) but I wanted to at least show one way of doing it.

Alternatively, [applyLinkedEditing](https://github.com/eclipse-lsp4e/lsp4e/blob/0480a5b13ba86af99477812c1cbe9f316758154f/org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/linkedediting/LSPLinkedEditingReconcilingStrategy.java#L216) could be changed to restore both the selection and the caret by [creating a selection with the end before the start](https://github.com/eclipse-platform/eclipse.platform.ui/discussions/2958#discussioncomment-13047053) similar to the one used in the test I added.

### How to test

- Run Eclipse with wildwebdeveloper.xml and the modified version of lsp4e (that's easiest to reproduce it with)
- Make a selection on the editor across multiple XML tags but start at the end of the selection (i.e. selecting backwards such that the caret should be at the beginning after the selection)
- Notice that the caret moves to the end of the selection.

If automated tests are necessary for fixes like that, I'd still have to figure out how to access the data there (`LinkedEditingTest` didn't seem to be that useful there until now).

### DIFF
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/linkedediting/LSPLinkedEditingReconcilingStrategy.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/linkedediting/LSPLinkedEditingReconcilingStrategy.java
@@ -152,7 +152,7 @@ public class LSPLinkedEditingReconcilingStrategy extends LSPLinkedEditingBase
 	}
 
 	private void updateLinkedEditing(ISelection selection) {
-		if (selection instanceof ITextSelection textSelection) {
+		if (selection instanceof ITextSelection textSelection && textSelection.getLength() == 0) {
 			updateLinkedEditing(textSelection.getOffset());
 		}
 	}
--- a/org.eclipse.lsp4e.jdt/META-INF/MANIFEST.MF
+++ b/org.eclipse.lsp4e.jdt/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: JDT Integration for LSP4E
 Bundle-SymbolicName: org.eclipse.lsp4e.jdt;singleton:=true
-Bundle-Version: 0.13.6.qualifier
+Bundle-Version: 0.13.7.qualifier
 Export-Package: org.eclipse.lsp4e.jdt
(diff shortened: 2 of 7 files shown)


mickaelistria:
With the proposed code, I'm afraid that some cases can still be troublesome in case selection is changed programatically or via some advanced text control operations, leading to selection size not being changed while the selection still spans over several ranges.
I think we could make it even simpler and more reliable by just updating the selection when selection.getLength() == 0, and skipping otherwise.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/linkedediting/LSPLinkedEditingReconcilingStrategy.java
```diff
@@ -59,6 +59,7 @@ public class LSPLinkedEditingReconcilingStrategy extends LSPLinkedEditingBase
 	private @Nullable EditorSelectionChangedListener editorSelectionChangedListener;
 	private @Nullable Job highlightJob;
 	private @Nullable LinkedModeModel linkedModel;
+	private int lastSelectionLength;
```
mickaelistria, line 62: This field now seems useless. Can you please remove it?

danthe1st: Done. I was still working on a test.

danthe1st:
I now also added a test that ensures it is skipped properly (and used the chance to clean up some minor duplication in the existing tests).

rubenporras:
Hi @danthe1st ,

could you bump the version of the plug-in so that the jenkins continuous integration job passes?

Thanks

danthe1st:
> could you bump the version of the plug-in so that the jenkins continuous integration job passes?

Done.

PR #1260, opened 2025-05-05
[#1259] add lsp4e rename command

### LINKED ISSUE #1259: Refactorings -> Rename shows wrong key binding
As the rename command under Refactorings -> Rename in the generic text editors context menu uses the default Eclipse rename command with the ID `org.eclipse.ui.edit.rename`, a wrong key binding is shown:

Image

F2 is the key that has been bind to the rename command on resources:

Image

So F2 on in the generic text editor does not work.

Expected behavior:
The key shown in Refactorings -> Rename should be Alt+Shift+R (on Windows) or generally M2+M3+R

ghentschke:
and do not reuse the org.eclipse.ui.edit.rename command, because this leads to confusion in key bindings (F2 is associated with org.eclipse.ui.edit.rename and Alt+Shift+R is associated with the LSP based renaming of symbols)

fixes #1259


ghentschke:
image
 and
 
image

mickaelistria:
I'm not in favor of redefining a command for an operation that is already defined by Platform. The purpose of commands is exactly to make them reusable and to delegate to the most accurate handler. I'm more curious about why F2 is not working as expected (ie trigger the rename handler in LSP4E).
Editors all redefining a "rename" command instead of feeding Platform's one is somehow an anti-pattern.

ghentschke:
>  I'm more curious about why F2 is not working as expected (ie trigger the rename handler in LSP4E).

Yes. As I've seen that other editors like the old C/C++ editor have their own rename commands I suggested that there is no other way to resolve the issue that there are more than one key binding for a command. I'll try to solve that.

ghentschke:
AFAIK it's currently not possible to tell Eclipse to use the overwritten key bindings for a given command id:
> The most common reason the overwritten key binding is not shown in the context menu is that the menu entries are statically initialized and do not dynamically reflect key binding changes. Ensure your contextId matches, resolve conflicts, and restart Eclipse to address this issue.

ghentschke:
IMO the `org.eclipse.ui.edit.rename` command has a different scope. It's used to rename workspace resources and not symbols in editors. From that perspective we do not redefine the command. Instead reusing the command for a refactoring rename is not correct.

sebthom:
@mickaelistria @rubenporras Even though I was asked to review this, I don’t have any opinion on this. How should we proceed with this PR?

ghentschke:
Will be fixed by this platform.ui [PR #2969](https://github.com/eclipse-platform/eclipse.platform.ui/pull/2969)

PR #1263, opened 2025-05-08
Correct File URI for Outline

### LINKED ISSUE #1262: Document Outline called with wrong URI for FileStore input resulting in an empty Outline
When opening an editor with a IFileStore as input, the document outline function is called with the wrong URI. It is called with file:/<path> instead of file:///<path>. This in turn causes the document outline function to return an empty response, atleast for xml and json language servers that wildwebdeveloper provides. 

According to my somewhat limited debug efforts, its root cause is in probably the LSSymbolsContentProvider#OutlineViewerInput where you call file.toURI(), which will return the URI in the format file:/<path>. A quick fix would be to manipulate the URI there. Or to manipulating it before calling document outline function of the language server.

joeljohansson99:
fixes #1262

### DIFF
--- a/org.eclipse.lsp4e.test/META-INF/MANIFEST.MF
+++ b/org.eclipse.lsp4e.test/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: Tests for language server bundle (Incubation)
 Bundle-SymbolicName: org.eclipse.lsp4e.test;singleton:=true
-Bundle-Version: 0.15.26.qualifier
+Bundle-Version: 0.15.27.qualifier
 Fragment-Host: org.eclipse.lsp4e
 Bundle-Vendor: Eclipse LSP4E
 Bundle-RequiredExecutionEnvironment: JavaSE-17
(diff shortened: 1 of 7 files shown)


rubenporras:
Hi @joeljohansson99 ,

could you sign the legal agreement so that the ECA check passes?

also, would you be able to a test?

Kind Regards

joeljohansson99:
Hi @rubenporras, 
Sure I will sign the agreement.

Do you mean add a test? I looked at it and since OutlineViewerInput keeps the URI private I am not sure if its possible to check it. Also since you are having a mock server we cannot call the language server and check for a non-empty response.

An option would be to add an assert in the documentSymbol function of the mock server to check if the function is called with the correct URI. And then try to open an editor with a FileStore input. What do you think about that option? Or do you know any other way we could test this?

I will be at JCON next week, I will update the PR the week after.

sebthom:
the builds succeed now. 

@joeljohansson99 I can see that your ECA status is valid for your gmail address but not for your company address. You either need to sign the ECA with your company email or use your gmail address as author in your commits.

joeljohansson99:
Unfortunately I still have to wait for my legal department to review the ECA. They are taking their time but we are pushing. Sorry about the long delay.

joeljohansson99:
@sebthom, good news! I now signed the agreement. 
Let me know if you need anything else.

sebthom:
Thanks!

PR #1283, opened 2025-06-05
Synchronize markers updates

BoykoAlex:
Sending the same diagnostics from LS twice for example results in race conditions in `LSPDiagnosticsToMarkers`, namely the job for updating the Eclipse resource markers in `updateMarkers(...)` method. The job's scheduling rule is `null`. There is an expression for computing proper marker modification scheduling rule but the value of the resultant rule is `null`. Therefore, marker update jobs can run in parallel and result in some markers duplicated sometimes.
Fixing this cabbe a different scheduling rule, i.e. the resource but this might affect the build and other things. The other option is just sync block for updating the markers within the job body. The "sync block" solution seemed to work and is included in the PR.

### DIFF
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/diagnostics/LSPDiagnosticsToMarkers.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/diagnostics/LSPDiagnosticsToMarkers.java
@@ -82,7 +82,7 @@ public class LSPDiagnosticsToMarkers implements Consumer<PublishDiagnosticsParam
 	}
 
 	@Override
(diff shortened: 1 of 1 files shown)


sebthom:
@BoykoAlex you need to bump the LSP4E plugin version since it was just released.

rubenporras:
thanks

sebthom:
@BoykoAlex can you verify that reducing the synchronized block to only computing the markers and running the for loops calling `updateMarkers` unsynchronized still solves the issue you are facing?

mickaelistria:
@sebthom @BoykoAlex FWIW, I don't think it's worth the investigation at this moment. The current approach does fix the bug, and the amount and duration of marker updates make it probably neglect-able to use the resource as sync monitor.

PR #1287, opened 2025-06-20
fix: missing default HTML presenter in LSP text hover

ghentschke:
add statusFieldText leads constructor call which involves the default HTMLTextPresenter instead on null. Otherwise plain HTML code will be displayed in the hover.


rubenporras (approved):

rubenporras:
@ghentschke,  bump the plugin id?

PR #1294, opened 2025-07-06
Fix condition for calling rangeFormatting

fabrizioiannetti:
due to a refactor the condition that enables rangeFormatting was false for non-empty text selections. Restore the original logic.


org.eclipse.lsp4e/META-INF/MANIFEST.MF
```diff
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: Language Server Protocol client for Eclipse IDE (Incubation)
 Bundle-SymbolicName: org.eclipse.lsp4e;singleton:=true
-Bundle-Version: 0.18.24.qualifier
+Bundle-Version: 0.18.25.qualifier
```
sebthom, line 5: You also need to bump the version in the pom.xml

fabrizioiannetti:
Thanks for the hint, will fix that too.

Just to clarify what this is about: currently lsp4e will send a full file format request even if the client (e.g. CDT, or the equivalent I'm trying to port to corrosion) gives a non empty text selection (the modified section) and the range formatting is supported by the language server. 

From a UI point of view this is e.g. the case where in CDT the Save Actions->Format Source Code->Format Edited lines is selected.

The condition for the if clause was changed in commit 5a9f4612aa743a382ac32f8caabd757fef03a230.

PR #1300, opened 2025-07-14
Don't acquire lock on context while holding lock on LS wrapper object

iloveeclipse:
Fixes https://github.com/eclipse-lsp4e/lsp4e/issues/1299


org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServerWrapper.java
```diff
@@ -207,7 +207,7 @@ public void underlyingFileDeleted(IFileBuffer buffer) {
 	}
 
 	private static class LanguageServerContext {
-		boolean cancelled = false;
+		volatile boolean cancelled;
```
rubenporras, line 210: Thanks for the PR. What do you think of making this an AtomicBoolean as it was before (see commit ca55022f6d7441d1b90e462675be85e8dc1275b7). I think this is more clear that working with volatiles.

iloveeclipse: As you like, you are the code owner, IMO thee two options are equal and it is just matter of style.

iloveeclipse: So if you want `AtomicBoolean`, just say it :-)

rubenporras (approved):

rubenporras:
@iloveeclipse , I will be on holidays for three weeks after this Friday, do you need a release of LSP4E before I go on holidays?

iloveeclipse:
> @iloveeclipse , I will be on holidays for three weeks after this Friday, do you need a release of LSP4E before I go on holidays?

I would never say "no" to a new release :)
If there is no much overhead, it would be good. If not, I assume there is some nightly build I could try.

iloveeclipse:
https://github.com/eclipse-lsp4e/lsp4e/pull/1294 seem to be a nice candidate for a new release too.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServerWrapper.java
```diff
@@ -207,7 +208,7 @@ public void underlyingFileDeleted(IFileBuffer buffer) {
 	}
 
 	private static class LanguageServerContext {
-		boolean cancelled = false;
+		final AtomicBoolean cancelled = new AtomicBoolean(false);
```
sebthom, line 211: wouldn't it have been sufficient to make the boolean volatile plus removing the `synchronized(contextToStop) {` guard?

Issue #1305, opened 2025-07-17
DocumentLinks Underline Disappear on Document Change


sdaimwood:
Hello, in our editors the underline on document links renders when a document is first opened thanks to the `initialReconcile`, but all links disappear whenever a change is made to the document.

I've debugged and found this happens because this `reconcile` method intentionally does nothing currently: 

https://github.com/eclipse-lsp4e/lsp4e/blob/cc3fb63d9925b9acb57f299dd9511640d6482b8d/org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/documentLink/LSPDocumentLinkPresentationReconcilingStrategy.java#L181-L184

What's the reason for this current implementation? I couldn't find more detail in #231, but found when I added an `underline()` call in the empty `recocile()`, it works as expected. See the following gifs for before/after:

Before adding `underline()` in the `reconcile(DirtyRegion dirtyRegion, @Nullable IRegion subRegion)` method:

![Image](https://github.com/user-attachments/assets/58648835-058b-4e7f-b935-9705aa18d768)

After adding `underline()` in the `reconcile(DirtyRegion dirtyRegion, @Nullable IRegion subRegion)` method:

![Image](https://github.com/user-attachments/assets/716d313d-27b8-46f9-b881-15dfa3bc645f)

Is there something I'm missing in our TextEditor or ProjectionViewer extended classes to hit the other reconcile that does perform the underline? I can't seem to find a way to get into that method so would appreciate any help on that front if possible.

PR #1307, opened 2025-07-21
Call underline() for region-based reconcile also

sdaimwood:
Currently when a change is made in an `ExtensionBasedTextEditor` that uses document links, the links disappear. This is caused by the region-based `reconcile` method here intentionally doing nothing:

https://github.com/eclipse-lsp4e/lsp4e/blob/cc3fb63d9925b9acb57f299dd9511640d6482b8d/org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/documentLink/LSPDocumentLinkPresentationReconcilingStrategy.java#L181-L184

This change adds support for editors that call this `reconcile` method on change.

See Issue #1305 for more details and example gifs showing the problem.


mickaelistria:
I think this is a good 1st implementation, although the method seems to expect some finer grain processing (ie process particular regions, not whole document). Still the proposal here seems better than the former "do nothing" strategy.
And thanks for bumping the version!

PR #1309, opened 2025-08-06
feat: expose language server's InitializeResult in LanguageServerWrapper

joaodinissf:
Allows information like the server's name and version to be retrieved by accessing the InitializeResult.

The InitializeResult itself can be accessed via `getInitializeResultAsync()`.

For convenience, a dedicated `getServerInfoAsync()` method is also provided.

Further information about InitializeResult is available here: https://microsoft.github.io/language-server-protocol/specifications/lsp/3.18/specification/#initializeResult

### DIFF
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServerWrapper.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServerWrapper.java
@@ -98,6 +98,7 @@ import org.eclipse.lsp4j.Registration;
 import org.eclipse.lsp4j.RegistrationParams;
 import org.eclipse.lsp4j.SelectionRangeRegistrationOptions;
 import org.eclipse.lsp4j.ServerCapabilities;
+import org.eclipse.lsp4j.ServerInfo;
 import org.eclipse.lsp4j.TextDocumentSyncKind;
 import org.eclipse.lsp4j.TextDocumentSyncOptions;
 import org.eclipse.lsp4j.TypeHierarchyRegistrationOptions;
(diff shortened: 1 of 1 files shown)


org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServerWrapper.java
```diff
@@ -895,6 +898,10 @@ protected CompletableFuture<LanguageServer> getInitializedServer() {
 		return CompletableFuture.completedFuture(context.languageServer);
 	}
 
+	public ServerInfo getInitializedServerInfo() {
+		return initializedServerInfo;
```
mickaelistria, line 902: Should it 1st join on initializeFuture being complete?
Or maybe return a `CompletionFuture<ServerInfo>` with `return initializeFuture.thenApply(_ -> this.initializeServerInfo)` ?

mickaelistria (commented):
Can you please elaborate on a possible use-case for it?

joaodinissf:
Thanks for the speedy review, @mickaelistria. Really appreciate it.

The main use case for this PR is that, as far as I could tell, there's no existing way to query the language server's version. However, this is returned [as part of the InitializeResult](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.18/specification/#initializeResult), and therefore could be made available with the change in this PR.

If this looks good to you, I'll update the commit/PR with this explanation + fix the build failure + refactor the getter to use a `CompletableFuture`, as you suggested, as soon as I get the chance.

mickaelistria:
Shouldn't we keep the whole InitializeResult object instead and have a `getInitializeResult()` ? So if later some other items are added in it, we wouldn't ened a new API to leverage them.

rubenporras (approved):

Issue #1265, opened 2025-05-22
BufferManager is not disconnect after use


jdneo:
In some cases, the LSPEclipseUtil will call connect() from the BufferManager:

https://github.com/eclipse-lsp4e/lsp4e/blob/main/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LSPEclipseUtils.java#L573

But it's not disconnect after that, which may potentially leak memory

mickaelistria:
Thanks for reporting. Could you please try submitting a PR remediating that?

PR #1323, opened 2025-09-07
[#1322] disconnect only files which are connected by LSPEclipseUtils

### LINKED ISSUE #1322: CDT-LSP Editor crashes when disconnectTextFileBuffer called in LanguageServerWrapper on Linux since LSP4E 0.26.6
When the clangd LS gets **restarted** due to a change in the `compile_commands.json` or the user has modified the LS settings in the preferences and a C/C++ file which is connected to the LS is being opened in the editor it crashes like this:

Image

I did some debugging and found out that it crashes since this PR #1266 

It seems to be the disconnectTextFileBuffer call. All documents in the file buffers getting disconnected and buffers will be cleared. When a file is still open in an editor, the LSP4E framework does the enable checks for a lot of editor providers. To check if the LS should be enabled for the given file, it tries to fetch the document. First attempt is to check via the file buffer manager. This fails because the prior stop disconnected all files from the buffer.
So it will be loaded temporary for the check. After the check the file buffer will be cleared which triggers a fireBufferDisposed event which calls the `org.eclipse.lsp4e.LanguageServerWrapper.LSFileBufferListener.bufferDisposed(IFileBuffer)` again which calls `disconnect(URI uri)` which calls startStopTimerTask()  in every check...

Something in this repeating calls triggers the behavior described above. It happens on some Linus distros only (for example Linux Mint, Ubuntu 22.04 LTS). I could not reproduce this behavior on a Windows 11 machine. It may depends on the used Java environment.

A solution could be to not remove documents from file buffer which are still opened in an editor.

Image

See also this cdt-lsp bug issue https://github.com/eclipse-cdt/cdt-lsp/issues/525

ghentschke:
to prevent a global file buffer manager disconnect when other consumers like editors rely on a connected file.
The reason to disconnect is to prevent memory leaks. Each caller on org.eclipse.core.filebuffers.IFileBufferManager.connect has to ensure to disconnect after usage.

fixes #1322, #1265

### DIFF
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LSPEclipseUtils.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LSPEclipseUtils.java
@@ -186,6 +186,7 @@ public final class LSPEclipseUtils {
 	private static final String MARKDOWN = "markdown"; //$NON-NLS-1$
 	private static final String MD = "md"; //$NON-NLS-1$
 	private static final int MAX_BROWSER_NAME_LENGTH = 30;
+	private static Set<IPath> connectedFiles = new HashSet<IPath>();
 
 	private LSPEclipseUtils() {
 		// this class shouldn't be instantiated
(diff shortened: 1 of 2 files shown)


org.eclipse.lsp4e/src/org/eclipse/lsp4e/LSPEclipseUtils.java
```diff
@@ -562,27 +563,46 @@ public static IDocument getDocument(@Nullable IResource resource) {
...
+		}
+		return document;
+	}
 
-			ITextFileBuffer buffer = bufferManager.getTextFileBuffer(resource.getFullPath(), LocationKind.IFILE);
-			if (buffer != null) {
-				document = buffer.getDocument();
+	public static void disconnectFromFileBuffer(@Nullable IPath path) {
+		synchronized(connectedFiles) {
+			if (path == null || !connectedFiles.contains(path)) {
```
rubenporras, line 593: could we use just `remove` and use the returned `boolean` to exit early instead of calling `contains` and then `remove`?

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LSPEclipseUtils.java
```diff
@@ -562,27 +563,46 @@ public static IDocument getDocument(@Nullable IResource resource) {
 		if (resource == null) {
 			return null;
 		}
+		IDocument document = null;
+		synchronized(connectedFiles) {
+			document = getExistingDocument(resource);
 
-		IDocument document = getExistingDocument(resource);
+			if (document == null && resource.getType() == IResource.FILE) {
```
rubenporras, line 570: should we not skip the `connect` if `connectedFiles` already contains the path?

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LSPEclipseUtils.java
```diff
@@ -562,27 +563,46 @@ public static IDocument getDocument(@Nullable IResource resource) {
...
+				ITextFileBufferManager bufferManager = FileBuffers.getTextFileBufferManager();
+				if (bufferManager == null)
```
rubenporras, line 572: to be consistent with the rest of the codebase, could you add `{}` around the `return document` statement?

rubenporras (changes requested):

ghentschke:
@rubenporras any objections against a merge?

rubenporras (approved):
All fine, thanks.

PR #1326, opened 2025-09-11
[#1325] fix broken outline view for clangd LS

### LINKED ISSUE #1325: Outline view broken for clangd LS
The problem is that `documentURI = LSPEclipseUtils.toUri(path);` returns a project relative path and not a absolute path in the file system: e.g. `file:///MyProject/file.cpp` and not `file:///home/user/ws/MyProject/file.cpp` . The clangd LS needs an absoulte path in the filesystem.

_Originally posted by @ghentschke in https://github.com/eclipse-lsp4e/lsp4e/pull/1263#discussion_r2342209910_

ghentschke:
fixes #1325

### DIFF
--- a/org.eclipse.lsp4e.test/src/org/eclipse/lsp4e/test/outline/OutlineViewerInputTest.java
+++ b/org.eclipse.lsp4e.test/src/org/eclipse/lsp4e/test/outline/OutlineViewerInputTest.java
@@ -11,7 +11,6 @@
  *******************************************************************************/
 package org.eclipse.lsp4e.test.outline;
 
-import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertTrue;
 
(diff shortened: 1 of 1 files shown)


ghentschke:
Seems that `Missing requirement: repository raw:0.13.1.'SNAPSHOT'/format(n[.n=0;[.n=0;[-S]]]):0.13.1-SNAPSHOT requires 'org.eclipse.equinox.p2.iu; org.commonmark-gfm-tables 0.0.0' but it could not be found` is the problem here

sebthom:
This probably relates to the latest Eclipse 2025-09 release and the fact that the target file https://github.com/eclipse-lsp4e/lsp4e/blob/main/target-platforms/target-platform-latest/target-platform-latest.target always points to the latest eclipse repos - which are "unstable" since the provided artifacts change with each Eclipse release. That's why in tm4e we don't do this anymore, we point to explicit releases.

ghentschke:
Hm, Eclipse can resolve all plug-ins locally:
grafik

rubenporras (approved):

rubenporras:
I did not have any plans so far. I can just do one today or any other day that suits you. Just let me know.

ghentschke:
@rubenporras  Thank you very much!

rubenporras:
@ghentschke thanks to you for fixing the problem!

PR #1337, opened 2025-10-08
Fix format whole doc when no selection is given

mickaelistria:
and only range formatting is available

### DIFF
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/format/LSPFormatter.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/format/LSPFormatter.java
@@ -78,7 +78,7 @@ public class LSPFormatter {
 		final var rangeParams = new DocumentRangeFormattingParams();
 		rangeParams.setTextDocument(docId);
 		rangeParams.setOptions(formatOptions);
-		final boolean isFullFormat = textSelection.isEmpty();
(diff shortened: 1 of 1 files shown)


org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/format/LSPFormatter.java
```diff
@@ -78,7 +78,7 @@ public static DocumentRangeFormattingParams getRangeFormattingParams(IDocument d
 		final var rangeParams = new DocumentRangeFormattingParams();
 		rangeParams.setTextDocument(docId);
 		rangeParams.setOptions(formatOptions);
-		final boolean isFullFormat = textSelection.isEmpty();
+		final boolean isFullFormat = textSelection.isEmpty() || textSelection.getLength() == 0;
```
garretwilson, line 81: I have zero knowledge of this codebase, but from the looks of it I would think that `isEmpty()` is meant to be a semantic method that checks if the text selection is empty, like `String.isEmpty()` checks to see if the string length is zero. So shouldn't the real fix be to update the `textSelection.isEmpty()` method to check for the length being zero? Otherwise there will be bugs like this all over the place, for anything that uses `textSelection`.

Again I don't know anything about the codebase, or where `textSelection` comes from, but just looking at this snippet, I think this is is question worth asking. Maybe there's a reasonable answer why it's being done this way.

PR #1340, opened 2025-10-14
[#1339] add preference to enable content assist trigger on semicolon

### LINKED ISSUE #1339: Do not invoke completion proposal on statement terminator
When using the completion proposals on typing (code completion/ content assist):

Image

the proposals are still invoked when a statement terminator (like `;`) has been entered.
When pressing enter, the proposal is used an not a new line has been added. See this cdt-lsp issue: https://github.com/eclipse-cdt/cdt-lsp/issues/550

The proposal is valid since in C++ its allowed to start the next statement after a space (see https://github.com/clangd/clangd/issues/2508#issuecomment-3344772007) 

IMO in 90% the user wants to enter a new line after a ;

ghentschke:
fixes #1339

### DIFF
--- a/org.eclipse.lsp4e/plugin.properties
+++ b/org.eclipse.lsp4e/plugin.properties
@@ -22,6 +22,7 @@ codeactions.menu.label=Code Actions
 languageservers.preferences.page=Language Servers
 languageservers.preferences.folding.page=Folding
 languageservers.preferences.formatter.page=Formatter
+languageservers.preferences.contentassist.page=Content Assist
 languageservers.preferences.logging.page=Logs
 notification.category.label = LSP
 notification.event.label = LSP Notification
--- a/org.eclipse.lsp4e/plugin.xml
+++ b/org.eclipse.lsp4e/plugin.xml
@@ -138,6 +138,12 @@
             id="org.eclipse.lsp4e.preferences.formatter"
             name="%languageservers.preferences.formatter.page">
       </page>
+      <page
+            category="org.eclipse.lsp4e.preferences"
+            class="org.eclipse.lsp4e.ui.ContentAssistPreferencePage"
+            id="org.eclipse.lsp4e.preferences.contentassist"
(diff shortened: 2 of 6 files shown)


ghentschke:
> JDT has "trigger chars" which allow users to configure more than 1 character in a freetext field. Could we mimic it for this use case?

Sure. I can replace the check box with a free text field to allow users to specify other, language specific,  characters.

> I wonder whether those "excluded" strings should instead be configured by language server. For example, what about } ? Or, if we're doing Pascal or Ada, those should exclude end; if doing bash, it should exclude fi ?

I think this is a little complicated. It was my first idea as well and I posted a clangd issue. But this [response](https://github.com/clangd/clangd/issues/2508#issuecomment-3344772007) pushed it back to the editor side. In C++ it is allowed to write a new statement after `;` without any space. So the LS works correctly when returning a proposal after a `;`. But for many users its at least annoying to get a proposal pop-up when they just want to enter a new line. The proposal is not shown in VS code.

mickaelistria:
I'm genuinely wondering how users would like (or not) get get `;` replaced by eg `;\t\t\n` when tying it as last char of a line... Maybe that would actually please user to have such auto-edit in the LS and it would resolve the story decently.

ghentschke:
That's exactly what the on-type-formatting does. But that's a different story.

ghentschke:
Please take look of the original cdt-lsp issue https://github.com/eclipse-cdt/cdt-lsp/issues/550. There is a GIF that explains the problem during typing. 

The on-type-formatting kicks in after the new line has been applied. But there is no new line applied after pressing the enter key. Instead, the proposal i being applied which is not wanted at that point.

mickaelistria:
> The on-type-formatting kicks in after the new line has been applied.

The onTypeFormatting notification could probably kick it anytime. Could the Language Server trigger it on pressing `;` to automatically add a new line and indentation?

PR #1342, opened 2025-10-16
[#1341] trigger onTypeFormatting on trigger char only

### LINKED ISSUE #1341: Trigger onTypeFormatting on trigger chars only
The onTypeFormatting reconciler should be triggered on trigger chars from the LS only. See also https://github.com/eclipse-cdt/cdt-lsp/issues/556

ghentschke:
- prevents onTypeFormatting on copy/paste actions in the editor.

fixes #1341


ghentschke:
@mickaelistria If you don't have any objections, I would like to merge this PR as it fixes this [bug](https://github.com/eclipse-cdt/cdt-lsp/issues/556) in cdt-lsp

PR #1347, opened 2025-10-22
ci: update workflows (add codeql, use MacOS 15 runner, re-enable Linux)

sebthom:
- upgrade from deprecated MacOS 13 Intel runner to MacOS 15 Intel runner
- re-enable previously disabled Linux runners
- add CodeQL workflow
- update URLs to workflow reference documentation
- replace -Dtycho.disableP2Mirrors=true with -Declipse.p2.mirrors=false

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,4 +1,4 @@
-# https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file
+# https://docs.github.com/en/code-security/dependabot/working-with-dependabot/dependabot-options-reference
 version: 2
 updates:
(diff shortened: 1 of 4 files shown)


rubenporras:
Hi @sebthom ,

why are we replacing -Dtycho.disableP2Mirrors=true with -Declipse.p2.mirrors=false? Is the problem that lead to disabling mirrors no longer an issue?

sebthom:
> Hi @sebthom ,
> 
> why are we replacing -Dtycho.disableP2Mirrors=true with -Declipse.p2.mirrors=false? Is the problem that lead to disabling mirrors no longer an issue?


It is the same setting but the logic of the new flag is reversed.

rubenporras:
right :)

PR #1364, opened 2025-11-04
fix: NPE when trying to open Quickfix in Problems view

FlorianKroiss:
+ Add null-check for `org.eclipse.lsp4j.CodeAction.getDiagnostics()`


rubenporras (changes requested):
Can you bump the plugin version?

FlorianKroiss:
@rubenporras I bumped the versions manually, but is there an automated way of doing this?

rubenporras (approved):

PR #1365, opened 2025-11-05
feat: Make `AdjustIndentation` the default `insertTextMode` for completion proposals without explicit `insertTextMode`

FlorianKroiss:
[VS Code](https://github.com/microsoft/vscode-languageserver-node/blob/cd32d3e505e8ec8df660bc8e328d5f650673c2d8/client/src/common/completion.ts#L90) also uses this as the default


mickaelistria:
No opinion nor objection, let's merge and see ;)

PR #1370, opened 2025-11-06
feat: Remove `final` modifier from *Message methods in `DefaultLanguageClient`

### LINKED ISSUE #1367: Make all methods in `DefaultLanguageClient` non-final
[`DefaultLanguageClient`](https://github.com/eclipse-lsp4e/lsp4e/blob/ca8736e2c528af5a53baa9f8ae25e30103857224/org.eclipse.lsp4e/src/org/eclipse/lsp4e/client/DefaultLanguageClient.java#L61) is intended to be subclassed to create a custom language client.
But some methods in this class are marked as `final`. My use-case is that I need to override the `logMessage` method, which is also marked as final.

If there is no particular reason for these methods to be final I would propose to remove the final keyword from all methods in this class.

FlorianKroiss:
This makes it easier for developers using LSP4E to create a language client with custom behavior.

Resolves #1367

### DIFF
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/client/DefaultLanguageClient.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/client/DefaultLanguageClient.java
@@ -96,12 +96,12 @@ public class DefaultLanguageClient implements LanguageClient {
 	}
 
 	@Override
-	public final CompletableFuture<MessageActionItem> showMessageRequest(ShowMessageRequestParams requestParams) {
+	public CompletableFuture<MessageActionItem> showMessageRequest(ShowMessageRequestParams requestParams) {
 		return ServerMessageHandler.showMessageRequest(wrapper, requestParams);
 	}
 
 	@Override
-	public final void showMessage(MessageParams messageParams) {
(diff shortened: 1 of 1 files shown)


FlorianKroiss:
> Can you please provide more info about why this is necessary to change the logging behaviour in your case?

Our application has a Log View which displays log messages coming from all over the application, to make it easier for users to diagnose problems. This includes log messages coming the LSP4E Plugin. 

We are using a language server which is quite verbose during startup and publishes a lot of INFO messages which are handled by `DefaultLanguageClient.logMessage()`. I would like to lower their log-level to DEBUG to reduce the log noise before handing the messages to the logger, but I can't intercept the message because `logMessage` is final.

> Could it be that for your use case it could be better to allow customization of the ServerMessageHandler?

Making the logging behavior in `ServerMessageHandler` configurable would probably work, but a developer might only want to change the log level if the message comes from a specific LS or has certain content.

Given that every custom LanguageClient must be a subclass of `DefaultLanguageClient`, I thought that it's reasonable to make all of it's functionality overridable. For example, a developer might also want to change how `showMessage` is handled.

rubenporras (approved):

Issue #97, opened 2022-03-18
TimeoutException on "Open Declaration" for slow Language Server


sebthom:
I am trying to integrate a 3rd party language that unfortunately takes more then 500ms when responding to a "Open Declaration" request for for large projects.

So when putting the cursor on a method in the source code and hitting F3 to navigate to the declaration the UI shows:
![image](https://user-images.githubusercontent.com/426959/159085032-7b63dc48-ccc6-4310-812b-56bc2456a2f9.png)

And the following stacktrace is logged:
```python
java.util.concurrent.TimeoutException
	at java.base/java.util.concurrent.CompletableFuture.timedGet(CompletableFuture.java:1886)
	at java.base/java.util.concurrent.CompletableFuture.get(CompletableFuture.java:2021)
	at org.eclipse.lsp4e.operations.declaration.OpenDeclarationHyperlinkDetector.detectHyperlinks(OpenDeclarationHyperlinkDetector.java:101)
	at org.eclipse.ui.texteditor.HyperlinkDetectorRegistry$HyperlinkDetectorDelegate.detectHyperlinks(HyperlinkDetectorRegistry.java:81)
	at org.eclipse.jface.text.hyperlink.HyperlinkManager.findHyperlinks(HyperlinkManager.java:289)
	at org.eclipse.jface.text.hyperlink.HyperlinkManager.openHyperlink(HyperlinkManager.java:585)
	at org.eclipse.jface.text.TextViewer.doOperation(TextViewer.java:3895)
	at org.eclipse.jface.text.source.SourceViewer.doOperation(SourceViewer.java:1064)
	at org.eclipse.jface.text.source.projection.ProjectionViewer.doOperation(ProjectionViewer.java:1466)
	at org.eclipse.ui.texteditor.TextOperationAction.lambda$0(TextOperationAction.java:130)
	at org.eclipse.swt.custom.BusyIndicator.showWhile(BusyIndicator.java:74)
	at org.eclipse.ui.texteditor.TextOperationAction.run(TextOperationAction.java:130)
	at org.eclipse.jface.action.Action.runWithEvent(Action.java:474)
	at org.eclipse.jface.commands.ActionHandler.execute(ActionHandler.java:121)
	at org.eclipse.ui.internal.handlers.E4HandlerProxy.execute(E4HandlerProxy.java:97)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
[...]

PR #1376, opened 2025-11-10
fix: Avoid UI freezes on Open Declaration with slow lang servers

sebthom:
Without the PR we currently wait up to 800ms which is for some LS too short (esp. when requests pile up) so the operation may consistently fail with: 

159085032-7b63dc48-ccc6-4310-812b-56bc2456a2f9

Since the wait happens in the UI thread the timeout cannot be extended without making the UI freeze longer.

With this PR the UI thread will be blocked for max 200ms (or shorter if we agree on). If the LS does not respond in that time, a `DeferredOpenDeclarationHyperlink` instance is returned which asynchronously opens the link once LS responded. Opening the link will be dismissed if the the editor was closed in the meantime, the document modified or the response took longer than 5 seconds.

Addresses #97

### DIFF
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/internal/DocumentOffsetAsyncCache.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/internal/DocumentOffsetAsyncCache.java
@@ -1,0 +1,141 @@
+/*******************************************************************************
+ * Copyright (c) 2025 Vegard IT GmbH and others.
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *  Sebastian Thomschke (Vegard IT GmbH) - initial implementation.
+ *******************************************************************************/
+package org.eclipse.lsp4e.internal;
+
+import java.time.Duration;
+import java.util.Collections;
+import java.util.Map;
+import java.util.WeakHashMap;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
+import java.util.concurrent.TimeUnit;
+import java.util.function.Supplier;
+
+import org.eclipse.jdt.annotation.Nullable;
+import org.eclipse.jface.text.IDocument;
+import org.eclipse.jface.text.IDocumentExtension4;
+
+/**
+ * Generic, per-document+offset cache for asynchronous results that avoids
+ * starting the same work twice by sharing a single running task.
+ *
+ * <p>
+ * Features:
+ * <li>Weakly keys by {@link IDocument} to avoid memory leaks.
+ * <li>Per-document concurrent maps for thread-safe access from UI and
+ * background.
+ * <li>Eviction: TTL-based using {@link System#nanoTime()} and document-change
+ * invalidation when a stable modification stamp is available.
+ * <li>In-flight de-duplication: only one running task per document+offset.
+ * <li>Stale-result protection: if the document changes while a value is being
+ * computed, the result is delivered to callers but is not cached.
+ */
+public final class DocumentOffsetAsyncCache<V> {
+
+	private record Entry<V>(V value, long createdNanos, long docModStamp) {
+		boolean stale(final long ttlNanos, final long currentDocStamp) {
+			return System.nanoTime() - createdNanos > ttlNanos //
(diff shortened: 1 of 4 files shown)


org.eclipse.lsp4e/src/org/eclipse/lsp4e/internal/DocumentOffsetAsyncCache.java
```diff
@@ -0,0 +1,130 @@
...
+	public void invalidate(final IDocument doc) {
+		cache.remove(doc); // synchronizedMap handles its own locking
+		final var map = inFlight.remove(doc); // remove returns the per-doc map, if any
+		if (map != null) {
+			map.values().forEach(f -> f.cancel(true));
+		}
+	}
+
+	public void put(final IDocument doc, final int offset, final V value) {
+		final ConcurrentMap<Integer, Entry<V>> byOffset;
```
rubenporras:
```suggestion
		final Map<Integer, Entry<V>> byOffset;
```

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/declaration/OpenDeclarationHyperlinkDetector.java
```diff
@@ -93,22 +125,27 @@ public class OpenDeclarationHyperlinkDetector extends AbstractHyperlinkDetector
...
 	 *            the LSP locations
 	 */
-	private static Collection<LSBasedHyperlink> toHyperlinks(IDocument document, IRegion region,
-			String locationType, @NonNullByDefault({}) Either<List<? extends Location>, List<? extends LocationLink>> locations) {
-		if (locations == null) {
-			return Collections.emptyList();
+	private static List<LSBasedHyperlink> toHyperlinks(final IDocument doc, final IRegion region,
+			final List<LabeledLocations> locations) {
+		final var allLinks = new LinkedHashMap<Either<Location, LocationLink>, LSBasedHyperlink>();
+		for (final LabeledLocations locs : locations) {
```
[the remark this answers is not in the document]
sebthom: I don't understand your comment. Could you please rephrase it.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/declaration/OpenDeclarationHyperlinkDetector.java
```diff
@@ -58,32 +68,54 @@ public class OpenDeclarationHyperlinkDetector extends AbstractHyperlinkDetector
...
 			LanguageServerPlugin.logError(e);
 		} catch (InterruptedException e) {
 			LanguageServerPlugin.logError(e);
 			Thread.currentThread().interrupt();
 		} catch (TimeoutException e) {
-			LanguageServerPlugin.logWarning("Could not detect hyperlinks due to timeout after 800 milliseconds");  //$NON-NLS-1$
-		}
-		if (allLinks.isEmpty()) {
-			return null;
+			if (!canShowMultipleHyperlinks) {
```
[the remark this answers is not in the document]
sebthom: I added `DeferredOpenMultiDeclarationHyperlink` which will show this dialog if more than one link are available:

image

org.eclipse.lsp4e/src/org/eclipse/lsp4e/internal/DocumentOffsetAsyncCache.java
```diff
@@ -0,0 +1,141 @@
...
+	 * Returns cached value if present and valid; otherwise returns the single
+	 * running task or starts one via {@code supplier}. A value is valid if it has
+	 * not expired by TTL and (when stamps are available) matches the current
+	 * document stamp. Results computed for an older stamp are not cached.
+	 */
+	public CompletableFuture<V> computeIfAbsent(final IDocument doc, final int offset,
+			final Supplier<CompletableFuture<V>> supplier) {
+		// Fast path: return a completed future if a fresh value is already cached
+		final @Nullable V cachedNow = getNow(doc, offset);
+		if (cachedNow != null)
```
rubenporras: in this PR, ifs with only one statement do not have braces around the statement. I think LSP4E has the convention so far to always have braces, doesn't it? Can we have them here as well?

sebthom: AFAIK we do not use braces for single return statement. At least there are plenty of cases like this.

rubenporras (approved):

sebthom:
> @sebthom , do you plan to do more fixes soon or shall I create a new release with this and your last improvement? I think they are both very nice improvements from the user point of view.

@rubenporras I would actually appreciate a prompt release

PR #1381, opened 2025-11-12
feat: prompt user when attempting to format a read-only file

### LINKED ISSUE #1078: LSP4E context menu Format action should not be allowed in read-only documents
Currently, if you open a read-only file and try to perform any of the Format action, the contents of the file will appear to change.

Note that it is not possible to save the file with these modifications.

Nevertheless, it should not be possible to apply commenting to a read-only document, even just at the UI level, if the document is read only, the menu item should be disabled, as for the delete/copy actions in the screenshot.

image-2024-08-13-15-37-13-332

sebthom:
Implemented similar as in TM4E https://github.com/eclipse-tm4e/tm4e/pull/896

Trying to format a read-only file displays a prompt to make the file writable.

image

Fixes #1078


org.eclipse.lsp4e/src/org/eclipse/lsp4e/DocumentContentSynchronizer.java
```diff
@@ -287,7 +290,10 @@ private void formatDocument() {
 				var edits = requestFormatting(document, textSelection).get(lsToWillSaveWaitUntilTimeout(), TimeUnit.SECONDS);
 				if (edits != null) {
 					try {
-						edits.apply();
+						// Avoid applying edits to read-only documents
```
rubenporras: same here

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LSPEclipseUtils.java
```diff
@@ -1625,4 +1626,17 @@ public static boolean isReadOnly(final IResource resource) {
...
 	}
+
+	public static IStatus setWritable(final IFile file) {
+		ResourceAttributes attrs = file.getResourceAttributes();
+		if (attrs != null) {
+			try {
+				attrs.setReadOnly(false);
+				file.setResourceAttributes(attrs);
+			} catch (CoreException e) {
+				return e.getStatus();
```
rubenporras: this does not seem to be correct. From the javadoc of `getStatus()`:

> <b>IMPORTANT:</b><br>
	 * The result must NOT be used for logging, error reporting, or as a method
	 * return value, since...

rubenporras (approved):

PR #1384, opened 2025-11-12
feat: display line numbers in debugger stack frame elements

### LINKED ISSUE #627: [Debug] show line on stackframe element
Currently, the stackframe element only show the label. It would be useful to also show the line number if not already part of the label. So "Copy stack" would be more useful.

sebthom:
Fixes #627

image

### DIFF
--- a/org.eclipse.lsp4e.debug/src/org/eclipse/lsp4e/debug/presentation/DSPDebugModelPresentation.java
+++ b/org.eclipse.lsp4e.debug/src/org/eclipse/lsp4e/debug/presentation/DSPDebugModelPresentation.java
@@ -33,6 +33,7 @@ import org.eclipse.jface.viewers.IFontProvider;
 import org.eclipse.jface.viewers.LabelProvider;
 import org.eclipse.lsp4e.debug.DSPPlugin;
 import org.eclipse.lsp4e.debug.debugmodel.DSPDebugElement;
+import org.eclipse.lsp4e.debug.debugmodel.DSPStackFrame;
 import org.eclipse.lsp4e.debug.debugmodel.DSPThread;
 import org.eclipse.osgi.util.NLS;
 import org.eclipse.swt.SWT;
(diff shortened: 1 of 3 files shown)


org.eclipse.lsp4e.debug/src/org/eclipse/lsp4e/debug/presentation/DSPDebugModelPresentation.java
```diff
@@ -72,6 +72,26 @@ public String getText(Object element) {
...
+							file = source;
+						}
+						final String suffix = '(' + file + ":" + line + ')'; //$NON-NLS-1$
+						if (!endsWith(label, suffix)) {
+							label.append(' ');
+							label.append(suffix);
+						}
+					}
+				} catch (final Exception ex) {
+					// Ignore - if we cannot get a line number, skip the suffix
```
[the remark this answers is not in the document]
rubenporras: Actually, if we check for `source != null`, I think I would remove the try block and let whatever unexpected exception might happen just propagate

org.eclipse.lsp4e.debug/src/org/eclipse/lsp4e/debug/presentation/DSPDebugModelPresentation.java
```diff
@@ -72,6 +72,26 @@ public String getText(Object element) {
 		} else {
 			// Use default TODO should the entire default be copied here?
 			label.append(DebugUIPlugin.getDefaultLabelProvider().getText(element));
+
+			if (element instanceof DSPStackFrame frame) {
+				try {
+					final int line = frame.getLineNumber();
+					if (line > 0) {
+						final String source = frame.getSourceName();
```
rubenporras: `org.eclipse.lsp4j.debug.StackFrame.getSource()` is optional, so I think we need to check for `null`. Don't you think so?

sebthom: wow, you have eagle eyes! this was actually a bug in DSPStackFrame.

PR #1403, opened 2025-11-16
refactor(tests): Upgrade tests to JUnit 5

FlorianKroiss:
Migrate tests to JUnit5, mostly following the [migration guide](https://docs.junit.org/current/user-guide/#migrating-from-junit4)

PR #1405, opened 2025-11-16
refactor(tests): Use `@ParameterizedTest` instead of looping over test data

FlorianKroiss:
Followup to #1403


sebthom (changes requested):
See inline comment

PR #1407, opened 2025-11-18
fix: changing breakpoint condition has no immediate effect

### LINKED ISSUE #1406: Changing breakpoint condition has no immediate effect (with Wild Web Developer)
Using Wild Web Developer, with a basic JS file
```javascript
for (var i = 0; i < 1000; i++) {
  console.log(i);
}
```
I then put a breakpoint on the `console.log` line, and go to edit the condition with eg `(i % 37) == 9`
and creating a Node.js debug configuration for this file.
Debug using this launch
A breakpoint is hit with i == 46, perfect.
Then I change the condition to `(i % 37) == 2`.
and then hit "Continue" on the debugged application
EXPECTED: Stop at 76
GOT: Stop at 83

The change on the breakpoint condition got ignored (I suspect it didn't trigger a refresh of the breakpoint to the debug adapter)
WORKAROUND: disable and re-enable breakpoint after condition change, and then the new condition takes effect

sebthom:
Fixes #1406


mickaelistria (approved):
Verified, and it works well with Wild Web Developer debuggers.
Thanks!

PR #1408, opened 2025-11-18
feat: add `workspace/didChangeWatchedFiles` support

### LINKED ISSUE #547: Support for DidChangeWatchedFiles Notification
It would be really nice that LSP4e supports [DidChangeWatchedFiles Notification](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#workspace_didChangeWatchedFiles).

I need this feature for our Qute LS support which manages html files. The main idea is to validate closed HTML files (and not only opened HTML files) which can have some references between them.

When an HTML file is created or deleted (without opening it), I need to b enotified by this change and revalidate some HTML files which have references to the created / deleted HTML file.

sebthom:
Fixes #547

### DIFF
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServerWrapper.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServerWrapper.java
@@ -14,10 +14,11 @@
  *  Kris De Volder (Pivotal, Inc.) - dynamic command registration
  *  Tamas Miklossy (itemis) - bug 571162
  *  Rubén Porras Campo (Avaloq Evolution AG) - documentAboutToBeSaved implementation
+ *  Sebastian Thomschke (Vegard IT GmbH) - textDocument/completion, workspace/didChangeWatchedFiles support; bug fixes
  *******************************************************************************/
 package org.eclipse.lsp4e;
 
-import static org.eclipse.lsp4e.internal.NullSafetyHelper.castNonNull;
+import static org.eclipse.lsp4e.internal.NullSafetyHelper.*;
 
 import java.io.BufferedReader;
 import java.io.File;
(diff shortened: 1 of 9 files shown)


sebthom:
@angelozerr can you please check if this is what you need? You can download an update site with this PR from https://github.com/eclipse-lsp4e/lsp4e/actions/runs/19482961536/artifacts/4607414883

angelozerr:
@sebthom I am sorry https://github.com/jbosstools/jbosstools-quarkus has been give up, I will not have time to test your PR

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServerWrapper.java
```diff
@@ -1442,6 +1470,90 @@ private boolean isValid(@Nullable WorkspaceFolder wsFolder) {
...
+			final LanguageServer currentServer = context.languageServer;
+			if (currentServer != null) {
+				currentServer.getWorkspaceService()
+						.didChangeWatchedFiles(new DidChangeWatchedFilesParams(fileEvents));
+			}
+		}
+
+		private List<FileEvent> toFileEvents(final IResourceChangeEvent event) {
+			if (event.getType() != IResourceChangeEvent.POST_CHANGE || event.getDelta() == null) {
+				return new ArrayList<>();
```
rubenporras:
```suggestion
				return List.of();
```
to avoid creating a new list?

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServerWrapper.java
```diff
@@ -1225,6 +1285,20 @@ private void addRegistration(Registration reg, Runnable unregistrationHandler) {
 		}
 	}
 
+	synchronized void disableWatchedFiles() {
+		if (watchedFilesListenerInstalled && !fileSystemWatcherManager.hasFilePatterns()) {
```
rubenporras: I wonder if we can remove the field `watchedFilesListenerInstalled`. From what I see in add and remove resource change listener, it is safe to remove if it has not been added, and it is safe to add the same listener twice, just a bit more expensive because it needs to go over the list of all listeners. Given that this is removal/addition does not happen frequently, I would do away with the class field.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServerWrapper.java
```diff
@@ -314,6 +327,18 @@ private LanguageServerWrapper(@Nullable IProject project, LanguageServerDefiniti
...
+		try {
+			if (initialProject != null) {
+				final var loc = initialProject.getLocationURI();
+				if (loc != null)
+					watchedFilesBasePath = Paths.get(loc);
+			}
+		} catch (IllegalArgumentException ex) {
+			LanguageServerPlugin.logError(ex);
+		}
+		this.fileSystemWatcherManager = new FileSystemWatcherManager(watchedFilesBasePath);
```
rubenporras: To keep the LSW simple, would it make sense to just pass `IProject` to its constructor and move all the preceding code into that constructor?

org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServerWrapper.java
```diff
@@ -1442,6 +1516,119 @@ private boolean isValid(@Nullable WorkspaceFolder wsFolder) {
 
 	}
 
+	/**
+	 * Resource listener that translates Eclipse resource change events into LSP
+	 * file watch events and dispatches them if the language server is still active
+	 */
+	private final class WatchedFilesListener implements IResourceChangeListener {
```
rubenporras: move to its own class file in the internal package?

sebthom: For that LanguageServerWrapper is in the wrong package. I would not want to pollute the root package with a "random" WatchedFilesListener class file that is an internal listener implementation (like LSFileBufferListener or WorkspaceFolderListener) of LanguageServerWrapper. If LoC of LanguageServerWrapper is a concern we should better think about moving all the generic utility methods like getRootCauseMessage/getThrowableMessage/disconnectTextFileBuffer/supportsWorkspaceFolders etc. out of LanguageServerWrapper but I would do this in a dedicated PR focused on refactoring/cleanup of the class.

PR #1415, opened 2025-11-21
feat: language-server powered file/folder create, rename, delete

### LINKED ISSUE #823: Add support for willRenameFiles/willCreateFiles/willDeleteFiles

sebthom:
Project Explorer actions now call LSP willCreate/willRename/willDelete and apply returned edits when supported. Supports both files and folders; honors server filters (glob + file/folder kind). Ensures pending didChange is sent before didClose to preserve LSP event order during renames.

Fixes #823

### DIFF
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/rename/LSPFileOperationParticipantSupport.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/rename/LSPFileOperationParticipantSupport.java
@@ -12,13 +12,19 @@
 package org.eclipse.lsp4e.operations.rename;
 
 import java.nio.file.Path;
+import java.util.List;
 import java.util.Objects;
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
 import java.util.function.BiFunction;
 import java.util.function.Function;
 
 import org.eclipse.core.resources.IResource;
+import org.eclipse.core.runtime.CoreException;
+import org.eclipse.core.runtime.IStatus;
+import org.eclipse.core.runtime.Status;
 import org.eclipse.jdt.annotation.Nullable;
 import org.eclipse.lsp4e.LSPEclipseUtils;
 import org.eclipse.lsp4e.LanguageServerPlugin;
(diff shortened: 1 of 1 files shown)


org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/rename/LSPFileOperationParticipantSupport.java
```diff
@@ -0,0 +1,124 @@
...
+	private static final long FILE_OP_TIMEOUT_SECONDS = 10;
+
+	public static <P> @Nullable Change computePreChange(final String changeName, final P params,
+			final List<LanguageServerWrapper> servers,
+			final BiFunction<WorkspaceService, P, CompletableFuture<@Nullable WorkspaceEdit>> request) {
+
+		final List<CompositeChange> changes = servers.stream() //
+				.map(wrapper -> wrapper.execute(ls -> request //
+						.apply(ls.getWorkspaceService(), params)) //
+						.orTimeout(FILE_OP_TIMEOUT_SECONDS, TimeUnit.SECONDS).exceptionally(ex -> null)
```
rubenporras: can we log the exception?

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/rename/LSPCreateParticipant.java
```diff
@@ -0,0 +1,83 @@
...
+	private URI newURI = lateNonNull();
+	private List<LanguageServerWrapper> servers = lateNonNull();
+
+	@Override
+	public String getName() {
+		return "LSP4E Create"; //$NON-NLS-1$
+	}
+
+	@Override
+	protected boolean initialize(final Object element) {
```
rubenporras: It looks like we are not using the return value for anything, are we? (same goes for all participants), could we just return `void`?

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/rename/LSPDeleteParticipant.java
```diff
@@ -0,0 +1,81 @@
...
+	private List<LanguageServerWrapper> servers = lateNonNull();
+
+	@Override
+	public String getName() {
+		return "LSP4E Delete"; //$NON-NLS-1$
+	}
+
+	@Override
+	protected boolean initialize(final Object element) {
+		if (element instanceof IFile || element instanceof IFolder) {
```
rubenporras:
```suggestion
		if (element instanceof IResource res && (res instanceof IFile || res instanceof IFolder)) {
```
and then remove the manual cast?

Same for the other participants

rubenporras (changes requested):
Most of my comments are about style, but I think https://github.com/eclipse-lsp4e/lsp4e/pull/1415#discussion_r2555054227 needs to be addressed.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/rename/LSPFileOperationParticipantSupport.java
```diff
@@ -0,0 +1,144 @@
...
+			final IResource resource,
+			final Function<FileOperationsServerCapabilities, @Nullable FileOperationOptions> optionsProvider,
+			final BiFunction<WorkspaceService, P, CompletableFuture<@Nullable WorkspaceEdit>> request) {
+		return computePreChange(changeName, params, createFileOperationExecutor(resource, optionsProvider), request);
+	}
+
+	public static <P> @Nullable Change computePreChange(final String changeName, final P params,
+			final LanguageServerProjectExecutor executor,
+			final BiFunction<WorkspaceService, P, CompletableFuture<@Nullable WorkspaceEdit>> request) {
+		final CompositeChange[] changes = executor.collectAll((wrapper, ls) -> request //
```
rubenporras: I tihnk we can reorder a bit the block before the `join()` and improve indentation, plus have a better error handling:
```
    final CompositeChange[] changes = executor //
        .collectAll((wrapper, ls) -> request //
            .apply(ls.getWorkspaceService(), params) //
            .thenApply(edits -> edits == null || isEmptyEdit(edits) //
                ? (CompositeChange) null
                : LSPEclipseUtils.toCompositeChange(edits, wrapper.serverDefinition.label)) //
            .orTimeout(FILE_OP_TIMEOUT_SECONDS, TimeUnit.SECONDS) //
            .exceptionally(ex -> {
              String logHeader = "File operation pre-change '" + changeName; //$NON-NLS-1$
              if (ex instanceof TimeoutException) {
                LanguageServerPlugin.logWarning(logHeader + "' timed out for server: " //$NON-NLS-1$
                    + wrapper.serverDefinition.label + " after " + FILE_OP_TIMEOUT_SECONDS + " seconds"); //$NON-NLS-1$ //$NON-NLS-2$
              } else {
                LanguageServerPlugin.logError(logHeader + "' failed for server: " //$NON-NLS-1$
                    + wrapper.serverDefinition.label, ex);
              }
              return null;
            })) //```

what do you think of that?

File: org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/rename/LSPFileOperationParticipantSupport.java [hunk 5]
```diff
@@ -0,0 +1,144 @@
...
+				.orTimeout(FILE_OP_TIMEOUT_SECONDS, TimeUnit.SECONDS).exceptionally(ex -> {
+					LanguageServerPlugin.logWarning(
+							"File operation pre-change '" + changeName + "' failed or timed out for server: " //$NON-NLS-1$ //$NON-NLS-2$
+									+ wrapper.serverDefinition.label,
+							ex);
+					return null;
+				}).thenApply(edits -> edits == null || isEmptyEdit(edits) //
+						? (@Nullable CompositeChange) null
+						: LSPEclipseUtils.toCompositeChange(edits, wrapper.serverDefinition.label))) //
+				.join() //
```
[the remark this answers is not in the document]
rubenporras: After looking at it in more detail, yes, I think we should have the 12 lines. The reason is that according to `org.eclipse.ltk.core.refactoring.participants.RefactoringParticipant.createPreChange(IProgressMonitor)`, execeptions happen, they should be converted to a `CoreException`. Even if we would use join, we would still need a catch block for the unchecked exception, so we can as well use then get and catch the checked one.

Get will in addition force catching `InterruptedException`, but that is also probably good as well not to wait forever if the LS process dies. I am not so sure that this would throw `InterruptedException` and not some type of `IOException`, but probably not worth to deviate from the standard pattern.

rubenporras (approved):

rubenporras (approved):

sebthom:
> fyi: I would like to do a release before Friday so that we follow [eclipse.dev/simrel?file=wiki/SimRel/2025-12.md](https://eclipse.dev/simrel/?file=wiki/SimRel/2025-12.md) and then wait with new features until the 10th of December.

Ok, let's hold back merging the PR for now.

PR #1428, opened 2025-11-26
build: upgrade target file to use Eclipse 4.38

rubenporras:


target-platforms/target-platform-latest/target-platform-latest.target
```diff
@@ -13,10 +13,10 @@
     <location includeAllPlatforms="false" includeConfigurePhase="true" includeMode="planner" includeSource="true" type="InstallableUnit">
       <unit id="org.eclipse.ui.tests.harness" version="0.0.0"/>
       <unit id="org.eclipse.sdk.ide" version="0.0.0"/>
-      <repository location="https://download.eclipse.org/eclipse/updates/4.37/"/>
+      <repository location="https://download.eclipse.org/eclipse/wnload.eclipse.org/eclipse/updates/4.38-I-builds/I20251125-1800/"/>
```
FlorianKroiss: That looks like a copy/paste error

rubenporras:
That is great, don't worry, there is no rush.

rubenporras:
> I had a same/similar issue in TM4E and solved it by setting `<providerHint>junit5</providerHint>`

Thanks a lot, that was the solution.

PR #1429, opened 2025-11-26
fix: solve JDT Null Analysis Mismatch Error

rubenporras:


org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/symbols/LSPSymbolInWorkspaceDialog.java
```diff
@@ -230,10 +230,10 @@ private static List<WorkspaceSymbol> toWorkspaceSymbols(@Nullable List<? extends
 		return res;
 	}
 
-	static List<? extends WorkspaceSymbol> eitherToWorkspaceSymbols(
+	static List<@Nullable ? extends WorkspaceSymbol> eitherToWorkspaceSymbols(
 			final @Nullable Either<List<? extends SymbolInformation>, List<@Nullable ? extends WorkspaceSymbol>> source) {
 		return source == null //
-				? List.of()
```
sebthom: Does List.of not work?

rubenporras: Not for me, then I get the marker:

> Null type mismatch (type annotations): required '@NonNull List<@Nullable ? extends WorkspaceSymbol>' but this expression has type '@NonNull List<@NonNull WorkspaceSymbol>'

It looks to me that the JDT does not like than one is List.of() is guaranteed to be not null, and the other expression is `Nullable`.

I tried  `NullSafetyHelper.castNullable(List.of())` which also causes the JDT to complain.

It looks to me like a litimation of the JDT null analysis, or at least I do not understand why it is complaining.

sebthom: List.of has the contract that it only returns lists with non-null elements. But you can replace `List.of` with `Collections.emptyList()` and it should work.

PR #1440, opened 2025-12-01
feat: support linked inline editing on rename

### LINKED ISSUE #891: Rename symbol always opens a dialog
It would be nice to have inline editing (note - not asking for linked editing of all references ) so there is no extra dialog opened.

sebthom:
Fixes #891

### DIFF
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/rename/LSPInlineRenameLinkedMode.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/rename/LSPInlineRenameLinkedMode.java
@@ -1,0 +1,480 @@
+/*******************************************************************************
+ * Copyright (c) 2025 Vegard IT GmbH and others.
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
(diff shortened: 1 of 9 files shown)


org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/rename/LSPInlineRenameLinkedMode.java
```diff
@@ -0,0 +1,460 @@
...
+						final StyledText widget = viewer.getTextWidget();
+						if (widget == null || widget.isDisposed()) {
+							boolean success = false;
+							try {
+								document.replace(position.getOffset(), position.getLength(), originalName);
+								success = doRename(newName);
+							} catch (BadLocationException e) {
+								LanguageServerPlugin.logError(e);
+							} finally {
+								if (!success) {
```
rubenporras: I am a bit surprised by this. Since `doRename` catches already the expected exceptions, why do we need the success variable out of the try block so that we can use it in `finally`? Can we not the check just inside the try block, same as in lines 428-431?

rubenporras (changes requested):
could you bump the version numbers of the test plugin?

PR #1462, opened 2025-12-09
fix: handle race conditions accessing elements in DOM

rubenporras:
Fixed the browser to not appear to be 'ready' to set the body before the text to be set has been processed.

Otherwise, because the content is set asynchronously in a browser after the call to browser.setText, it can be that the first call to document.readyState is done before the browser has processed the request to modify the page, and therefore document.readyState returns true even if the page has no content. To fix this, we check that the document has at least one element before asking for document.readyState.


FlorianKroiss:
This feels like a symptomatic fix to me, because now we probably won't get the exceptions, because the JavaScript expressions will always evaluate to something, but in cases where we previously had the errors, we now fallback to a default width/height, which will probably not fit the entire content once its actually loaded.

Maybe I can find some time this week to also have a go at this bug

PR #1463, opened 2025-12-10
testBodyReplace

rubenporras:


org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/hover/FocusableBrowserInformationControl.java
```diff
@@ -183,7 +193,15 @@ public void setInput(@Nullable Object input) {
...
 				if (html != null && !html.isBlank()) {
-					super.setInput(styleHtml(html));
+					try {
+						Field f = BrowserInformationControl.class.getDeclaredField("fBrowser"); //$NON-NLS-1$
+						f.setAccessible(true);
+						Browser browser = (Browser) f.get(this);
+						Objects.requireNonNull(browser);
+						setTextInBrowser(browser, html);
+					} catch (NoSuchFieldException | SecurityException | IllegalArgumentException | IllegalAccessException e) {
+						super.setInput(html);
```
rubenporras: should never happen, but it is just a safety net. We should add some logging in this case.

PR #1496, opened 2026-02-01
refactor: Get rid of some deprecated APIs

FlorianKroiss:
+ Use `setComparator` instead of `setSorter`
+ Don't use deprecated URL constructor when opening a Browser
+ Refactor `DSPImages` to be more similar to `LSPImages`, to get rid of URL constructor usage
+ Bump minor version of Debug Plugin due to removed constants from DSPImages

### DIFF
--- a/org.eclipse.lsp4e.debug/src/org/eclipse/lsp4e/debug/DSPImages.java
+++ b/org.eclipse.lsp4e.debug/src/org/eclipse/lsp4e/debug/DSPImages.java
@@ -13,74 +13,58 @@
  *******************************************************************************/
 package org.eclipse.lsp4e.debug;
 
-import java.net.MalformedURLException;
 import java.net.URL;
 
+import org.eclipse.core.runtime.FileLocator;
+import org.eclipse.core.runtime.Path;
 import org.eclipse.core.runtime.Platform;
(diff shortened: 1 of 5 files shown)


org.eclipse.lsp4e.debug/src/org/eclipse/lsp4e/debug/DSPImages.java
```diff
@@ -13,74 +13,64 @@
...
-		fgIconBaseURL = Platform.getBundle(DSPPlugin.PLUGIN_ID).getEntry("/icons/"); //$NON-NLS-1$
-	}
-
-	private static final String T_TABS = "view16/"; //$NON-NLS-1$
-	@SuppressWarnings("unused") // none yet, leave for the future
-	private static final String T_OBJS = "obj16/"; //$NON-NLS-1$
-
-	public static final String IMG_VIEW_DEBUGGER_TAB = NAME_PREFIX + "debugger_tab.svg"; //$NON-NLS-1$
+	@Deprecated(forRemoval = true)
+	public static @Nullable ImageDescriptor DESC_TAB_DEBUGGER = null;
```
rubenporras: since this will break any consumer I think it is better to remove it and bump the second number of the plugin to make clear that there has been a breaking change. What do you think?

PR #1505, opened 2026-02-26
feat: Inlay hints with text edits can be applied when clicking on them

FlorianKroiss:
Inlay hints can have text-edits, which VS Code applies when double clicking on them. We already had support for executing commands when clicking on individual parts of the inlay hint. With this PR, we additionally check if there are text-edits and apply them, when the user clicks on it.

It looks like this in action:
https://github.com/user-attachments/assets/aa7bd610-e727-40bf-b326-b8ca468552c0

In the above example, the LS inferred the type of the right hand side expression and the user can accept the inferred type.

### DIFF
--- a/org.eclipse.lsp4e.test/META-INF/MANIFEST.MF
+++ b/org.eclipse.lsp4e.test/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: Tests for language server bundle (Incubation)
 Bundle-SymbolicName: org.eclipse.lsp4e.test;singleton:=true
-Bundle-Version: 0.16.12.qualifier
+Bundle-Version: 0.16.13.qualifier
 Fragment-Host: org.eclipse.lsp4e
 Bundle-Vendor: Eclipse LSP4E
 Bundle-RequiredExecutionEnvironment: JavaSE-21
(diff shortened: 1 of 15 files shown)


FlorianKroiss:
Labelled as draft, because I'm not entirely happy with it. I'm not sure if the LSP defines whether a command or the text edits takes precedence, when the user clicks on the inlay hint.

I also think that accepting the inlay hint with a single click makes it easy to trigger by accident.
Maybe we can say

- Apply text-edit, if there is one, on double click. In this case we should show a tooltip or something, to let the user know about this.  ( <- VS Code does this)
- Execute command, if there is one, on single click

FlorianKroiss:
@rubenporras I would be happy if you can find some time to review this 😄

rubenporras (approved):

rubenporras:
Sorry, I forgot to review the PR. I have rebased it get a new set of builds, if they pass, I will merge it

rubenporras:
@FlorianKroiss , can you bump the test plugin version? Or do we wait for your other PR to go first and rebase on top?

FlorianKroiss:
> Sorry, I forgot to review the PR. I have rebased it get a new set of builds, if they pass, I will merge it

No worries. 

Yes, we should merge https://github.com/eclipse-lsp4e/lsp4e/pull/1578 first and then merge this one.

PR #1517, opened 2026-04-09
Add preference option for non-buffered file handling in Xtext editors.

raghucssit:
see https://github.com/eclipse-lsp4e/lsp4e/issues/1500

### DIFF
--- a/org.eclipse.lsp4e/plugin.xml
+++ b/org.eclipse.lsp4e/plugin.xml
@@ -145,9 +145,7 @@
       <initializer
             class="org.eclipse.lsp4e.operations.linkedediting.LSPLinkedEditingBase$PreferenceInitializer">
       </initializer>
-      <initializer
-            class="org.eclipse.lsp4e.operations.rename.LSPInlineRenameLinkedMode$PreferenceInitializer">
-      </initializer>
+      <initializer class="org.eclipse.lsp4e.ui.LanguageServerPreferencePage$PreferenceInitializer"/>
    </extension>
 
    <extension point="org.eclipse.ui.genericeditor.foldingReconcilers">
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LSPEclipseUtils.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LSPEclipseUtils.java
@@ -95,6 +95,7 @@ import org.eclipse.lsp4e.internal.ResourceForUriCache;
 import org.eclipse.lsp4e.refactoring.CreateFileChange;
 import org.eclipse.lsp4e.refactoring.DeleteExternalFile;
 import org.eclipse.lsp4e.refactoring.LSPTextChange;
+import org.eclipse.lsp4e.ui.LanguageServerPreferencePage;
 import org.eclipse.lsp4e.ui.Messages;
 import org.eclipse.lsp4e.ui.UI;
 import org.eclipse.lsp4j.CallHierarchyPrepareParams;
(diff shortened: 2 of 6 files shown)


rubenporras:
Hi @raghucssit ,

I think the end user should not be interested in such flag. Could we instead enable it in the Activator of CoPilot plugin?

iloveeclipse:
> I think the end user should not be interested in such flag. Could we instead enable it in the Activator of CoPilot plugin?

The point is, it is needed by Xtext & Copilot (as of today), but provided by LSP4E. If there will be some other plugin that needs same preference, should there be another preference page? Surely not. I don't see a big problem in providing an option in LSP4E for this feature, if there will be no complains with the new code, it can be removed later.

@raghucssit : can you please attach a screenshot of the preference page? I do not have time to load this patch into  IDE, and code reading doesn't replace visual check.

Beside this, **I miss one important thing** in this PR: the new support should by **enabled by default** with this preference page, that is one of the main reasons to have the page at all - if the new support breaks something in LSP4E/clients of it, it should be possible to users to fix that without waiting for new release.

raghucssit:
> > I think the end user should not be interested in such flag. Could we instead enable it in the Activator of CoPilot plugin?
> 
> The point is, it is needed by Xtext & Copilot (as of today), but provided by LSP4E. If there will be some other plugin that needs same preference, should there be another preference page? Surely not. I don't see a big problem in providing an option in LSP4E for this feature, if there will be no complains with the new code, it can be removed later.
> 
> @raghucssit : can you please attach a screenshot of the preference page? I do not have time to load this patch into IDE, and code reading doesn't replace visual check.
> 
> Beside this, **I miss one important thing** in this PR: the new support should by **enabled by default** with this preference page, that is one of the main reasons to have the page at all - if the new support breaks something in LSP4E/clients of it, it should be possible to users to fix that without waiting for new release.

image

rubenporras:
Hi @raghucssit ,

would it be possible to follow on https://github.com/eclipse-lsp4e/lsp4e/issues/1500#issuecomment-4171051251? I think the user should not have this as an option on the UI. It is very hard to make the link between this UI option and copilot, so an user might unclick it and wonder why something breaks, and also, if something is malfunctioning, he will really not know what unchecking this option will do.

iloveeclipse:
> would it be possible to follow on [#1500 (comment)](https://github.com/eclipse-lsp4e/lsp4e/issues/1500#issuecomment-4171051251)? I think the user should not have this as an option on the UI. 

The point I have is that the preference management belongs to the plugin who provides the preference. 

As [I've mentioned above](https://github.com/eclipse-lsp4e/lsp4e/pull/1517#issuecomment-4237153696), if there will be another plugin or product that would like to provide access to this preference, we will have multiple places that manage same thing.

> It is very hard to make the link between this UI option and copilot, 

So should this single option be moved to a "Copilot & Xtext" preferences subtree of "Language Servers" preferences? It would be more clear that it is Copilot related. But I would not consider this as a "Copilot & Xtext only", it is added because of Copilot & Xtext, but it affects all "non-standard" editors and all "document" users of LSP4E.

> so an user might unclick it and wonder why something breaks, and also, if something is malfunctioning, he will really not know what unchecking this option will do.

If users "per occasion" click somewhere, they can always click on "Restore defaults" in the preferences.

rubenporras:
> > would it be possible to follow on [#1500 (comment)](https://github.com/eclipse-lsp4e/lsp4e/issues/1500#issuecomment-4171051251)? I think the user should not have this as an option on the UI.
> 
> The point I have is that the preference management belongs to the plugin who provides the preference.
> 
> As [I've mentioned above](https://github.com/eclipse-lsp4e/lsp4e/pull/1517#issuecomment-4237153696), if there will be another plugin or product that would like to provide access to this preference, we will have multiple places that manage same thing.

Why does it need to be that way? My main point is that we do not need to provide access to this preference in the UI. 

> > It is very hard to make the link between this UI option and copilot,
> 
> So should this single option be moved to a "Copilot & Xtext" preferences subtree of "Language Servers" preferences? It would be more clear that it is Copilot related. But I would not consider this as a "Copilot & Xtext only", it is added because of Copilot & Xtext, but it affects all "non-standard" editors and all "document" users of LSP4E.

The default integration of LSP4E is with Platform's Generic and Extensible editor, as announced in [the project page](https://projects.eclipse.org/projects/technology.lsp4e). There it is also mentioned "some code may be used as API to let integration be done with other Eclipse-based editors" which I think is the level we should stay on this feature. Maybe a preference was the wrong proposal on my side?

PR #1521, opened 2026-04-16
Make overlay icons in various views like the outline view customizable in client plug-ins

travkin79:
This PR makes overlay icon calculation customizable in client plug-ins (e.g. in CDT LSP) by introducing a new extension point for custom `SymbolIcenProvider` implementations. Custom `SymbolIcenProvider`s are registered for certain content types and are used in all views / dialogs like outline view, call hierarchy, and type hierachy.

An exemplary customization (introducing a custom overlay icon for destructors) is implemented in CDT LSP PR https://github.com/eclipse-cdt/cdt-lsp/pull/609, see corresponding issue https://github.com/eclipse-cdt/cdt-lsp/issues/608.

### DIFF
--- a/org.eclipse.lsp4e/schema/symbolIconsProvider.exsd
+++ b/org.eclipse.lsp4e/schema/symbolIconsProvider.exsd
@@ -98,23 +98,7 @@ This extension point offers client plug-ins options to adapt placement and selec
       </documentation>
    </annotation>
 
-   <annotation>
-      <appinfo>
-         <meta.section type="examples"/>
-      </appinfo>
-      <documentation>
-         [Enter extension point usage example here.]
-      </documentation>
-   </annotation>
 
-   <annotation>
-      <appinfo>
-         <meta.section type="apiinfo"/>
-      </appinfo>
-      <documentation>
-         [Enter API information here.]
-      </documentation>
-   </annotation>
 
    <annotation>
       <appinfo>
--- a/org.eclipse.lsp4e/schema/symbolIconsProvider.exsd
+++ b/org.eclipse.lsp4e/schema/symbolIconsProvider.exsd
@@ -23,7 +23,9 @@
          <attribute name="point" type="string" use="required">
             <annotation>
                <documentation>
-                  
+                  The fully qualified identifier of the extension point.
+
+This extension point offers client plug-ins options to adapt placement and selection of overlay icons for document symbols and other related elements.
                </documentation>
             </annotation>
          </attribute>
@@ -62,7 +64,7 @@
          <attribute name="class" type="string" use="required">
             <annotation>
                <documentation>
-                  
+                  The fully qualified name of the class implementing this symbol icon provider. The class must extend &lt;code&gt;org.eclipse.lsp4e.ui.SymbolIconProvider&lt;/code&gt;. Clients can override individual methods to customize overlay icons and their placement (top-left, top-right, bottom-left, bottom-right, underlay).
                </documentation>
                <appinfo>
                   <meta.attribute kind="java" basedOn="org.eclipse.lsp4e.ui.SymbolIconProvider:"/>
@@ -77,7 +79,7 @@
          <attribute name="contentTypeId" type="string" use="required">
             <annotation>
                <documentation>
-                  
+                  The unique identifier of a content type (as registered with the &lt;code&gt;org.eclipse.core.contenttype.contentTypes&lt;/code&gt; extension point) for which this icon provider should be used. The provider is also used for all derived content types that do not have a more specific provider registered.
                </documentation>
                <appinfo>
                   <meta.attribute kind="identifier" basedOn="org.eclipse.core.contenttype.contentTypes/content-type/@id"/>
@@ -92,7 +94,7 @@
          <meta.section type="since"/>
       </appinfo>
       <documentation>
-         [Enter the first release in which this extension point appears.]
+         0.19.10
       </documentation>
    </annotation>
 
@@ -119,7 +121,7 @@
          <meta.section type="implementation"/>
       </appinfo>
       <documentation>
-         [Enter information about supplied implementation of this extension point.]
+         LSP4E provides a default implementation via &lt;code&gt;org.eclipse.lsp4e.ui.SymbolIconProvider&lt;/code&gt;. It is used automatically for all content types that have no specific provider registered through this extension point. It renders symbol icons based on the LSP &lt;code&gt;SymbolKind&lt;/code&gt; and up to four overlay icons derived from &lt;code&gt;SymbolTag&lt;/code&gt;s (visibility, static/final/abstract/etc., marker severity) plus a deprecation underlay.
       </documentation>
    </annotation>
 
--- a/org.eclipse.lsp4e/schema/symbolIconsProvider.exsd
+++ b/org.eclipse.lsp4e/schema/symbolIconsProvider.exsd
@@ -1,0 +1,140 @@
+<?xml version='1.0' encoding='UTF-8'?>
+<!-- Schema file written by PDE -->
+<schema targetNamespace="org.eclipse.lsp4e" xmlns="http://www.w3.org/2001/XMLSchema">
+<annotation>
+      <appinfo>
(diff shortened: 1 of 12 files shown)


travkin79:
Hi @rubenporras and @FlorianKroiss,
Do you have some time to review this PR?

org.eclipse.lsp4e/schema/symbolIconsProvider.exsd
```diff
@@ -0,0 +1,140 @@
...
+         </appinfo>
+      </annotation>
+      <complexType>
+         <sequence minOccurs="1" maxOccurs="unbounded">
+            <element ref="iconProvider"/>
+         </sequence>
+         <attribute name="point" type="string" use="required">
+            <annotation>
+               <documentation>
+
```
rubenporras: can you provide some documentation instead of the empty one?

rubenporras: this applies to the other items as well

org.eclipse.lsp4e/schema/symbolIconsProvider.exsd
```diff
@@ -0,0 +1,140 @@
...
+         </attribute>
+      </complexType>
+   </element>
+
+   <annotation>
+      <appinfo>
+         <meta.section type="since"/>
+      </appinfo>
+      <documentation>
+         [Enter the first release in which this extension point appears.]
```
rubenporras: I am not sure if we need all of these annotations documenting the extension point. If you can provide them, they are nice, but then we need meaningful <documentation> tags.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/symbols/internal/SymbolIconProviderRegistry.java
```diff
@@ -0,0 +1,154 @@
...
+			INSTANCE = new SymbolIconProviderRegistry();
+		}
+		return INSTANCE;
+	}
+
+	private void loadExtensions() {
+		IExtensionPoint extensionPoint = Platform.getExtensionRegistry().getExtensionPoint(EXTENSION_POINT_ID);
+
+		if (extensionPoint == null) {
+			Platform.getLog(getClass()).error("No extension point found for ID " + EXTENSION_POINT_ID); //$NON-NLS-1$
```
rubenporras: should we not use org.eclipse.lsp4e.LanguageServerPlugin.logError in this and other examples in this class?

org.eclipse.lsp4e/src/org/eclipse/lsp4e/ui/SymbolIconProvider.java
```diff
@@ -206,25 +247,48 @@ protected String getImageKeyFromSymbolKindWithVisibility(SymbolKind kind, List<S
...
-	 * @see #getImageFor(SymbolKind, List)
+	 * @see #getImageFor(SymbolKind, List, Object)
 	 */
 	public @Nullable Image getImageFor(final @Nullable SymbolKind symbolKind,
-			final @Nullable List<SymbolTag> symbolTags, int severity) {
+			final @Nullable List<SymbolTag> symbolTags, int severity, Object symbol) {
 
 		if (symbolKind == null) {
+			// return empty image
 			return LSPImages.imageFromSymbolKind(symbolKind);
```
rubenporras: should we create a method LSPImages.emptyImage to make it clearer instead of a comment?

rubenporras (changes requested):
See comments

org.eclipse.lsp4e/schema/symbolIconsProvider.exsd
travkin79: What do you mean by meaningful tags?

org.eclipse.lsp4e/src/org/eclipse/lsp4e/ui/SymbolIconProvider.java
travkin79: Good idea. That makes the code self-explaining. 👍

org.eclipse.lsp4e/schema/symbolIconsProvider.exsd
travkin79: I've extended the documentation. Is that what you had in mind?

org.eclipse.lsp4e/src/org/eclipse/lsp4e/ui/SymbolIconProvider.java
```diff
@@ -33,6 +39,35 @@
...
+	 * {@link SymbolInformation}, or {@link WorkspaceSymbol},
+	 * or {@link CallHierarchyItem}, or {@link TypeHierarchyItem},
+	 * returns <code>null</code> otherwise.
+	 *
+	 * @param symbol the symbol to get the name for
+	 * @return the symbol's name or <code>null</code>.
+	 */
+	protected @Nullable String getName(Object symbol) {
+		String name = null;
+		if (symbol instanceof SymbolInformation info) {
```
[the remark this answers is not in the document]
FlorianKroiss: I agree with this warning and would prefer `return switch(symbol) {/*cases*/};`

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/symbols/internal/SymbolIconProviderRegistry.java
```diff
@@ -0,0 +1,158 @@
...
+				}
+				candidate = candidate.getBaseType();
+			}
+		}
+
+		return defaultIconProvider;
+	}
+
+	private @Nullable URI getUri(Object symbol) {
+		if (symbol instanceof SymbolInformation info)
```
FlorianKroiss: Similarly, I would prefer `return switch(...);` here

FlorianKroiss (changes requested):

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/symbols/internal/SymbolIconProviderRegistry.java
travkin79: Yes, you're absolutely right. I forgot that with newer Java versions we can handle such cases with a simpler, more readable switch expression.

travkin79:
Hi @rubenporras and @FlorianKroiss,
Thank you for your feedback. I think, I've fixed all mentioned issues. Could you take a look at the latest version and acknowledge? @rubenporras, could you please merge if everything is ok?

It would be great if we could create a new LSP4E release with that change soon, since my related CDT LSP PR https://github.com/eclipse-cdt/cdt-lsp/pull/609 depends on this PR.

FlorianKroiss (approved):

org.eclipse.lsp4e/schema/symbolIconsProvider.exsd
```diff
@@ -0,0 +1,142 @@
...
+   </annotation>
+
+   <annotation>
+      <appinfo>
+         <meta.section type="apiinfo"/>
+      </appinfo>
+      <documentation>
+         [Enter API information here.]
+      </documentation>
+   </annotation>
```
FlorianKroiss: Similar to what @rubenporras mentioned. If these elements don't have meaningful content, we should omit them (if that's valid?)

FlorianKroiss (changes requested):

org.eclipse.lsp4e/schema/symbolIconsProvider.exsd
travkin79: I created the extension point (exsd file) using the PDE UI and compared my file with the existing `languageServer.exsd`. Since it has the same dummy entries, I assumed we cannot omit these entries, but after removing the dummy entries (text) using the editor (via PDE UI) and saving the file, the entries were removed. So it seems, removing them is safe.

I guess, we should remove similar dummy entries in `languageServer.exsd`, too (but using a separate PR/commit).

FlorianKroiss (approved):

rubenporras:
there is a new release

travkin79:
Thank you very much, @rubenporras and @FlorianKroiss! You're feedback is very welcome and always helpful. Thank you also for quickly releasing the changes @rubenporras. 👍

PR #1523, opened 2026-04-16
Log server Info messages only if TRACE is enabled.

raghucssit:
see https://github.com/eclipse-lsp4e/lsp4e/issues/1522


FlorianKroiss:
Hi @raghucssit,

I don't think it's a good idea to hide these message is general.

That being said, I ran into a similar problem with a LS that was logging a lot of messages with INFO level.
To solve it, we made the *Message methods on DefaultLanguageClient non-final, see #1370.
This way, you can override the logMessage method in your own subclass and filter out some messages there.

rubenporras:
I also do not think this is a problem that should be solved by suppressing messages on LSP4E by default. I think https://github.com/eclipse-lsp4e/lsp4e/pull/1370 or fixing the offending language server is a better option.

raghucssit:
I have moved this filtering to github copilot repo. I will close this as needed later once that pr is merged.

Issue #1529, opened 2026-04-23
FileSystemWatcherManager crashes with FileSystemNotFoundException when workspace contains non-file URI projects


xinyi-gong:
### Environment:

- LSP4E version: (0.19.7)
- Eclipse: 2025-03
- OS: Windows

### Description:

When a workspace contains a project with a non-file URI scheme (e.g. SAP ABAP projects using semanticfs://), opening any file that is mapped to a language server via contentTypeMapping (e.g. a .md file matched by TM4E Language Pack's markdown content type) causes an unhandled FileSystemNotFoundException.

The error is thrown from FileSystemWatcherManager's constructor

Stack Trace:
```
org.eclipse.e4.core.di.InjectionException: java.nio.file.FileSystemNotFoundException: Provider "semanticfs" not installed
    at org.eclipse.e4.core.internal.di.MethodRequestor.execute(MethodRequestor.java:68)
    ...
Caused by: java.nio.file.FileSystemNotFoundException: Provider "semanticfs" not installed
    at java.base/java.nio.file.Path.of(Path.java:213)
    at java.base/java.nio.file.Paths.get(Paths.java:98)
    at org.eclipse.lsp4e.internal.files.FileSystemWatcherManager.<init>(FileSystemWatcherManager.java:57)
    at org.eclipse.lsp4e.LanguageServerWrapper.<init>(LanguageServerWrapper.java:328)
    at org.eclipse.lsp4e.LanguageServiceAccessor.getLSWrappers(LanguageServiceAccessor.java:339)
    at org.eclipse.lsp4e.LanguageServers$LanguageServerDocumentExecutor.getServers(LanguageServers.java:289)
    at org.eclipse.lsp4e.internal.LSPDocumentAbstractHandler$LanguageServerDocumentHandlerExecutor.anyMatching(LSPDocumentAbstractHandler.java:59)
    at org.eclipse.lsp4e.internal.LSPDocumentAbstractHandler.setEnabled(LSPDocumentAbstractHandler.java:123)
    at org.eclipse.lsp4e.operations.rename.LSPRenameHandler.setEnabled(LSPRenameHandler.java:79)
```

### Steps to Reproduce:

1. Open a workspace containing a project with a non-file URI scheme (e.g. SAP ABAP with semanticfs://)
2. Have TM4E Language Pack installed (provides `org.eclipse.tm4e.language_pack.markdown content type` with an LSP4E language server mapping)
[...]

PR #1531, opened 2026-04-27
fix: do not log exception if provider identified by the URI's scheme

rubenporras:
component is not installed


rubenporras:
Solves #1531

FlorianKroiss:
#1531 is this PR? Do you mean #1529 ?

PR #1532, opened 2026-04-27
Fix path parsing on Windows in SymbolIconProviderRegistry & add tests

travkin79:
On Windows a path parsing issue can occur while retrieving overlay icons for the outline view, type hierarchy, or call hierarchy. This PR provides a fix for [an issue reported in CDT LSP](https://github.com/eclipse-cdt/cdt-lsp/pull/609#issuecomment-4318375948) and adds a test for reproducing that issue.

### DIFF
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/symbols/internal/SymbolIconProviderRegistry.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/symbols/internal/SymbolIconProviderRegistry.java
@@ -12,7 +12,6 @@
 package org.eclipse.lsp4e.operations.symbols.internal;
 
 import java.net.URI;
-import java.nio.file.Path;
 import java.util.HashMap;
 import java.util.Map;
 
@@ -22,6 +21,7 @@ import org.eclipse.core.runtime.IExtensionPoint;
 import org.eclipse.core.runtime.Platform;
 import org.eclipse.core.runtime.content.IContentType;
 import org.eclipse.jdt.annotation.Nullable;
+import org.eclipse.lsp4e.LSPEclipseUtils;
 import org.eclipse.lsp4e.LanguageServerPlugin;
 import org.eclipse.lsp4e.outline.SymbolsModel.DocumentSymbolWithURI;
 import org.eclipse.lsp4e.ui.SymbolIconProvider;
@@ -106,11 +106,8 @@ public class SymbolIconProviderRegistry {
 			return defaultIconProvider;
 		}
 
-		String fileName = null;
-		try {
-			fileName = Path.of(uri.getPath()).getFileName().toString();
-		} catch (Exception e) {
-			LanguageServerPlugin.logWarning("Failed to parse file name from URI " + uri, e); //$NON-NLS-1$
+		String fileName = LSPEclipseUtils.getFileName(uri);
+		if (fileName == null) {
 			return defaultIconProvider;
 		}
 
@@ -146,8 +143,8 @@ public class SymbolIconProviderRegistry {
 		}
 
 		try {
-			return URI.create(uri);
-		} catch (IllegalArgumentException e) {
+			return LSPEclipseUtils.toUri(uri);
+		} catch (Exception e) {
 			LanguageServerPlugin.logWarning("Failed to parse URI " + uri, e); //$NON-NLS-1$
 			return null;
 		}
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/symbols/internal/SymbolIconProviderRegistry.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/symbols/internal/SymbolIconProviderRegistry.java
@@ -144,7 +144,7 @@ public class SymbolIconProviderRegistry {
 
 		try {
 			return LSPEclipseUtils.toUri(uri);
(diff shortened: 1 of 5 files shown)


org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/symbols/internal/SymbolIconProviderRegistry.java
```diff
@@ -108,7 +108,7 @@ private SymbolIconProvider getIconProvider(Object symbol) {
 
 		String fileName = null;
 		try {
-			fileName = Path.of(uri.getPath()).getFileName().toString();
+			fileName = getFileName(uri);
```
FlorianKroiss: There are already a lot of methods to convert between URIs/Paths/etc. in LSPEclipseUtils. I think it would be better to remove the conversion logic here and either use an existing utility method in LSPEclipseUtils or create a new one there. That should also make testing a lot easier.

FlorianKroiss (changes requested):

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/symbols/internal/SymbolIconProviderRegistry.java
travkin79: Good point. I'll try improving the code accordingly.

travkin79: I couldn't reuse any existing method in `LSPEclipseUtils`, but I moved the new method to that class and simplified production and test code.

FlorianKroiss (approved):

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/symbols/internal/SymbolIconProviderRegistry.java
FlorianKroiss, line 140: Nitpick: I guess we could also move this method to LSPEclipseUtils.

There already is a `toUri(String uri)` there, but without error handling / logging

travkin79: Yes, I saw that method in `LSPEclipseUtils`. I call it now (after checking and testing the `URI` conversion magic in it), but I still have a local `toUri` method for exception handling.

travkin79:
Thank you @FlorianKroiss for your review.

Hello @rubenporras, do you think, we could merge now? I squashed and rebased my PR commits. Oh, I see, the license check fails, but I don't see the reason. Any suggestions?

Besides, I'm wondering about some of our tests sporadically failing for strange reasons. Are there any ideas how we could avoid that?

FlorianKroiss (approved):

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/symbols/internal/SymbolIconProviderRegistry.java
```diff
@@ -146,8 +143,8 @@ private SymbolIconProvider getIconProvider(Object symbol) {
 		}
 
 		try {
-			return URI.create(uri);
-		} catch (IllegalArgumentException e) {
+			return LSPEclipseUtils.toUri(uri);
+		} catch (Exception e) {
```
rubenporras: Could we still catch only `IllegalArgumentException` as before? 

Could you as well as to the javadoc of `LSPEclipseUtils.toUri` that it may throw `IllegalArgumentException`?

rubenporras:
> Thank you @FlorianKroiss for your review.
> 
> Hello @rubenporras, do you think, we could merge now? I squashed and rebased my PR commits. 

I have just a minor comment, could you take a look and follow my suggestion or comment if you disagree? Then we can merge. Sorry for the late review, I have been on holidays for a week and forgot to set my status on GitHub, so it was not visible.

> Oh, I see, the license check fails, but I don't see the reason. Any suggestions?

@travkin79 , I have triggered the job again and it succeded. I must have been a temporal glitch.

> 
> Besides, I'm wondering about some of our tests sporadically failing for strange reasons. Are there any ideas how we could avoid that?

I have been wondering as well but I have currently not much time to work on LSP4E so I cannot help.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/symbols/internal/SymbolIconProviderRegistry.java
travkin79: I've added the javadoc to` LSPEclipseUtils.toUri(...)` and adapted the catched exceptions type.

travkin79:
Hi @rubenporras,

> I have just a minor comment, could you take a look and follow my suggestion or comment if you disagree? Then we can merge. Sorry for the late review, I have been on holidays for a week and forgot to set my status on GitHub, so it was not visible.

I agree with your suggestion and adapted the code accordingly.

> Sorry for the late review, I have been on holidays for a week and forgot to set my status on GitHub, so it was not visible.

No problem.

rubenporras (approved):

travkin79:
Thank you @rubenporras.

rubenporras:
fyi: I have also created a new release

PR #1543, opened 2026-06-04
refactor(tests): Use JUnit Jupiter mechanisms to handle general test setup and reporting

FlorianKroiss:
- The `junit.jupiter.extensions.autodetection.enabled=true` configuration automatically loads all listeners/extensions defined in META-INF/services. For example, we previously used the `TestInfoExtension` to log when a test starts. However, for this to work, you either had to manually add the extension to your test class or extend AbstractTest. Otherwise there would have been no output.
- `CloseIntroExtension` is now a regular Extension which just makes sure the Intro is closed. This has the same advantage, that you don't have to use it explicitly or inherit from AbstractTest
- `LoggingTestExecutionListener` is now used to report when a test case or test container starts/finish. This implements `TestExecutionListener` which has the advantage that it is called for each test, even for skipped ones and that you also get access to the result of the test case execution, which is something you cannot do with something like beforeEach/afterEach.
- `LoggingLauncherSessionListener` is now used to redirect stderr and configure the java.util.logging. This implements `LauncherSessionListener` which is only called once per test session, instead of previously being done before each test case.

### DIFF
--- a/org.eclipse.lsp4e.test/src/org/eclipse/lsp4e/test/utils/AbstractTest.java
+++ b/org.eclipse.lsp4e.test/src/org/eclipse/lsp4e/test/utils/AbstractTest.java
@@ -11,50 +11,17 @@
  *******************************************************************************/
 package org.eclipse.lsp4e.test.utils;
 
-import java.io.PrintStream;
-
 import org.eclipse.lsp4e.tests.mock.MockLanguageServer;
 import org.eclipse.lsp4j.ServerCapabilities;
-import org.junit.jupiter.api.AfterAll;
-import org.junit.jupiter.api.BeforeAll;
-import org.junit.jupiter.api.Order;
 import org.junit.jupiter.api.extension.RegisterExtension;
 
 /**
- * Test base class that configures a {@link AllCleanExtension} and a
- * {@link TestInfoExtension} and works around a surefire-plugin issue which
- * suppresses output to stderr
+ * Test base class that configures a {@link AllCleanExtension}.
  */
 public abstract class AbstractTest {
 
-	private static PrintStream originalSystemErr;
-
-	private static final boolean isExecutedBySurefirePlugin = System.getProperty("surefire.real.class.path") != null;
-
-	@BeforeAll
-	public static void setUpSystemErrRedirection() throws Exception {
-		if (isExecutedBySurefirePlugin) {
-			// redirect stderr to stdout during test execution as it is otherwise suppressed
-			// by the surefire-plugin
-			originalSystemErr = System.err;
-			System.setErr(System.out);
-		}
-	}
-
-	@AfterAll
-	public static void tearDownSystemErrRedirection() throws Exception {
-		if (isExecutedBySurefirePlugin) {
-			System.setErr(originalSystemErr);
-		}
-	}
-
 	@RegisterExtension
-	@Order(1)
 	public final AllCleanExtension allCleanRule = new AllCleanExtension(this::getServerCapabilities);
-	
-	@RegisterExtension
-	@Order(0)
-	public final TestInfoExtension testInfo = new TestInfoExtension();
 
 	/**
 	 * Override if required, used by {@link #allCleanRule}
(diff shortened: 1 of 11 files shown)


FlorianKroiss:
I have quite a few more changes lined up to make our tests less flaky but I did not want to make one mega PR 😄

FlorianKroiss:
This should be good to go. The existing failures also exist on main.
The logging also seems to work:
```
  12:51:47 INFO [LoggingTestExecutionListener] Starting testFilterNonmatchingCompletions() 
  12:51:47 INFO [LoggingTestExecutionListener] Finished testFilterNonmatchingCompletions(): SUCCESSFUL 
  12:51:47 INFO [LoggingTestExecutionListener] Starting testPrefixCaseSensitivityWithoutTextEditAtOffset() 
  12:51:47 INFO [LoggingTestExecutionListener] Finished testPrefixCaseSensitivityWithoutTextEditAtOffset(): SUCCESSFUL 
  12:51:47 INFO [LoggingTestExecutionListener] Starting testCompleteOnFileEnd() 
  12:51:47 INFO [LoggingTestExecutionListener] Finished testCompleteOnFileEnd(): SUCCESSFUL 
```

rubenporras:
@FlorianKroiss , I am not very familiar with JUnit best practices and extensions. Why do you think this is better that what LSP4E was doing so far? Could you explain it for a dummy user of JUnit?

FlorianKroiss:
@rubenporras It's basically about using the right callback for each purpose.

- The `junit.jupiter.extensions.autodetection.enabled=true` configuration automatically loads all listeners/extensions defined in META-INF/services. For example, we previously used the `TestInfoExtension` to log when a test starts. However, for this to work, you either had to manually add the extension to your test class or extend AbstractTest. Otherwise there would have been no output.
- `CloseIntroExtension` is now a regular Extension which just makes sure the Intro is closed. This has the same advantage, that you don't have to use it explicitly or inherit from AbstractTest
- `LoggingTestExecutionListener` is now used to report when a test case or test container starts/finish. This implements `TestExecutionListener` which has the advantage that it is called for each test, even for skipped ones and that you also get access to the result of the test case execution, which is something you cannot do with something like beforeEach/afterEach.
- `LoggingLauncherSessionListener` is now used to redirect stderr and configure the java.util.logging. This implements `LauncherSessionListener` which is only called once per test session, instead of previously being done before each test case.

org.eclipse.lsp4e.test/src/org/eclipse/lsp4e/test/utils/AbstractTest.java
```diff
@@ -51,10 +51,6 @@ public static void tearDownSystemErrRedirection() throws Exception {
 	@RegisterExtension
 	@Order(1)
```
rubenporras: Does `@Order(1)` still make sense if there is no `@Order(0)` in the file?

rubenporras:
> @rubenporras It's basically about using the right callback for each purpose.
> 
> * The `junit.jupiter.extensions.autodetection.enabled=true` configuration automatically loads all listeners/extensions defined in META-INF/services. For example, we previously used the `TestInfoExtension` to log when a test starts. However, for this to work, you either had to manually add the extension to your test class or extend AbstractTest. Otherwise there would have been no output.
> * `CloseIntroExtension` is now a regular Extension which just makes sure the Intro is closed. This has the same advantage, that you don't have to use it explicitly or inherit from AbstractTest
> * `LoggingTestExecutionListener` is now used to report when a test case or test container starts/finish. This implements `TestExecutionListener` which has the advantage that it is called for each test, even for skipped ones and that you also get access to the result of the test case execution, which is something you cannot do with something like beforeEach/afterEach.
> * `LoggingLauncherSessionListener` is now used to redirect stderr and configure the java.util.logging. This implements `LauncherSessionListener` which is only called once per test session, instead of previously being done before each test case.

Thanks, could you add this explanation to the PR and commit message?

org.eclipse.lsp4e.test/src/org/eclipse/lsp4e/test/utils/AbstractTest.java
FlorianKroiss: Since there is only one extension left, I removed the Order annotation entirely.

rubenporras (approved):

FlorianKroiss:
Sorry, I forgot to commit some removed code.

rubenporras (approved):

PR #1544, opened 2026-06-07
fix: Send a single hover request every time the user hovers

FlorianKroiss:
I looked at the Eclipse Platform code and the protocol for `getHoverRegion` and `getHoverInfo2` seems to be:
```java
IRegion region = hover.getHoverRegion(viewer, offset);
// [... setup some stuff with region ...]
hover.getHoverInfo2(viewer, region);
```
Most importantly, the region returned from `getHoverRegion` is also the one passed back to `getHoverInfo2`. We can use this to return a custom Region object which holds the request sent to the LS. This way, `getHoverInfo2` can simply use that stored request instead of having to do its own bookkeeping to know if the last stored request is the correct one to use.
This also makes sure that exactly one request is sent per getHoverRegion/getHoverInfo2 cycle.

I also added another fast-path to avoid async loading, if the result is already present.

Fixes #1514

### DIFF
--- a/org.eclipse.lsp4e.test/META-INF/MANIFEST.MF
+++ b/org.eclipse.lsp4e.test/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: Tests for language server bundle (Incubation)
 Bundle-SymbolicName: org.eclipse.lsp4e.test;singleton:=true
-Bundle-Version: 0.16.11.qualifier
+Bundle-Version: 0.16.12.qualifier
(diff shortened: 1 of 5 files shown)


FlorianKroiss:
This is the alternative fix for #1514 that I mentioned in https://github.com/eclipse-lsp4e/lsp4e/pull/1533#issuecomment-4639614767

@betamaxbandit I incorporated the test cases from your PR which also pass with these changes.

I prefer the changes of this PR because it greatly reduces the complexity / state of LSPTextHover

rubenporras (approved):
Hi Florian,

thanks for taking a look. Let us try this approach.

Regards

PR #1545, opened 2026-06-07
fix(tests): Stop timer thread when shutting down dispatchers

FlorianKroiss:
We call `LanguageServerWrapper.stopDispatcher()` after each test to clean up the spawned executors/threads. However, we did not include the Timer instance, which would cause many leaked Threads during test execution


FlorianKroiss:
Before this change, running the tests locally for a while looked like this in the debugger:
Screenshot 2026-06-07 152310

rubenporras (approved):

PR #1552, opened 2026-06-23
fix: Check content type before LSP enablement

### LINKED ISSUE #1551: Avoid eager language-server enablement checks
When checking available language servers, LSP4E currently evaluates `mapping.isEnabled(uri)` before checking whether the file content type matches the mapping.

For mappings with an `enabledWhen`, this can be expensive. In our case, CDT LSP’s `enabledWhen` loads the file as a document. This also happens for files that do not match the CDT content types, for example large `.vcd` files. With very large files this can freeze the UI and may even lead to OOM errors.

The content type check should happen first, so `enabledWhen` is only evaluated for mappings that can actually apply to the file.

So 

```java
	private Set<LanguageServerDefinition> getAvailableLSFor(Collection<IContentType> contentTypes, @Nullable URI uri) {
		final var res = new HashSet<LanguageServerDefinition>();
		contentTypes = expandToSuperTypes(contentTypes);
		for (ContentTypeToLanguageServerDefinition mapping : this.connections) {
			if (mapping.isEnabled(uri) && contentTypes.contains(mapping.getKey())) {
				res.add(mapping.getValue());
			}
		}
		return res;
	}
```

should become to:

```java
	private Set<LanguageServerDefinition> getAvailableLSFor(Collection<IContentType> contentTypes, @Nullable URI uri) {
		final var res = new HashSet<LanguageServerDefinition>();
		contentTypes = expandToSuperTypes(contentTypes);
		for (ContentTypeToLanguageServerDefinition mapping : this.connections) {
			if (contentTypes.contains(mapping.getKey()) && mapping.isEnabled(uri)) {
				res.add(mapping.getValue());
			}
		}
		return res;
	}
```

bytehaufen:
This changes the order of checks when looking up available language servers.

Before, LSP4E evaluated `mapping.isEnabled(uri)` before checking whether the mapping's content type matched the file. For mappings with an `enabledWhen`, that can be expensive and may load the file as a document even when the mapping does not apply.

With this change, the content type is checked first, so `enabledWhen` is only evaluated for matching mappings.

Fixes #1551

### DIFF
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServersRegistry.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServersRegistry.java
@@ -473,7 +473,7 @@ public class LanguageServersRegistry {
 		final var res = new HashSet<LanguageServerDefinition>();
 		contentTypes = expandToSuperTypes(contentTypes);
 		for (ContentTypeToLanguageServerDefinition mapping : this.connections) {
-			if (mapping.isEnabled(uri) && contentTypes.contains(mapping.getKey())) {
+			if (contentTypes.contains(mapping.getKey()) && mapping.isEnabled(uri)) {
 				res.add(mapping.getValue());
 			}
 		}


FlorianKroiss (approved):
Thanks for the PR! The fix looks straightforward.

rubenporras:
Thank you for the PR!

bytehaufen:
Thanks for the quick merge, and thanks for your work!

PR #1554, opened 2026-06-28
fix: Properly wait for pending document setups

FlorianKroiss:
The CompletableFutures that we want to wait for during testing are only added to `PENDING_CONNECTIONS` after a delay of one second. So it's possible, that `waitForAll` has already returned but there are still pending document setups

### DIFF
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/ConnectDocumentToLanguageServerSetupParticipant.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/ConnectDocumentToLanguageServerSetupParticipant.java
@@ -16,6 +16,8 @@ import java.util.Set;
 import java.util.WeakHashMap;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ExecutionException;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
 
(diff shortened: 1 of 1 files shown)


FlorianKroiss:
A small fix which I extracted from #1546 which can be merged independently.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/ConnectDocumentToLanguageServerSetupParticipant.java
```diff
@@ -49,24 +60,39 @@ public void setup(IDocument document) {
...
 			try {
-				cf.get(1000, TimeUnit.MILLISECONDS);
+				future.get(1, TimeUnit.SECONDS);
 			} catch (InterruptedException | ExecutionException | TimeoutException e) {
-				LanguageServerPlugin.logInfo("Interrupted trying to cancel document setup"); //$NON-NLS-1$;
+				LanguageServerPlugin.logError("Interrupted trying to cancel document setup", e); //$NON-NLS-1$ ;
 			}
 		});
+
+		// Create new executor for next test.
```
rubenporras:
```suggestion
		// Create new executor for next setup.
```

rubenporras (changes requested):
Would you mind changing the comment as I suggest or otherwise? It looks to me from the class comment that this class in not "test" only.

org.eclipse.lsp4e/src/org/eclipse/lsp4e/ConnectDocumentToLanguageServerSetupParticipant.java
rubenporras: Also, I think the PR title should be fix rather than fix(test) following the same reasoning

rubenporras (approved):

PR #1561, opened 2026-07-21
fix: LSP4E DeleteParticipant fails with FileSystemNotFoundException for custom EFS URI schemes

StefanFu:
Fix for https://github.com/eclipse-lsp4e/lsp4e/issues/1559

### DIFF
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/rename/LSPFileOperationParticipantSupport.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/rename/LSPFileOperationParticipantSupport.java
@@ -111,13 +111,14 @@ public final class LSPFileOperationParticipantSupport {
 			final Function<FileOperationsServerCapabilities, @Nullable FileOperationOptions> optionsProvider) {
 		final var uri = LSPEclipseUtils.toUri(res);
 		final var project = res.getProject();
-		if (uri == null) {
-			// No URI means we cannot match any file operation filters; return an executor
+		if (uri == null || !"file".equalsIgnoreCase(uri.getScheme())) { //$NON-NLS-1$
+			// No URI or file scheme means we cannot match any file operation filters; return an executor
 			// that will not match any servers.
 			return LanguageServers.forProject(project).withFilter(capabilities -> false);
 		}
 
-		final var path = Path.of(uri);
+		final Path path = Path.of(uri);
+
 		return LanguageServers.forProject(project).withFilter(capabilities -> {
 			final var workspace = capabilities.getWorkspace();
 			if (workspace == null)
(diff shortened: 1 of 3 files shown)


org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/rename/LSPFileOperationParticipantSupport.java
```diff
@@ -117,18 +118,23 @@ public static LanguageServerProjectExecutor createFileOperationExecutor(final IR
...
-			if (workspace == null)
-				return false;
-			final var fileOps = workspace.getFileOperations();
-			if (fileOps == null)
-				return false;
-
-			final var options = optionsProvider.apply(fileOps);
-			return matches(options, path, res.getType() == IResource.FOLDER);
-		});
+		try {
```
FlorianKroiss: For clarity, I would prefer if we move the try-catch up, and only wrap the `Path.of` in the try-catch.
So something like:
```java
final Path path;
try {
	path = Path.of(uri);
} catch (FileSystemNotFoundException | IllegalArgumentException e) {
    // Not an NIO path, so we cannot match any file operation filters
	return LanguageServers.forProject(project).withFilter(capabilities -> false);
}

// Remaining return ...
```

FlorianKroiss (changes requested):
Thanks for the PR. I just have a small remark.

FlorianKroiss:
@StefanFu since this is the first change to `org.eclipse.lsp4e` since the last release, you have to bump the versions in `org.eclipse.lsp4e/META-INF/MANIFEST.MF` and `org.eclipse.lsp4e/pom.xml` from `0.19.13` to `0.19.14` to make the build work.

FlorianKroiss (approved):

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/rename/LSPFileOperationParticipantSupport.java
```diff
@@ -117,7 +118,15 @@ public static LanguageServerProjectExecutor createFileOperationExecutor(final IR
...
 		}
 
-		final var path = Path.of(uri);
+		Path path;
+
+		try {
+			path = Path.of(uri);
+		} catch (FileSystemNotFoundException | IllegalArgumentException e) {
+		    // Not an NIO path, so we cannot match any file operation filters
+			return LanguageServers.forProject(project).withFilter(capabilities -> false);
```
rubenporras: As I understand the code, only file schema URIs are handled, could not exit by checking "file".equalsIgnoreCase(uri.getScheme()) and avoid relying on unchecked exceptions? I think this is a bit cleaner.

If you agree, could you change the code to that?

rubenporras (changes requested):

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/rename/LSPFileOperationParticipantSupport.java
StefanFu: Yes, it would solve the issue as well. I updated the code.

rubenporras (approved):
thanks

StefanFu:
@rubenporras Should I raise the version so the build can work?

rubenporras:
Yes, please

PR #1569, opened 2026-08-02
fix(codeql): Address some easy to fix CodeQL warnings

FlorianKroiss:
Also contains a few minor test cleanups

### DIFF
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/outline/LSSymbolsContentProvider.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/outline/LSSymbolsContentProvider.java
@@ -48,7 +48,6 @@ import org.eclipse.jface.text.reconciler.DirtyRegion;
 import org.eclipse.jface.text.reconciler.IReconcilingStrategy;
 import org.eclipse.jface.viewers.ITreeContentProvider;
 import org.eclipse.jface.viewers.ITreeSelection;
-import org.eclipse.jface.viewers.StructuredViewer;
 import org.eclipse.jface.viewers.TreePath;
 import org.eclipse.jface.viewers.TreeSelection;
 import org.eclipse.jface.viewers.TreeViewer;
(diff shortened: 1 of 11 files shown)


org.eclipse.lsp4e/src/org/eclipse/lsp4e/outline/LSSymbolsContentProvider.java
```diff
@@ -242,7 +241,7 @@ public void resourceChanged(IResourceChangeEvent event) {
 					event.getDelta().accept(delta -> {
 						if (delta.getResource().equals(this.resource)) {
 							viewer.getControl().getDisplay().asyncExec(() -> {
-								if (!viewer.getControl().isDisposed() && viewer instanceof StructuredViewer) {
```
rubenporras, line 245: Hi @angelozerr,

I see that you introduced the check on the `instanceof` in https://github.com/eclipse-lsp4e/lsp4e/commit/4209f8e64b047365065c6186c5fd711f7836bd5b to "Reduce refresh of outline when editor content changed". The commit is very old, but do you remember which kind of viewers you wanted to filter out? Do you think the condition is still relevant?

rubenporras, line 245: @FlorianKroiss , thanks for the fixes. All the others are clear to me, but I am hesitant on this one. Would you like to wait for an answer on my previous comment or would you prefer to split it out to another PR so that we can merge the rest? Or do you know that the reason to put the condition does no longer apply?

FlorianKroiss: `viewer` is a TreeViewer, as it comes from here
https://github.com/eclipse-lsp4e/lsp4e/blob/bc77bf3b9171c0c499b9a4b1db53416c8595bfdc/org.eclipse.lsp4e/src/org/eclipse/lsp4e/outline/LSSymbolsContentProvider.java#L259

Every TreeViewer is also an instance of StructuredViewer:
image
So this test here always evaluated to true.

The relevant check https://github.com/eclipse-lsp4e/lsp4e/security/code-scanning/155 says the same

rubenporras (approved):

org.eclipse.lsp4e/src/org/eclipse/lsp4e/outline/LSSymbolsContentProvider.java
rubenporras: I agree, thanks for the explanation.

rubenporras: @FlorianKroiss has explained already why the check can be removed.

PR #1575, opened 2026-08-13
fix(outline): avoid deep DocumentSymbol hashing on every outline refresh

vogella:
LSP4J generates `DocumentSymbol.hashCode()` and `equals()` over all fields including the children list, so both recurse over the whole subtree. The outline tree viewer calls them per element on every refresh, which makes a refresh cost the sum of all subtree sizes rather than O(1) per element. Profiling a 100,000 line JSON file in the Generic Editor showed 2 to 3.6 seconds of UI thread work per keystroke with the Outline view open, versus 35 to 49 ms with it closed, with 86% of busy UI samples in `DocumentSymbol.hashCode()` reached from `setExpandedTreePaths()`.

Setting an `IElementComparer` that identifies elements by name, kind and selection start fixes this without any JFace change, because JFace routes both the element map and `TreePath.hashCode()` through the comparer, neutralising the deep `hashCode` and the deep `equals` at once. After the change an open Outline view costs about the same as a closed one again, roughly 100 ms against a 99 ms closed-outline control. It is set on `LSSymbolsContentProvider`, so it covers both the Outline view and the Quick Outline.

The key includes the position because `(name, kind)` alone collides across the tree (`items/0/id` and `items/1/id` are both `name="id" kind=Number`), and `CommonViewer` forces `setUseHashlookup(true)`, so duplicate equal elements anywhere break element to item lookup. It uses the selection range rather than the full range because the full range grows while typing inside a symbol body.

Note this does not on its own make outlines of large files fast. The Common Navigator Framework keeps element keyed collections that ignore the viewer comparer and still hash symbols deeply on every refresh (`NavigatorContentService.contributionMemory` and `contributionMemoryFirstClass`, plus `ContributorTrackingSet`); those account for most of what remains and need a fix on the Platform side.

### DIFF
--- a/org.eclipse.lsp4e.test/src/org/eclipse/lsp4e/test/outline/SymbolsElementComparerTest.java
+++ b/org.eclipse.lsp4e.test/src/org/eclipse/lsp4e/test/outline/SymbolsElementComparerTest.java
@@ -1,0 +1,183 @@
+/*******************************************************************************
+ * Copyright (c) 2026 Vogella GmbH and others.
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *  Lars Vogel (Vogella GmbH) - initial implementation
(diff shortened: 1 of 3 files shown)


rubenporras (approved):
Thanks a lot for the performance improvement!

PR #1578, opened 2026-08-18
feat: Support active parameter in signature help

FlorianKroiss:
+ Introduce `LSContextInformation` to hold information about currently active parameter
+ Introduce `LSContextInformationValidator` to update active parameter when cursor moves and update styling of displayed context info
+ `LSCompletionProposal` no longer returns itself as `IContextInformation`. I think this behavior is wrong in the context of LSP, because completions don't have knowledge about their signature help.
+ Add a few test cases

### DIFF
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/completion/LSCompletionProposal.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/completion/LSCompletionProposal.java
@@ -806,7 +806,7 @@ public class LSCompletionProposal
 
 	@Override
 	public @Nullable IContextInformation getContextInformation() {
-		return this;
+		return null;
 	}
 
 	@Override
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/completion/LSContentAssistProcessor.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/completion/LSContentAssistProcessor.java
@@ -37,8 +37,6 @@ import org.eclipse.jface.text.BadLocationException;
 import org.eclipse.jface.text.IDocument;
 import org.eclipse.jface.text.ITextViewer;
 import org.eclipse.jface.text.contentassist.CompletionProposal;
-import org.eclipse.jface.text.contentassist.ContextInformation;
-import org.eclipse.jface.text.contentassist.ContextInformationValidator;
 import org.eclipse.jface.text.contentassist.ICompletionProposal;
 import org.eclipse.jface.text.contentassist.IContentAssistProcessor;
 import org.eclipse.jface.text.contentassist.IContextInformation;
@@ -340,7 +338,9 @@ public class LSContentAssistProcessor implements IContentAssistProcessor {
 		if (docString != null && !docString.isEmpty()) {
 			signature.append('\n').append(docString);
 		}
-		return new ContextInformation(information.getLabel(), signature.toString());
+
+		LocationInString activeParameter = CompletionProposalTools.extractActiveParameter(information);
+		return new LSContextInformation(information.getLabel(), signature.toString(), activeParameter);
 	}
 
 	private void getFuture(@Nullable CompletableFuture<?> future) {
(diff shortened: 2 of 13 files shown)


org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/completion/LSContextInformationValidator.java
```diff
@@ -0,0 +1,92 @@
...
+		if (!(fViewer instanceof ITextViewer viewer)) {
+			return false;
+		}
+		if (!(fContextInformation instanceof LSContextInformation contextInfo)) {
+			return false;
+		}
+
+		// Check if the current context information is still valid for the given offset.
+		IContextInformation[] infos = fProcessor.computeContextInformation(viewer, offset);
+		if (infos != null && infos.length > 0) {
```
rubenporras: could be simplified as 
```suggestion
		if (infos != null) {
```
, right?

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/completion/LSContentAssistProcessor.java
```diff
@@ -335,12 +333,14 @@ private static List<ICompletionProposal> toProposals(IDocument document, int off
 	}
 
 	private static IContextInformation toContextInformation(SignatureInformation information) {
-		final var signature = new StringBuilder(information.getLabel());
```
rubenporras, line 338: since this is not the goal of the PR and I do not think the change is important, would you mind keeping `signature` to make the change a bit smaller?

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/completion/LSCompletionProposal.java
```diff
@@ -806,7 +806,7 @@ protected String getInsertText() {
 
 	@Override
 	public @Nullable IContextInformation getContextInformation() {
-		return this;
+		return null;
```
rubenporras, line 809: I think this is what you mention in the commit message in `LSCompletionProposal no longer returns itself as IContextInformation. I think this behavior is wrong in the context of LSP, because completions don't have knowledge about their signature help`, right?

I am not following the argument, the class implements all the methods of `IContextInformation` so it will return an object which provides meaningful information, will it not?

In addition, this would not be required to add the new capability, so we could if in doubt leave it unchanged, couldn't we?

rubenporras (changes requested):

org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/completion/LSCompletionProposal.java
rubenporras: @FlorianKroiss , what about this comment?

FlorianKroiss: Sorry, I was too fast to re-request 😄 

> I am not following the argument, the class implements all the methods of IContextInformation so it will return an object which provides meaningful information, will it not?

The problem is the following:
1. User accepts a proposal
2. Eclipse calls `getContextInformation` on it to check if the proposal already provides the context which Eclipse can display
3. Since the proposal returns itself, Eclipse shows the information provided by this proposal
4. Immediately after, Eclipse validates if the context is valid by checking if the same context is returned when querying the ContentAssistProcessor and if not, hides the context info again

The check at (4) will almost always fail and cause a flickering popup, because
- the LS probably returns entirely different content for proposals and signature help. For example, the completion string might only be the name of a function, whereas the signature help is only the list of parameters.
- the text from the CompletionProposal is HTML, but the text from `ContentAssistProcessor.computeContextInformation` is plain text

In our application, this leads to the following popup briefly flickering, containing the info from the proposal:
image

If the user continues to type, the actual content from the signature help is shown:
image

Note that the content is not only in a different format (HTML vs plaintext) but also contains different information regarding the signature.

So I think it doesn't make sense for `getContextInformation` to return anything, because whatever we return is most likely not what the server will provide.

rubenporras (approved):

Issue #1419, opened 2025-11-24
LSPClientFeatures support


angelozerr:
As LSP4E seems very active, I would like to share a feature I’ve implemented in **[LSP4IJ](https://github.com/redhat-developer/lsp4ij)** that provides an extremely simple way to customize LSP server integration and to cleanly check server capabilities.

You can find the **LSP client features API** here:  
https://github.com/redhat-developer/lsp4ij/blob/main/docs/LSPApi.md

This API is very powerful because it allows you to customize almost anything in a clean and structured way. For example, you can disable specific LSP features, override labels returned by the language server, etc.

If a plugin wants to integrate a language server that provides formatting and needs to expose custom UI preferences—for instance, to configure `tabSize`—it simply has to register a custom [LSPClientFeatures](https://github.com/redhat-developer/lsp4ij/blob/main/src/main/java/com/redhat/devtools/lsp4ij/client/features/LSPClientFeatures.java) with a custom [LSPFormattingFeature](https://github.com/redhat-developer/lsp4ij/blob/main/src/main/java/com/redhat/devtools/lsp4ij/client/features/LSPFormattingFeature.java) and override `getTabSize` so that it retrieves the tab size from the plugin’s UI settings.

If the language server needs to support **WSL**, the plugin can override URI building to translate WSL paths and generate the correct URI expected by the server.

Each feature implements an `isSupported(PsiFile file)` method, which relies on server capabilities (both static and dynamic). The codebase is very clear: for example, if you want to see how codelens capability detection is implemented, you can look at `LSPCodeLensFeature#isSupported`. This method can be overridden if a plugin wants to provide UI settings to enable or disable codelens for a specific language server.

The `isSupported` method relies on capability registries such as:  
https://github.com/redhat-developer/lsp4ij/blob/main/src/main/java/com/redhat/devtools/lsp4ij/server/capabilities/CodeLensCapabilityRegistry.java

These registries extend abstract classes that implement full capability checking according to the LSP specification (documentSelector, filters, etc.), for example:  
https://github.com/redhat-developer/lsp4ij/blob/6f41f6d22a7146f31e0218cb459513abd5dc16d3/src/main/java/com/redhat/devtools/lsp4ij/server/capabilities/TextDocumentServerCapabilityRegistry.java#L134

If you like this design (many IntelliJ plugins appreciate this feature), I believe you could reuse a significant part of the code and adapt it for **LSP4E**.

[...]

PR #1586, opened 2026-09-01
Dynamic registration: recompute effective capabilities instead of pat…

### LINKED ISSUE #1565: Dynamic Registration of Code Actions Ignores Server Payload
Just ran into this as part of a rewrite of our Eclipse lsp4e-based plugin so that it more closely mirrors our VSCode offering. We switched over to using dynamic registration in our language server and I've just hit an awkward side effect:

we support code actions + (newly) dynamic registration

We don't support `codeAction/resolve` as we don't need it. This wasn't a problem with static registration, as we didn't populate `CodeActionOptions.resolveProvider`. However with dynamic registration, the detailed payload is ignored and `LanguageServerWrapper` just sets `true` (the left side of the `boolean | CodeActionOptions` union in the server capabilities).
https://github.com/eclipse-lsp4e/lsp4e/blob/06a95f28d3c117387c625d4404cacd40b37e2912/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LanguageServerWrapper.java#L1221

Then when `CodeActionMarker.run` executes, if the action doesn't have an edit, it calls `CodeActionCompletionProposal.isCodeActionResolveSupported()`  -

https://github.com/eclipse-lsp4e/lsp4e/blob/06a95f28d3c117387c625d4404cacd40b37e2912/org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/codeactions/CodeActionCompletionProposal.java#L49

Because this only has the blanket `true` that the dynamic registration logic has stored, this also evaluates to true, so it then tries to call the unsupported `codeAction/resolve` method on our language server, which falls over...

Any thoughts on this? Looking at the rest of the dynamic registration code, is there any reason not to just unserialise the payload and store it in the server capabilities? The existing code seems to be a mixture - for some capabilities we unserialise the full payload and use it in its entirety, and for others we just register a yes/no capability. I'm assuming that this is mostly because there wasn't any demand for fuller support and so it just never needed to be  implemented, rather than any technical limitation (e.g. problems unserialising)? If so, I could presumably submit a small patch that enhances what's there currently with an appropriate test?

Thanks

ahmedneilhussain:
Fixes eclipse-lsp4e/lsp4e#1565.

The current implementation of dynamic registration (in `LanguageServerWrapper.registerCapability()`) has a number of issues:
 * Not all the methods attempted to use the detailed `RegistrationOptions` payload
    * In many cases we just ignored the payload and set the capability to `true`, resulting in dynamic registration support that was less rich than if the same features were registered statically
    * In other cases we weren't deserializing and applying the payload correctly
 * Our implementation didn't take into account the fact that dynamic registration payloads are _not_ necessarily the same as the corresponding subtree of `ServerCapabilities` - in many cases they are a mixin of the corresponding subtree and `TextDocumentRegistrationOptions` which carries a document selector. This has the following consequences:
   * Dynamic registrations can be sent multiple times for the same method but with different document selectors, so our implementation of registration/deregistration could result in a lot of cross talk, especially if overlapping registrations were not unregistered in strict reverse order
   * The way that we query server capabilities in the consuming code elsewhere in LSP4e assumes that there is only ever a single, monolithic server capabilities for a given language server, with no way of taking into account that these might vary by document.
   * The `mixin` types are not always of the correct type to be inserted into the appropriate subtree of `ServerCapabilities`: this is an awkward side-effect of how LSP4j has to translate the LSP schema into Java DTOs - some corner cases that are trivially expressed in Typescript's structural typing system are harder in Java's nominally-typed single-rooted hierarchy.

The solution:
 * Store all dynamic registrations sent by the server, with their correctly unserialised `RegistrationOption` payloads
 * For each method/payload type, have specialised code that can copy properties across into `ServerCapabilities` if the dynamic types are not directly insertable.
 * Add the ability to compute an 'effective server capabiltiies', taking into account the document type, and hence work out which document selectors are in play. We compute these capabilities lazily, and cache them according to the set of registrations that apply: the bet is that for a given language server there are only going to be a handful at most of different file types that it might support in different ways (eg source files and cargo files for Rust).
 * Use the document URI to retrieve the effective server capabilities in LanguageServerDocumentExecutor.filter()`

Also made the following miscellaneous related changes:
- codeAction/resolve is only used when resolveProvider is explicitly true (a second bug uncovered in #1565 - we were assuming that the coarse-grained 'code action supported' capability, with no fine-grained information, meant resolveProvider could be assumed, whereas it should only be invoke if explicitly declared by the server)
- advertise dynamicRegistration for rangeFormatting, selectionRange and typeHierarchy, which registerCapability already handled
- MockTextDocumentService gains a hook to observe codeAction/resolve

### DIFF
--- a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/internal/DocumentSelectorMatcher.java
+++ b/org.eclipse.lsp4e/src/org/eclipse/lsp4e/internal/DocumentSelectorMatcher.java
@@ -1,0 +1,127 @@
+/*******************************************************************************
+ * Copyright (c) 2026 Cocotec Ltd and others.
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *  Ahmed Hussain (Cocotec Ltd) - initial implementation
+ *
+ *******************************************************************************/
+package org.eclipse.lsp4e.internal;
+
+import java.net.URI;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.ArrayList;
+import java.util.List;
+
+import org.eclipse.jdt.annotation.Nullable;
+import org.eclipse.lsp4e.internal.files.PathPatternMatcher;
+import org.eclipse.lsp4j.DocumentFilter;
+
+/**
+ * A compiled LSP {@code DocumentSelector}: decides whether a document URI is matched by the
+ * {@code documentSelector} of a dynamic capability registration.
+ * <p>
+ * Per the LSP specification a selector is a list of {@link DocumentFilter}s combined with OR, and
+ * within one filter the properties that are set ({@code language}, {@code scheme}, {@code pattern})
+ * are combined with AND. A filter's glob pattern matches against the document's absolute path,
+ * except for a {@code RelativePattern}, which matches against the path relative to its own base URI.
+ * <p>
+ * Compile a selector once (when the registration arrives) via {@link #compile(List)} and reuse the
+ * instance for all subsequent queries; the glob matchers are precompiled.
+ */
+public final class DocumentSelectorMatcher {
+
+	/**
+	 * Resolves the LSP language id of the document at a given URI, i.e. the id that is (or would be)
+	 * sent to this filter's server in {@code textDocument/didOpen} for that document.
+	 */
+	@FunctionalInterface
+	public interface LanguageIdResolver {
+		@Nullable
+		String getLanguageId(URI uri);
+	}
+
+	private record CompiledFilter(@Nullable String language, @Nullable String scheme,
+			@Nullable PathPatternMatcher pattern) {
+
+		boolean matches(final URI uri, final LanguageIdResolver languageIdResolver) {
+			if (language != null && !language.equals(languageIdResolver.getLanguageId(uri))) {
+				return false;
+			}
+			if (scheme != null && !scheme.equalsIgnoreCase(uri.getScheme())) {
+				return false;
+			}
+			if (pattern != null && !matchesPattern(uri)) {
+				return false;
+			}
+			return true;
+		}
+
+		private boolean matchesPattern(final URI uri) {
+			final PathPatternMatcher pattern = this.pattern;
+			if (pattern == null) {
+				return false;
+			}
+			final Path path;
+			try {
+				path = Paths.get(uri);
+			} catch (final Exception ex) {
+				// no file-system path can be derived (e.g. non-file scheme): the pattern cannot match
+				return false;
+			}
+			final Path basePath = pattern.getBasePath();
+			if (basePath != null) { // RelativePattern: match relative to its base URI
+				if (!path.startsWith(basePath)) {
+					return false;
+				}
+				return pattern.matches(basePath.relativize(path));
+			}
+			return pattern.matches(path);
+		}
+	}
+
+	private final List<CompiledFilter> filters;
+
+	private DocumentSelectorMatcher(final List<CompiledFilter> filters) {
+		this.filters = filters;
+	}
+
+	/**
+	 * Compiles the {@code documentSelector} of a registration.
+	 *
+	 * @return the compiled selector, or {@code null} for a {@code null} selector, which per the LSP
+	 *         specification means the registration applies to every document the server is used for
+	 */
+	public static @Nullable DocumentSelectorMatcher compile(final @Nullable List<DocumentFilter> documentSelector) {
+		if (documentSelector == null) {
+			return null;
+		}
+		final var filters = new ArrayList<CompiledFilter>(documentSelector.size());
+		for (final DocumentFilter filter : documentSelector) {
+			final String language = isNullOrBlank(filter.getLanguage()) ? null : filter.getLanguage();
+			final String scheme = isNullOrBlank(filter.getScheme()) ? null : filter.getScheme();
+			final PathPatternMatcher pattern = filter.getPattern() == null ? null
+					: PathPatternMatcher.fromGlobPattern(filter.getPattern(), null);
+			if (language == null && scheme == null && pattern == null) {
+				continue; // the spec requires at least one property to be set; ignore invalid filters
+			}
+			filters.add(new CompiledFilter(language, scheme, pattern));
+		}
+		// a non-null selector without any usable filter matches no documents
+		return new DocumentSelectorMatcher(List.copyOf(filters));
+	}
+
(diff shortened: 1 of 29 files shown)


org.eclipse.lsp4e/src/org/eclipse/lsp4e/internal/files/PathPatternMatcher.java
```diff
@@ -23,10 +23,68 @@
...
+			baseDir = workspaceFolder.getUri();
+		}
+		if (baseDir == null || baseDir.isBlank())
+			return null;
+
+		try {
+			return Paths.get(URI.create(baseDir));
+		} catch (final Exception ex) {
+			// Invalid baseUri, ignore the pattern
+			LanguageServerPlugin.logWarning(ex.getMessage(), ex);
```
rubenporras, line 83:
```suggestion
			LanguageServerPlugin.logWarning("Invalid base directory " + baseDir, ex);
```

org.eclipse.lsp4e/src/org/eclipse/lsp4e/internal/DocumentSelectorMatcher.java
```diff
@@ -0,0 +1,127 @@
...
+
+		private boolean matchesPattern(final URI uri) {
+			final PathPatternMatcher pattern = this.pattern;
+			if (pattern == null) {
+				return false;
+			}
+			final Path path;
+			try {
+				path = Paths.get(uri);
+			} catch (final Exception ex) {
```
rubenporras, line 73: should we not be more specific and check for `FileSystemNotFoundException`?

rubenporras:
@FlorianKroiss , I will do some code review but I do not have much time to do a proper review. Would you like to do that?

ahmedneilhussain:
> @FlorianKroiss , I will do some code review but I do not have much time to do a proper review. Would you like to do that?

Hi @rubenporras! Do you know of any eclipse consumers of LSP4e that definitely use dynamic registration? I was looking to see whether any of the original implementations was done by people who were after specific support for things for their own purposes, because they'd be in the best position to test the changes end-to-end and would also be the people affected if there were any problems.

I will take a look at Corrosion and also maybe if there is an integration of `clangd`.

angelozerr:
I repeat me, but I suggest that you see https://github.com/eclipse-lsp4e/lsp4e/issues/1419

In LSP4IJ I have implemented all dynamic  register capability with filters, etc

rubenporras:
> > @FlorianKroiss , I will do some code review but I do not have much time to do a proper review. Would you like to do that?
> 
> Hi @rubenporras! Do you know of any eclipse consumers of LSP4e that definitely use dynamic registration?

No, this an area of the codebase I do not know at all.

rubenporras:
> I repeat me, but I suggest that you see #1419
> 
> In LSP4IJ I have implemented all dynamic register capability with filters, etc

@ahmedneilhussain , did you look at the implementation in LSP4J?

ahmedneilhussain:
> > I repeat me, but I suggest that you see #1419
> > In LSP4IJ I have implemented all dynamic register capability with filters, etc
> 
> @ahmedneilhussain , did you look at the implementation in LSP4J?

Yes - see my comment https://github.com/eclipse-lsp4e/lsp4e/pull/1586#issuecomment-5570158754 above.

As I say, the architecture @angelozerr has for LSP4ij has a lot of pluses in its own right, but it's a lot of work to get LSP4e over to that architecture, since it would affect pretty much all call sites.

By my understanding, we would replace a lot of code that has something like

`LanguageServers.forDocument(doc).withFilter(Predicate<ServerCapabilities> predicate)` 

Where `predicate` might be something like `sc -> sc.getCodeLensProvider() != null`

E.g. https://github.com/eclipse-lsp4e/lsp4e/blob/0dcbf2df66e2bec83d029949850a85b6fc3bf689/org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/codelens/CodeLensProvider.java#L35

LSP4IJ instead exposes a [distributed, overridable] ClientFeatures api, so rather than querying the server capabilities directly, you have

```
        return getLanguageServers(file,
                        f -> f.getCodeLensFeature().isEnabled(file),
                        f -> f.getCodeLensFeature().isSupported(file))
```
as in https://github.com/redhat-developer/lsp4ij/blob/3bd9a7f13e4f3ae5f105b4f52ccf0561f22d8773/src/main/java/com/redhat/devtools/lsp4ij/features/codeLens/LSPCodeLensSupport.java#L58

So `LSPClientFeatures` exposes `LSPCodeLensFeature` which has methods such as 
`public boolean isCodeLensSupported(@NotNull PsiFile file)`
and `public boolean isResolveCodeLensSupported(@NotNull PsiFile file)`
thus calling code uses these higher-level methods to check for functionality rather than querying directly against the `ServerCapabilities`. 

Underlying `LSPCodeLensFeature` is a `CodeLensCapabilityRegistry` that links back to the top-level `ServerCapabilities` *and any dynamic registrations relevant to this feature*.

One of the things `CodeLensCapabilityRegistry` does is have canned static predicate functors against the `ServerCapabilities` and dynamic `CodeLensRegistrationOptions` - this is the sort of thing that we have at call sites in LSP4e, but which LSP4IJ wraps up so call sites can query the higher-level API supported by `LSPCodeLensFeatures` - but the predicates still need to live somewhere!

So whilst the LSP4IJ architecture is nice, it would mean rearchitecting pretty much everything, and would move where the fix lives but not reduce the amount of work needed for the fix.

One advantage of the LSP4IJ architecture is that it abandons any pretence that the dynamic capabilities can be inserted into the static `ServerCapabilities`, and just treats the two sets as things that need to be queried separately and the results combined.

The goal with my implementation was to be correct without a wholesale rewrite of *all* the calling sites, let alone a complete rearchitecture, hence the need to compute an 'effective server capabilities' so that all the existing calling sites could continue as they are.

angelozerr:
@ahmedneilhussain I don't say to move to LSPCLientFeatures for now (even if I think it is super nice because you can override the capability by using for instance a user preferences) but I suggest that you copy/paste TextDocumentServerCapabilityRegistry, CodeLensCapabilityRegistry  etc and adapt it to LSP4E. Those classe manages the same regsitry capability than vscode. Advanced lsp server like rust, metalls etc uses advanced register capability an dit covers all those lsp servers.

ahmedneilhussain:
> @ahmedneilhussain I don't say to move to LSPCLientFeatures for now (even if I think it is super nice because you can override the capability by using for instance a user preferences) but I suggest that you copy/paste TextDocumentServerCapabilityRegistry, CodeLensCapabilityRegistry etc and adapt it to LSP4E. Those classe manages the same regsitry capability than vscode. Advanced lsp server like rust, metalls etc uses advanced register capability an dit covers all those lsp servers.

But unfortunately how is calling code to make use of those registries, when all calling sites expect just to query a flat `ServerCapabilities`?

PR #1591, opened 2026-09-15
build: pin actions to a full-length commit SHA

rubenporras:
Pinning an action to a full-length commit SHA is currently the only way to use an action as an immutable release.

A tag can be moved or deleted if a bad actor gains access to the repository storing the action, pinning to a particular SHA helps mitigate this risk, as the bad actor would need to generate a SHA-1 collision for a valid Git object payload.


FlorianKroiss (approved):
Nice work! I think this also solves some CodeQL problems 😄

PR #1594, opened 2026-09-22
build: group dependabot updates of multi-path actions

akurtakov:
Sub-actions coming from one repository, like codeql-action/init and codeql-action/analyze, must be bumped in lockstep. Without grouping dependabot opens one PR per sub-action and each of them fails, as the mismatching versions are rejected by the codeql-action itself.

Group github/codeql-action* and actions/cache* so that every such repository is updated by a single PR.

Assisted-by: Anthropic Claude Code (claude-opus-5[1m])


akurtakov:
Merge as it will not be verified by any action.

